PR #346, opened 2018-10-21
Worker for watching NFT minting events

vimukthi-git:


manuelpolzhofer:
I reopened the PR under #400

PR #385, opened 2018-10-26
Flatten Core document and P2P packages

vedhavyas:


coredocument/processor.go
```diff
@@ -34,15 +32,22 @@ type Processor interface {
 	SendDocument(ctx context.Context, model documents.Model) error
 }
 
+// client defines the methods for p2pclient
+// we redefined it here so that we can avoid cyclic dependencies with p2p
+type client interface {
```
mikiquantum: This is a symptom of something that we can simplify even further. CoreDocument defines a model, and it is interchanged through the P2P layer. At some point, we will only have one copy or opened channel for a document state shared across the participants. 
What do you think about consolidating p2p as part of the coredocument package (or a different name)? But having the document and the operations over it more contained.

vedhavyas: To be honest, I want to delete the `P2PClient` interface definition because p2p package is implementing it. Ideally, it should be defined in coredocuemnt since that is using the interface. But combining both p2p and coredocument is something I'm really not sure because each of them seems to have specific purpose. Maybe something we can discuss this week @mikiquantum?

PR #387, opened 2018-10-29
Add Connect timeout to p2p conns

mikiquantum:


config/configuration.go
```diff
@@ -59,6 +59,11 @@ func (c *Configuration) GetP2PExternalIP() string {
 	return c.V.GetString("p2p.externalIP")
 }
 
+// P2P Connect Timeout
+func (c *Configuration) GetP2PConnectionTimeout() time.Duration {
```
vedhavyas: comment should start with `GetP2PConnectionTimeout ...`

p2p/client.go
```diff
@@ -62,17 +61,14 @@ func (d *defaultClient) OpenClient(target string) (p2ppb.P2PServiceClient, error
 	// so LibP2P knows how to contact it
 	hostInstance.Peerstore().AddAddr(peerID, targetAddr, pstore.PermanentAddrTTL)
 
-	// make a new stream from host B to host A
-	g, err := grpcProtoInstance.Dial(context.Background(), peerID, grpc.WithInsecure())
+	// make a new stream from host B to host A with timeout
+	// Retrial is handled internally, connection request will be cancelled by the connection timeout context
+	ctx, _ := context.WithTimeout(context.Background(), config.Config.GetP2PConnectionTimeout())
```
vedhavyas: we should either pass back the `cancelFunc` to the caller so that they cancel it if needed?

or we could just take in a `context` along with the `target` in the function params. I prefer this one over the first since it gives more flexibility of which context I need to pass

p2p/client.go
```diff
@@ -62,17 +61,14 @@ func (d *defaultClient) OpenClient(target string) (p2ppb.P2PServiceClient, error
...
+	g, err := grpcProtoInstance.Dial(ctx, peerID, grpc.WithInsecure(), grpc.WithBlock())
```
vedhavyas: `grpc.WithBlock()` nice!

p2p/client.go
mikiquantum: I am not sure if I need to expose that information to the caller, as the context is only used at this point for dialing and opening the connection and it will error out and be returned back to the caller, where it can be acted upon. 
Then the actual rpc call is done afterwards and for that one is passed a different context.

PR #388, opened 2018-10-30
Added timeout for eth read operations

mikiquantum:


anchors/anchor_test.go
```diff
@@ -5,10 +5,24 @@ package anchors
 import (
 	"testing"
 
+	"os"
```
vedhavyas: group imports?

PR #396, opened 2018-10-31
Identity Config cleanup

mikiquantum:


mikiquantum:
Not sure why it didn't trigger codecov feedback.

https://codecov.io/gh/centrifuge/go-centrifuge/commit/e3b4ac81ec624179d83e35c28621d539f77592d9

identity/identity.go
```diff
@@ -31,6 +31,52 @@ var IDService Service
...
+}
+
+// IdentityKey represents a key pair
+type IdentityKey struct {
+	PublicKey  []byte
+	PrivateKey []byte
+}
+
+// GetIdentityConfig returns the identity and keys associated with the node
+func GetIdentityConfig() (*IdentityConfig, error) {
```
vedhavyas: I think we `GetIdentityConfig` can take a purpose like `KeyPurposeSigning` return that config alone. What do you think?

vedhavyas:
@mikiquantum any chance we can improve the coverage for this ?

mikiquantum:
@vedhavyas I could improve the test coverage for areas that are not related to the change, if that is what you are asking.

identity/identity.go
mikiquantum: I thought about that at the beginning, but sounds that reading 1 or 3 keys does not create a significant penalty. 
Knowing that in the future we might read it once and pass it to through the context.

PR #397, opened 2018-10-31
WIP: Refactor anchor and geth

vedhavyas:


ethereum/geth_client.go
```diff
@@ -59,217 +58,251 @@ type Config interface {
...
+	// get current nonce
+	n, err := noncer.PendingNonceAt(ctx, opts.From)
+	if err != nil {
+		return fmt.Errorf("failed to get chain nonce for %s: %v", opts.From.String(), err)
 	}
 
-	return
-}
+	// set the nonce
+	opts.Nonce = new(big.Int).Add(new(big.Int).SetUint64(n), big.NewInt(1))
```
mikiquantum: I rather not increment by default, this could cause problems. I do not think we need this here, as it should only be incremented when there are many pending txs.

ethereum/geth_client.go
```diff
@@ -59,217 +58,251 @@ type Config interface {
...
+		opts.Nonce = nil
+		return nil
 	}
 
-	var res map[string]map[string]map[string][]string
-	// Important to not create a DeadLock if network latency
-	txctx, _ := context.WithTimeout(context.Background(), GetDefaultContextTimeout())
-	gc.GetRpcClient().CallContext(txctx, &res, "txpool_inspect")
+	ctx, cancel := context.WithTimeout(context.Background(), GetDefaultContextTimeout())
+	defer cancel()
```
mikiquantum: deferring `cancel` on an already done context would it have any effect?

ethereum/geth_client.go
```diff
@@ -59,217 +58,251 @@ type Config interface {
...
+// incrementNonce updates the opts.Nonce to next valid nonce
+// We pick the current nonce by getting latest transactions included in the blocks including pending blocks
+// then we check the txpool to see if there any new transactions from the address that are not included in any block
+// If there are no pending transactions in txpool, we use the current nonce + 1
+// else we increment the greater of current nonce or the nonce derived from txpool
+func incrementNonce(opts *bind.TransactOpts, txpoolAccessEnabled bool, noncer noncer, cc callContexter) error {
+	// check if the txpool access is enabled
+	if !txpoolAccessEnabled {
+		log.Warningf("Ethereum Client doesn't support txpool API, may cause transactions failures")
+		opts.Nonce = nil
```
mikiquantum: Why do we need to nil nonce here? is it what is passed by default right?

vedhavyas: I was passed as nil by default but we were `nil`ing nonce at GetOpts. Since this `incrementNonce` is the only place we check the nonce, I felt this is the right place. 

Coming to `nil`ing the nonce here, since the `txnpoolaccess` is not enabled, we will just let the geth-node pick the nonce from the pending transactions.

ethereum/geth_client.go
```diff
@@ -59,217 +58,251 @@ type Config interface {
...
 		keys = append(keys, ki)
 	}
+
+	// there are some pending transactions in txpool, check their nonce
+	// pick the largest one and increment it
 	sort.Ints(keys)
 	lastPoolNonce := keys[len(keys)-1]
-	if uint64(lastPoolNonce) >= chainNonce {
+	if uint64(lastPoolNonce) >= n {
 		opts.Nonce = new(big.Int).Add(big.NewInt(int64(lastPoolNonce)), big.NewInt(1))
```
mikiquantum: Can we keep the log.Info here?

ethereum/geth_client.go
mikiquantum: We still need to nil the `opts.nonce` at the beginning of any new tx request otherwise we will have a problem with high concurrency as cached opts per account will be outdated.

vedhavyas: Yes, this will be the place it gets either nil if the `transaction pool access` is not enabled. If enabled, we will get the nonce from the blocks including pending and then we will get the txns from the pool, get the highest nonce and and increment it ourselves.

ethereum/geth_client.go
vedhavyas: it is still there. Moved to upper block

ethereum/geth_client.go
vedhavyas: We are setting the nonce here but we will still look up for any pending txns. If we find the any pending txns, we will use that

ethereum/geth_client.go
vedhavyas: This would release the ctx resources if any

ethereum/geth_client.go
mikiquantum: ok, now looks good as we default to the pending nonce a bit below. Thanks!

PR #398, opened 2018-10-31
clean up anchors

vedhavyas:


vedhavyas:
@mikiquantum this is only anchors. Will give another PR for geth client.

mikiquantum:
@vedhavyas from the https://github.com/centrifuge/go-centrifuge/issues/337 issue, I thought that was targeting identities as an initial pass. Maybe if that is not accurate want to update the issue to be on the same page?

vedhavyas:
Updated @mikiquantum

mikiquantum:
Thanks!

PR #409, opened 2018-11-06
Hex convert webhook fields

vimukthi-git:
We are in the process of converting all our byte encoding to 0x prefixed hex. This adds support for hex in webhooks.


build/swagger_config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
```
mikiquantum: Should we change as well the `build/package.json` to match the new swagger version?

PR #441, opened 2018-11-12
 First batch of removing global config access

mikiquantum:
Focus of this pr:
- Remove config.Config() access only for struct member methods
- Added `config` to some structs that were easy
- Adapted the tests for it (hybrid with global bootstrapping)
- Un-exported some members of structs


api/service.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 // registerServices registers all endpoints to the grpc server
-func registerServices(ctx context.Context, grpcServer *grpc.Server, gwmux *runtime.ServeMux, addr string, dopts []grpc.DialOption) error {
+func registerServices(ctx context.Context, config Config, grpcServer *grpc.Server, gwmux *runtime.ServeMux, addr string, dopts []grpc.DialOption) error {
```
mikiquantum: Question to @vimukthi-git @vedhavyas . According to the go context.Context it is meant to pass arbitrary information that is needed for the downstream processing. As we know we are passing the context pretty much from the beginning of the request, is it good practice to add the config there? instead of a separate attribute? What about other shared resources?

vimukthi-git: I'm fine with either way, but seems `context.Context` is generally used for more dynamic info specific to a request or execution? I'm not quite sure. I'm fine with what you and @vedhavyas agrees on this.

coredocument/coredocument_test.go
```diff
@@ -25,6 +29,26 @@ var (
 	id5 = utils.RandomSlice(32)
 )
 
+var ctx = map[string]interface{}{}
+var cfg *config.Configuration
```
vimukthi-git: Can we come up with a way to not to use a global ctx and cfg variables in places like these? I hope there is a way to pass some context to each test?

mikiquantum: I agree that we shouldn't use those globals in tests. I will try to remove them as part of the next PR, otherwise it might need to wait until we get rid of the bootstrapping logic.

api/service.go
mikiquantum: Thanks @vimukthi-git . I will instead wrap the `context.Context` in the centrifuge `Context` plus the config, so we can have a way to pass request specific data downstream.

api/server.go
```diff
@@ -24,23 +24,19 @@ import (
 
 var log = logging.Logger("cent-api-server")
 
+type Config interface {
```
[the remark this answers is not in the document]
mikiquantum: ok! I thing I will have to remove this `Config` interface abstractions in future PR. As it causes limitations on knowledge of that service of what other config methods should be available downstream.

api/service.go
mikiquantum: Exactly what I faced, to many castings downstream. Instead I am passing the typed centrifuge context.

PR #442, opened 2018-11-12
nft: added format checks for collaborator proofField

manuelpolzhofer:
- added format check for collaborator proofField
- added format check for passed ethereum addresses

### DIFF
--- a/nft/ethereum_payment_obligation.go
+++ b/nft/ethereum_payment_obligation.go
@@ -107,7 +107,7 @@ func (s *ethereumPaymentObligation) MintNFT(documentID []byte, docType, registry
 		return nil, err
 	}
 
-	requestData, err := NewMintRequest(toAddress, anchorID, proofs.FieldProofs, rootHash,collaboratorField)
+	requestData, err := NewMintRequest(toAddress, anchorID, proofs.FieldProofs, rootHash, collaboratorField)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


nft/ethereum_payment_obligation.go
```diff
@@ -287,3 +293,26 @@ func getDocumentService(documentType string) (documents.Service, error) {
...
+
+// getCollaborator returns the needed collaboratorField for a PaymentObligation NFT
+// In the current contract implementation the proofField for collaborator is a separated parameter
+// pattern: 'collaborators' + '[i]'
+// examples: 'collaborators[0]','collaborators[1]', etc
+func getCollaboratorProofField(proofFields []string) (string, error) {
+
+	for _, proofField := range proofFields {
+
+		match, err := regexp.MatchString("collaborators\\[[0-9]*\\]", proofField)
```
vedhavyas: Can we use this instead `collaborators\[[0-9]+\]` ?

This will avoid the edge case `collaborators[]`

nft/handler.go
```diff
@@ -36,3 +43,17 @@ func (g grpcHandler) MintNFT(context context.Context, request *nftpb.NFTMintRequ
 	watchToken := <-confirmation
 	return &nftpb.NFTMintResponse{TokenId: watchToken.TokenID.String()}, watchToken.Err
 }
+
+func (g grpcHandler) validateParameters(request *nftpb.NFTMintRequest) error {
+
+	if !common.IsHexAddress(request.RegistryAddress) {
```
vedhavyas: Usage of these addresses part of next PR?

manuelpolzhofer: yes

nft/ethereum_payment_obligation.go
manuelpolzhofer: good catch. I thought about it yesterday to change it.

PR #443, opened 2018-11-12
NFT: using passed registryAddress for dynamic contract binding

manuelpolzhofer:
- binding contract to address happens dynamically in the mint method instead of inside the bootstrapper
- added registryAddress to queue task
- run queue task dynamically binds the contract with the provided address or default address


manuelpolzhofer:
please review PR #442 before

nft/ethereum_payment_obligation.go
```diff
@@ -55,20 +56,24 @@ type ethereumPaymentObligationContract interface {
 
 // ethereumPaymentObligation handles all interactions related to minting of NFTs for payment obligations on Ethereum
 type ethereumPaymentObligation struct {
-	paymentObligation ethereumPaymentObligationContract
 	identityService   identity.Service
 	ethClient         ethereum.Client
 	config            Config
-	setupMintListener func(tokenID *big.Int) (confirmations chan *WatchTokenMinted, err error)
+	setupMintListener func(tokenID *big.Int, registryAddress string) (confirmations chan *WatchTokenMinted, err error)
+	newContract       func(address common.Address) (*EthereumPaymentObligationContract, error)
```
vedhavyas: Can we rename this as `fetchContract` ?

nft/ethereum_payment_obligation.go
```diff
@@ -114,17 +119,36 @@ func (s *ethereumPaymentObligation) MintNFT(documentID []byte, docType, registry
...
 	}
 
-	watch, err := s.setupMintListener(requestData.TokenID)
+	watch, err := s.setupMintListener(requestData.TokenID, registryAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	var contract *EthereumPaymentObligationContract
+	if registryAddress == "" {
```
vedhavyas: I think we can avoid this if checks if we can update the registryAddress at the API level itself. What do you think ?

PR #449, opened 2018-11-14
Fix/p2p test

vedhavyas:


p2p/server.go
```diff
@@ -68,10 +68,9 @@ func (s *p2pServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr ch
 	// Set the grpc protocol handler on it
 	s.protocol = p2pgrpc.NewGRPCProtocol(ctx, s.host)
 	p2ppb.RegisterP2PServiceServer(s.protocol.GetGRPCServer(), &Handler{})
-	errOut := make(chan error)
-	go func(proto *p2pgrpc.GRPCProtocol, errOut chan<- error) {
-		errOut <- proto.Serve()
-	}(s.protocol, errOut)
+	go func(proto *p2pgrpc.GRPCProtocol) {
+		proto.Serve()
```
vimukthi-git: @vedhavyas @manuelpolzhofer why are we skipping error handling logic here? Please add it back if its just a work around. there is a reason why the handling of error here matters otherwise the p2p layer will hang.

vedhavyas: Skiping the error as startup error because `proto.Serve` this is not a startup error but rather a shutdown error. I'm handling this later in a different PR

PR #457, opened 2018-11-16
remove global registry service

### LINKED ISSUE #417: [Integration Test] Remove service registry global
Acceptance criteria:
- Service registry should not be accessible globally but passed in as a dependancy to the node.Server implementations if they need it. Must be created at Node level.

vedhavyas:
Removes global service registry

Fixes #417

### DIFF
--- a/documents/bootstrapper.go
+++ b/documents/bootstrapper.go
@@ -7,8 +7,6 @@ type Bootstrapper struct{}
 
 // Bootstrap sets the required storage and registers
 func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
-	ctx[BootstrappedRegistry] = &ServiceRegistry{
-		services: make(map[string]Service),
-	}
+	ctx[BootstrappedRegistry] = NewServiceRegistry()
 	return nil
 }
--- a/documents/bootstrapper.go
+++ b/documents/bootstrapper.go
@@ -1,8 +1,14 @@
 package documents
 
+// BootstrappedRegistry is the key to ServiceRegistry in Bootstrap context
+const BootstrappedRegistry = "BootstrappedRegistry"
+
 type Bootstrapper struct{}
 
 // Bootstrap sets the required storage and registers
-func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
+func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
+	ctx[BootstrappedRegistry] = &ServiceRegistry{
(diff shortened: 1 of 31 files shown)


p2p/server.go
```diff
@@ -67,10 +67,13 @@ func (s *p2pServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr ch
...
 	// Set the grpc protocol handler on it
 	s.protocol = p2pgrpc.NewGRPCProtocol(ctx, s.host)
-	p2ppb.RegisterP2PServiceServer(s.protocol.GetGRPCServer(), &Handler{})
-	go func(proto *p2pgrpc.GRPCProtocol) {
-		proto.Serve()
-	}(s.protocol)
+	p2ppb.RegisterP2PServiceServer(s.protocol.GetGRPCServer(), &handler{})
+
+	serveErr := make(chan error)
+	go func() {
```
vimukthi-git: 👍

keytools/generate_test.go
```diff
@@ -18,54 +18,41 @@ const (
...
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	os.Remove(publicFileName)
 	os.Remove(privateFileName)
-
-	return
-
+	return publicKey, privateKey
```
manuelpolzhofer: What is the recommended style guide?

documents/bootstrapper.go
```diff
@@ -1,8 +1,14 @@
...
 
+// BootstrappedRegistry is the key to ServiceRegistry in Bootstrap context
+const BootstrappedRegistry = "BootstrappedRegistry"
+
 type Bootstrapper struct{}
 
 // Bootstrap sets the required storage and registers
-func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
+func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
+	ctx[BootstrappedRegistry] = &ServiceRegistry{
```
manuelpolzhofer: Why not calling NewServiceRegistry() here?

keytools/generate_test.go
vedhavyas: Recommended style is to not just say `return` because a reader has to put an extra effort to see what instances we are returning.

Returning `return arg1, arg2` would be super helpful and avoids some confusion.

documents/bootstrapper.go
vedhavyas: good catch!

PR #460, opened 2018-11-18
Remove global configuration usage

mikiquantum:
- Added Config interface
- Usage of contextHeader
- Added default timeoutParam to queues

Will take care of test usage of global config in the package in a separate PR.


bootstrap/bootstrapper.go
```diff
@@ -3,14 +3,8 @@ package bootstrap
...
 const (
-	BootstrappedConfigFile       string = "BootstrappedConfigFile"
-	BootstrappedConfig           string = "BootstrappedConfig"
-	BootstrappedLevelDb          string = "BootstrappedLevelDb"
-	BootstrappedEthereumClient   string = "BootstrappedEthereumClient"
-	BootstrappedAnchorRepository string = "BootstrappedAnchorRepository"
-	BootstrappedP2PClient        string = "BootstrappedP2PClient"
-	BootstrappedP2PServer        string = "BootstrappedP2PServer"
-	BootstrappedAPIServer        string = "BootstrappedAPIServer"
+	BootstrappedP2PServer string = "BootstrappedP2PServer"
```
mikiquantum: Moved most of the bootstrapped objects to its ow packages, except those that are used commonly.

config/configuration.go
```diff
@@ -27,24 +27,35 @@ import (
...
-	defer configMu.RUnlock()
-	return config
-}
-
-// SetConfig sets the config
-func SetConfig(c *Configuration) {
-	configMu.Lock()
-	defer configMu.Unlock()
-	config = c
+type Config interface {
```
mikiquantum: Added config interface so it can be mocked during tests. In addition for multi-tenant nodes, will come handy when we split up in nodeConfig and tenantConfig.

mikiquantum:
Will add more tests.

anchors/anchor_test.go
```diff
@@ -12,11 +12,15 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var ctx = map[string]interface{}{}
+var cfg *config.Configuration
```
vimukthi-git: I would love to avoid adding these, they create problems for running tests in parallel.

anchors/ethereum_anchor_repository.go
```diff
@@ -142,9 +141,10 @@ func sendCommitTransaction(contract AnchorRepositoryContract, opts *bind.Transac
 	return
 }
 
+// TODO: This method is only used by setUpPreCommitEventListener below, it will be changed soon so we can remove the hardcoded `time.Second` and use the global one
```
vimukthi-git: lets create an issue instead of adding TODOs

coredocument/validator.go
```diff
@@ -201,7 +202,7 @@ func documentRootValidator() documents.Validator {
 // re-calculates the signature and compares with existing one
 // assumes signing_root is already generated and verified
 // Note: this needs to used only before document is sent for signatures from the collaborators
-func readyForSignaturesValidator() documents.Validator {
+func readyForSignaturesValidator(ctxHeader *header.ContextHeader) documents.Validator {
```
vimukthi-git: ctxHeader is a good thing as long as it doesn't replace everything we want to pass to a function with a single parameter. It can be a dramatic anti pattern in code since it wouldn't be obvious what you need to pass to a function by looking at its signature. Let's have a discussion on this change.

coredocument/validator.go
```diff
@@ -200,7 +201,7 @@ func documentRootValidator() documents.Validator {
 // re-calculates the signature and compares with existing one
 // assumes signing_root is already generated and verified
 // Note: this needs to used only before document is sent for signatures from the collaborators
-func readyForSignaturesValidator() documents.Validator {
+func readyForSignaturesValidator(centIDBytes, priv, pub []byte) documents.Validator {
```
mikiquantum: @vimukthi-git changed.

anchors/ethereum_anchor_repository.go
```diff
@@ -37,12 +36,12 @@ type WatchAnchorCommitted interface {
 }
 
 type EthereumAnchorRepository struct {
-	config                   Config
+	config                   config.Config
```
vimukthi-git: This breaks the pattern of what we are doing on other places. Can we just stick to one config interface per package for the sake of consistency? Using config.Config for mocking in tests is totally fine.
The reason for config per package is that its obvious and explicit what each package depends on rather than having a very broad definition of what it needs to function. Very useful for reuse. 

Excuse me as I didn't notice this change the previous review.

vimukthi-git: Also go convention is to define the interface at the place of usage(anchors package) rather than the place of implementation(config package).

coredocument/coredocument.go
```diff
@@ -209,12 +210,9 @@ func NewWithCollaborators(collaborators []string) (*coredocumentpb.CoreDocument,
 }
 
 //  GetExternalCollaborators returns collaborators of a document without the own centID
-func GetExternalCollaborators(doc *coredocumentpb.CoreDocument) ([][]byte, error) {
+func GetExternalCollaborators(ctxHeader *header.ContextHeader, doc *coredocumentpb.CoreDocument) ([][]byte, error) {
```
vedhavyas: Can we pass the `SelfCentID` here instead of entire config or context ?

api/server.go
```diff
@@ -25,15 +26,9 @@ import (
...
-type Config interface {
-	GetServerAddress() string
-	GetServerPort() int
-	GetNetworkString() string
-}
-
 // apiServer is an implementation of node.Server interface for serving HTTP based Centrifuge API
 type apiServer struct {
-	config   Config
+	config   config.Config
```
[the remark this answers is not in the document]
mikiquantum: Vimukthi and I discussed this earlier, with multi-tenant we will split up the current config and it will change enough that making some assumptions now might be a waste of time. We will tackle that later.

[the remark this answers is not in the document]
vimukthi-git: We need to have a deep discussion on this config and construction of objects in the node. I feel we are not on the same page in terms of the final architecture we want to achieve.

mikiquantum:
@vimukthi-git  As discussed I added the package Configs back and did some casting on those methods that propagate the config downstream.
We are leaving as common Config the handlers, some bootstrappers and the test cases as they need to access the write methods.

PR #469, opened 2018-11-19
Queue implements Server and removing the queue.Queue global

vimukthi-git:


anchors/anchor_confirmation_task.go
```diff
@@ -48,17 +48,11 @@ type anchorConfirmationTask struct {
 	AnchorCommittedFilterer anchorCommittedWatcher
 }
 
-// Name returns AnchorRepositoryConfirmationTaskName
-func (act *anchorConfirmationTask) Name() string {
```
vedhavyas: IMO, Name seems simple and better. Any reason why we are going with `TaskTypeName`?
It seems too verbose

anchors/bootstrapper.go
```diff
@@ -31,8 +32,12 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return err
 	}
 
-	var repo AnchorRepository
-	repo = NewEthereumAnchorRepository(cfg, repositoryContract, ethereum.GetClient)
+	if _, ok := ctx[bootstrap.BootstrappedQueueServer]; !ok {
```
vedhavyas: can you move the `BootstrappedQueueServer` to queue package instead. This is would help us remove the cyclic dependencies

anchors/ethereum_anchor_repository.go
```diff
@@ -38,10 +37,11 @@ type EthereumAnchorRepository struct {
 	config                   Config
 	anchorRepositoryContract AnchorRepositoryContract
 	gethClientFinder         func() ethereum.Client
+	queue                    *queue.Server
 }
 
-func NewEthereumAnchorRepository(config Config, anchorRepositoryContract AnchorRepositoryContract, gethClientFinder func() ethereum.Client) *EthereumAnchorRepository {
-	return &EthereumAnchorRepository{config: config, anchorRepositoryContract: anchorRepositoryContract, gethClientFinder: gethClientFinder}
+func NewEthereumAnchorRepository(config Config, anchorRepositoryContract AnchorRepositoryContract, queue *queue.Server, gethClientFinder func() ethereum.Client) *EthereumAnchorRepository {
```
vedhavyas: If this function is not being outside the package, can you un-export this?

If being used outside, Can you add comments to this?

bootstrap/test_bootstrapper.go
```diff
@@ -12,12 +12,12 @@ type TestBootstrapper interface {
 	TestTearDown() error
 }
 
-func RunTestBootstrappers(bootstrappers []TestBootstrapper, context map[string]interface{}) {
-	if context == nil {
-		context = map[string]interface{}{}
+func RunTestBootstrappers(bootstrappers []TestBootstrapper, ctx map[string]interface{}) {
+	if ctx == nil {
```
vedhavyas: 👍

identity/bootstrapper.go
```diff
@@ -35,25 +36,23 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
 		return err
 	}
 
-	IDService = NewEthereumIdentityService(cfg, idFactory, registryContract, ethereum.GetClient,
+	if _, ok := context[bootstrap.BootstrappedQueueServer]; !ok {
+		return errors.New("queue hasn't been initialized")
+	}
+	queueSrv := context[bootstrap.BootstrappedQueueServer].(*queue.Server)
+
+	IDService = NewEthereumIdentityService(cfg, idFactory, registryContract, queueSrv, ethereum.GetClient,
```
vedhavyas: I thought we got rid of this `IDService`, didn't we?

queue/bootstrapper.go
```diff
@@ -3,41 +3,22 @@ package queue
 import (
 	"errors"
 
+	"github.com/centrifuge/go-centrifuge/bootstrap"
 	"github.com/centrifuge/go-centrifuge/config"
+	"sync"
```
vedhavyas: Move the standard import to top?

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+	// GetNumWorkers gets the number of background workers to initiate
+	GetNumWorkers() int
+
+	// GetWorkerWaitTime gets the worker wait time for a task to be available while polling
+	// increasing this may slow down task execution while reducing it may consume a lot of CPU cycles
+	GetWorkerWaitTimeMS() int
+}
+
+// QueuedTaskType is a task to be queued in the centrifuge node to be completed asynchronously
+type QueuedTaskType interface {
```
vedhavyas: `QueuedTaskType` is not very intutive for me because Interface says it is type but the methods only has `TaskNameType`. I would go with `fmt.Stringer` for this

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// TaskTypeName of the task
+	TaskTypeName() string
+}
+
+// QueuedTaskResult represents a result from a queued task execution
+type QueuedTaskResult interface {
```
vedhavyas: `QueuedTaskResult` seems redundant and can be written as `TaskResult` IMO?

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+// TaskTypeName of the queue server
+func (qs *Server) Name() string {
+	return "QueueServer"
+}
+
+// Start the queue server
+func (qs *Server) Start(ctx context.Context, wg *sync.WaitGroup, startupErr chan<- error) {
+	defer wg.Done()
+	qs.lock.Lock()
+	defer qs.lock.Unlock()
```
vedhavyas: If you are locking it here, no other functions could get a lock

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+	var err error
+	qs.queue, err = gocelery.NewCeleryClient(
+		gocelery.NewInMemoryBroker(),
+		gocelery.NewInMemoryBackend(),
```
vedhavyas: Let's start using the LevelDB backend here?

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+func (qs *Server) RegisterTaskType(name string, task interface{}) {
+	qs.lock.Lock()
+	defer qs.lock.Unlock()
+	qs.taskTypes = append(qs.taskTypes, task.(QueuedTaskType))
+}
+
+// EnqueueJob enqueues a job on the queue server for the given taskTypeName
+func (qs *Server) EnqueueJob(taskTypeName string, params map[string]interface{}) (QueuedTaskResult, error) {
+	qs.lock.Lock()
+	defer qs.lock.Unlock()
```
vedhavyas: We dont need locking here since `DelayKwargs` is thread safe.

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+	for _, task := range qs.taskTypes {
+		qs.queue.Register(task.TaskTypeName(), task)
+	}
+	// start the workers
+	qs.queue.StartWorker()
+
+	qs.stop = make(chan bool)
+	for {
+		select {
+		case <-qs.stop:
```
vedhavyas: This would not work since select doesn't fallthrough. I would recommend using context.WithCancel and hold the cancelFunc instead of channel

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+	qs.lock.Lock()
+	defer qs.lock.Unlock()
+	return qs.queue.DelayKwargs(taskTypeName, params)
+}
+
+// Stop force stops the queue server
+func (qs *Server) Stop() error {
+	qs.lock.Lock()
+	defer qs.lock.Unlock()
+	if qs.stop != nil {
```
vedhavyas: Dont need locking here as well since channels are thread safe. Would prefer it if we have cancelFunc instead of channel here

queue/server.go
```diff
@@ -0,0 +1,138 @@
...
+
+// ParseBlockHeight parses blockHeight interface param to uint64
+func ParseBlockHeight(valMap map[string]interface{}) (uint64, error) {
+	if bhi, ok := valMap[BlockHeightParam]; ok {
+		bhf, ok := bhi.(float64)
+		if ok {
+			return uint64(bhf), nil
+		} else {
+			return 0, errors.New("value can not be parsed")
+		}
```
vedhavyas: else block is redundant here. Can we check `!ok` in `if` ?

queue/test_bootstrapper.go
```diff
@@ -2,11 +2,33 @@
...
 	return b.Bootstrap(context)
 }
 
 func (b *Bootstrapper) TestTearDown() error {
-	StopQueue()
+	srv := b.context[bootstrap.BootstrappedQueueServer].(*Server)
+	return srv.Stop()
+}
+
+type Starter struct {}
```
vedhavyas: Not sure I understand the Starter here.

identity/bootstrapper.go
vimukthi-git: thats the other pr

anchors/bootstrapper.go
vimukthi-git: thats the reason why the constants were created in the `bootstrap` package. it contains only interfaces, if someone creates cyclic dependancies using this there is something wrong with the code.

anchors/anchor_confirmation_task.go
vimukthi-git: The `Name` function is implemented by many unrelated stuff, I dont want to mix them up.

vimukthi-git:
@vedhavyas could you summarise what change is required?

queue/test_bootstrapper.go
vimukthi-git: server has  two stages of initiation, bootstrap and start. This bootstrapper is to start the queue for tests.

queue/server.go
vimukthi-git: This was added because of race conditions in tests, so go compiler disagrees with you

queue/server.go
vimukthi-git: its needed because of a race condition, still trying to fix

queue/server.go
vimukthi-git: no, functional changes please. Leveldb backend should be added when it's required.

queue/server.go
vimukthi-git: race condition in tests, still trying to fix

queue/bootstrapper.go
vimukthi-git: I told you my view on this

queue/server.go
vimukthi-git: Stringer is too generic to give any idea that it's an important part of the functionality of the queue. I want to make it explicit that this function is required to be implemented for task.

vedhavyas:
@vimukthi-git My bad.
- We do need locking on `Start`. This will block everything
- `stop` channnel doesn't work since select will not have fallthrough

queue/server.go
vedhavyas: interesting. I will take a look

vimukthi-git:
@vedhavyas Comments have been addressed.

anchors/ethereum_anchor_repository.go
```diff
@@ -209,7 +209,7 @@ func waitAndRoutePreCommitEvent(conf <-chan *EthereumAnchorRepositoryContractAnc
 }
 
 // waitAndRouteCommitEvent notifies the confirmations channel whenever a commit is being noted as Ethereum event
-func waitAndRouteCommitEvent(timeout time.Duration, asyncResult *gocelery.AsyncResult, confirmations chan<- *WatchCommit, commitData *CommitData) {
+func waitAndRouteCommitEvent(timeout time.Duration, asyncResult queue.TaskResult, confirmations chan<- *WatchCommit, commitData *CommitData) {
```
vedhavyas: `asyncResult` -> `taskResult` ?

api/server.go
```diff
@@ -1,9 +1,5 @@
 package api
 
-// LICENSE: Apache
-// This is taken from https://github.com/philips/grpc-gateway-example/
```
vedhavyas: let's keep the attribution `This is taken from https://github.com/philips/grpc-gateway-example/` unless we are not using the code we copied anymore

bootstrap/bootstrapper.go
```diff
@@ -3,9 +3,10 @@ package bootstrap
 // DO NOT PUT any app logic in this package to avoid any dependency cycles
 
 const (
-	BootstrappedP2PServer string = "BootstrappedP2PServer"
-	BootstrappedAPIServer string = "BootstrappedAPIServer"
-	NodeObjRegistry       string = "NodeObjRegistry"
+	BootstrappedP2PServer   string = "BootstrappedP2PServer"
+	BootstrappedAPIServer   string = "BootstrappedAPIServer"
+	BootstrappedQueueServer string = "BootstrappedQueueServer"
+	NodeObjRegistry         string = "NodeObjRegistry"
```
vedhavyas: Add comments to the exported constants pls

context/bootstrapper.go
```diff
@@ -31,15 +31,15 @@ func (m *MainBootstrapper) PopulateBaseBootstrappers() {
 		&config.Bootstrapper{},
 		&storage.Bootstrapper{},
 		ethereum.Bootstrapper{},
-		anchors.Bootstrapper{},
+		&queue.Bootstrapper{},
+		&anchors.Bootstrapper{},
```
vedhavyas: Why the pointer receiver for both the bootstrappers?

identity/bootstrapper.go
```diff
@@ -39,25 +40,23 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
 		return err
 	}
 
-	context[BootstrappedIDService] = NewEthereumIdentityService(cfg, idFactory, registryContract, ethereum.GetClient,
+	if _, ok := context[bootstrap.BootstrappedQueueServer]; !ok {
+		return errors.New("queue hasn't been initialized")
+	}
+	queueSrv := context[bootstrap.BootstrappedQueueServer].(*queue.Server)
```
vedhavyas: let's do a `,ok` check here as well

identity/ethereum_identity.go
```diff
@@ -335,13 +336,13 @@ func setUpRegistrationEventListener(config Config, identityToBeCreated Identity,
 }
 
 // waitAndRouteKeyRegistrationEvent notifies the confirmations channel whenever the key has been added to the identity and has been noted as Ethereum event
-func waitAndRouteKeyRegistrationEvent(timeout time.Duration, asyncRes *gocelery.AsyncResult, confirmations chan<- *WatchIdentity, pushThisIdentity Identity) {
+func waitAndRouteKeyRegistrationEvent(timeout time.Duration, asyncRes queue.TaskResult, confirmations chan<- *WatchIdentity, pushThisIdentity Identity) {
```
vedhavyas: `asyncRes` -> `taskResult` ?

identity/ethereum_identity.go
```diff
@@ -335,13 +336,13 @@ func setUpRegistrationEventListener(config Config, identityToBeCreated Identity,
...
 	_, err := asyncRes.Get(timeout)
 	confirmations <- &WatchIdentity{Identity: pushThisIdentity, Error: err}
 }
 
 // waitAndRouteIdentityRegistrationEvent notifies the confirmations channel whenever the identity creation is being noted as Ethereum event
-func waitAndRouteIdentityRegistrationEvent(timeout time.Duration, asyncRes *gocelery.AsyncResult, confirmations chan<- *WatchIdentity, pushThisIdentity Identity) {
+func waitAndRouteIdentityRegistrationEvent(timeout time.Duration, asyncRes queue.TaskResult, confirmations chan<- *WatchIdentity, pushThisIdentity Identity) {
```
vedhavyas: `asyncRes` -> `taskResult` ?

nft/bootstrapper.go
```diff
@@ -37,6 +38,14 @@ func (*Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return fmt.Errorf("identity service not initialised")
 	}
 
-	ctx[BootstrappedPayObService] = NewEthereumPaymentObligation(registry, idService, ethereum.GetClient(), cfg, setupMintListener, bindContract)
-	return queue.InstallQueuedTask(ctx, newMintingConfirmationTask(cfg.GetEthereumContextWaitTimeout(), ethereum.DefaultWaitForTransactionMiningContext))
+	if _, ok := ctx[bootstrap.BootstrappedQueueServer]; !ok {
+		return errors.New("queue hasn't been initialized")
+	}
+	queueSrv := ctx[bootstrap.BootstrappedQueueServer].(*queue.Server)
```
vedhavyas: let's do `,ok` check here as well

nft/ethereum_payment_obligation.go
```diff
@@ -176,7 +177,7 @@ func setupMintListener(config Config, tokenID *big.Int, registryAddress string)
 }
 
 // waitAndRouteNFTApprovedEvent notifies the confirmations channel whenever the key has been added to the identity and has been noted as Ethereum event
-func waitAndRouteNFTApprovedEvent(timeout time.Duration, asyncRes *gocelery.AsyncResult, tokenID *big.Int, confirmations chan<- *WatchTokenMinted) {
+func waitAndRouteNFTApprovedEvent(timeout time.Duration, asyncRes queue.TaskResult, tokenID *big.Int, confirmations chan<- *WatchTokenMinted) {
```
vedhavyas: `asyncRes` -> `taskResult` ?

node/bootstrapper.go
```diff
@@ -43,16 +43,20 @@ func (*Bootstrapper) Bootstrap(c map[string]interface{}) error {
...
+		return nil, fmt.Errorf("p2p server not initialized")
 	}
 
 	apiSrv, ok := ctx[bootstrap.BootstrappedAPIServer]
 	if !ok {
-		return nil, fmt.Errorf("API server not initiliase")
+		return nil, fmt.Errorf("API server not initialized")
+	}
+
+	queueSrv, ok := ctx[bootstrap.BootstrappedQueueServer]
```
vedhavyas:
```suggestion
	queueSrv, ok := ctx[bootstrap.BootstrappedQueueServer].(Server)
```

Can we change this to above two servers as well?

queue/bootstrapper.go
```diff
@@ -3,41 +3,21 @@ package queue
 import (
 	"errors"
 
+	"github.com/centrifuge/go-centrifuge/bootstrap"
 	"github.com/centrifuge/go-centrifuge/config"
 )
 
-const BootstrappedQueuedTasks string = "BootstrappedQueuedTasks"
-
 type Bootstrapper struct {
```
vedhavyas: Can we add tests for this bootstrapper?

queue/server.go
```diff
@@ -0,0 +1,121 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"github.com/centrifuge/gocelery"
+	logging "github.com/ipfs/go-log"
+)
+
+const (
+	BlockHeightParam string = "BlockHeight"
```
vedhavyas: Comment the exported constants?

queue/server.go
```diff
@@ -0,0 +1,121 @@
...
+	// GetNumWorkers gets the number of background workers to initiate
+	GetNumWorkers() int
+
+	// GetWorkerWaitTime gets the worker wait time for a task to be available while polling
+	// increasing this may slow down task execution while reducing it may consume a lot of CPU cycles
+	GetWorkerWaitTimeMS() int
+}
+
+// TaskType is a task to be queued in the centrifuge node to be completed asynchronously
+type TaskType interface {
```
vedhavyas: I'm not sure if the `TaskType` is the correct name for this since it only has one method that returns name of the task. 

I noticed we are not calling the interface method anywhere. Do we really need this ?
I would rather have us use `CelereyTask` interface which every task implements.

queue/server.go
```diff
@@ -0,0 +1,121 @@
...
+
+// TaskResult represents a result from a queued task execution
+type TaskResult interface {
+
+	// Get the result within a timeout from the queue task execution
+	Get(timeout time.Duration) (interface{}, error)
+}
+
+// Server represents the queue server currently implemented based on gocelery
+type Server struct {
```
vedhavyas: I think this can be unexported and a constructor could return the Server interface instead

queue/server.go
```diff
@@ -0,0 +1,121 @@
...
+		qs.queue.Register(task.TaskTypeName(), task)
+	}
+	// start the workers
+	qs.queue.StartWorker()
+	qs.lock.Unlock()
+
+	<-ctx.Done()
+	log.Info("Shutting down Queue server with context done")
+	qs.lock.Lock()
+	qs.queue.StopWorker()
```
vedhavyas: StopWorker is threadsafe. We dont need the lock here IMO

queue/server.go
```diff
@@ -0,0 +1,121 @@
...
+	qs.lock.Unlock()
+	log.Info("Queue server stopped")
+}
+
+// RegisterTaskType registers a task type on the queue server
+func (qs *Server) RegisterTaskType(name string, task interface{}) {
+	qs.taskTypes = append(qs.taskTypes, task.(TaskType))
```
vedhavyas: Lock is needed here since you are changing the state of qs.taskTypes.

queue/test_bootstrapper.go
```diff
@@ -2,11 +2,38 @@
...
+}
+
+func (s *Starter) TestBootstrap(ctx map[string]interface{}) error {
+	// handle the special case for running the queue server after task types have been registered (done by node bootstrapper at runtime)
+	qs := ctx[bootstrap.BootstrappedQueueServer].(*Server)
+	childErr := make(chan error)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	c, canF := context.WithCancel(context.Background())
+	s.canF = canF
```
vedhavyas: This should be protected by a Write lock

queue/test_bootstrapper.go
```diff
@@ -2,11 +2,38 @@
...
+	go qs.Start(c, &wg, childErr)
+	return nil
+}
+
+func (s *Starter) TestTearDown() error {
+	s.canF()
```
vedhavyas: This should be protected by read lock

queue/server.go
vimukthi-git: its needed because of the race detector complains about the `qs.queue` variable.

queue/server.go
vimukthi-git: no need as its not being used across threads because node init is single threaded. Besides race detector is enabled on our CI it would detect if there is an issue.

queue/test_bootstrapper.go
vimukthi-git: no need as its not being used across threads because node init is single threaded. Besides race detector is enabled on our CI it would detect if there is an issue.

queue/test_bootstrapper.go
vimukthi-git: no need as its not being used across threads because node init is single threaded. Besides race detector is enabled on our CI it would detect if there is an issue.

vimukthi-git:
@vedhavyas I made sure those are covered by integration tests. No need of mocked out unit tests that won't test them correctly anyway.

node/bootstrapper.go
vedhavyas: If the BootstrappedQueueServer is present with a type that doesn't implement server, then this will panic

queue/server.go
vedhavyas: Lets follow the standard :)

queue/server.go
vedhavyas: > no need as its not being used across threads because node init is single threaded. 
We cannot assume this will the be the case at a later time :)

vimukthi-git:
I don't mind at all, if there is a real provable concern, you can also take over the PR thanks 👍

vimukthi-git:
@vedhavyas as discussed and agreed we can address interface name and style in a later pr with queue/async stuff. I will merge this now.
Thank you!

PR #473, opened 2018-11-21
Remove usage of identity.IDService

vimukthi-git:


cmd/create_config.go
```diff
@@ -38,16 +38,16 @@ func generateKeys(config config.Config) {
 	keytools.GenerateSigningKeyPair(ethAuthPub, ethAuthPvt, "secp256k1")
 }
 
-func addKeys() error {
-	err := identity.IDService.AddKeyFromConfig(identity.KeyPurposeP2P)
+func addKeys(idService identity.Service) error {
```
vedhavyas: Why not make the function part of `identity.Service`? Even `createIdentity` ?

documents/purchaseorder/model_test.go
```diff
@@ -44,6 +44,7 @@ func TestMain(m *testing.M) {
 		&testlogging.TestLoggingBootstrapper{},
 		&config.Bootstrapper{},
 		&storage.Bootstrapper{},
+		&identity.Bootstrapper{},
```
vedhavyas: Can we remove pointer receiver for this bootstrap?

identity/ethereum_identity_integration_test.go
```diff
@@ -92,7 +92,7 @@ func TestAddKeyFromConfig(t *testing.T) {
 	assert.Nil(t, watchRegisteredIdentity.Error, "No error thrown by context")
 	assert.Equal(t, centrifugeId, watchRegisteredIdentity.Identity.CentID(), "Resulting Identity should have the same ID as the input")
 
-	err = identity.IDService.AddKeyFromConfig(identity.KeyPurposeEthMsgAuth)
+	err = identity.iDService.AddKeyFromConfig(identity.KeyPurposeEthMsgAuth)
```
vedhavyas: `iDService` -> `idService` ?

documents/invoice/model_test.go
```diff
@@ -44,6 +44,7 @@ func TestMain(m *testing.M) {
 		&testlogging.TestLoggingBootstrapper{},
 		&config.Bootstrapper{},
 		&storage.Bootstrapper{},
+		&identity.Bootstrapper{},
```
vedhavyas: Can we not have pointer receiver for this?

cmd/create_config.go
vimukthi-git: I dont want to do functional changes on this pr,

documents/purchaseorder/model_test.go
vimukthi-git: I default to pointer because that saves the overhead of copying for any purpose

documents/invoice/model_test.go
vimukthi-git: same comment as above, but lets discuss, agree and change this for all bootstrappers in a single pr later.

cmd/create_config.go
vimukthi-git: this would surely come in a later pr.

vimukthi-git: i take the above comment back, this is not part of business logic, its just a user utility doesn't make sense to be in a service.

documents/purchaseorder/model_test.go
vedhavyas: Pointers create overhead on Garbage collection since this will go on the heap instead of stack. It is recommended to avoid as much as we can unless the struct is so big that copying becomes an overhead.

PR #476, opened 2018-11-22
Pprof/endpoint

### LINKED ISSUE #474: Add pprof endpoints to the node
This will be useful to test the performance and bottlenecks of the node while it is running. It is recommended to run in production but we can make this optional through config.

Pprof should run on separate port and have an enable flag in the config

vedhavyas:
Fixes #474 

Changes:
- Add debug.pprof=false to default config
- if true, add pprof endpooints to client mux under `/debug/pprof/`


config/configuration.go
```diff
@@ -274,6 +274,11 @@ func (c *Configuration) GetEthAuthKeyPair() (pub, priv string) {
 	return c.GetString("keys.ethauth.publicKey"), c.GetString("keys.ethauth.privateKey")
 }
 
+// IsPProfEnabled returns true if the pprof is enabled
+func (c *Configuration) IsPProfEnabled() bool {
```
mikiquantum: To follow same convention as you started earlier should it be c.GetBool(key) and then add a method that does that casting?

PR #494, opened 2018-11-28
Fix cmds that were broken

vimukthi-git:
fixed and tested locally.

### DIFF
--- a/cmd/create_config.go
+++ b/cmd/create_config.go
@@ -82,18 +82,18 @@ func init() {
 
 			v, err := config.CreateConfigFile(data)
 			if err != nil {
-				panic(err)
+				log.Fatalf("error: %v", err)
 			}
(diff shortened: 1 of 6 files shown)


cmd/create_config.go
```diff
@@ -82,18 +82,18 @@ func init() {
 
 			v, err := config.CreateConfigFile(data)
 			if err != nil {
-				panic(err)
+				log.Fatalf("error: %v", err)
 			}
 			log.Infof("Config File Created: %s\n", v.ConfigFileUsed())
 
-			ctx := baseBootstrap(v.ConfigFileUsed())
+			ctx, canc, _ := commandBootstrap(v.ConfigFileUsed())
```
vedhavyas: let us not ignore the err here ?

vimukthi-git:
no idea why codecov complains about p2p/server.go

PR #512, opened 2018-11-28
Integration testing framework(Testworld) first iteration

vimukthi-git:


vimukthi-git:
@lucasvo @vedhavyas @mikiquantum assigning you guys to give you an idea of what it would be like. But this is not yet ready to be merged.

tests/peer.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/centrifuge/go-centrifuge/config"
+	ctx "github.com/centrifuge/go-centrifuge/context"
+	"github.com/centrifuge/go-centrifuge/documents/invoice"
+	"github.com/centrifuge/go-centrifuge/node"
+	logging "github.com/ipfs/go-log"
+)
+
+var log = logging.Logger("peer")
+
+type peer struct {
```
vimukthi-git: Peer abstraction for tests

tests/util.go
```diff
@@ -0,0 +1,75 @@
+// +build p2p_test
+
+package tests
+
+import (
```
vimukthi-git: utils to run shell scripts and initiate env like geth, smart contracts

testworld/park_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("contract addresses %+v\n", contractAddresses)
+	result := m.Run()
+	robert.stop()
+	os.Exit(result)
+}
+
+func TestHost_Happy(t *testing.T) {
```
vimukthi-git: this is the final version of the basic test

testworld/park_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"os"
+)
+
+// TODO remember to cleanup config files generated
+
+var robert *manager
+
+func TestMain(m *testing.M) {
+	// TODO start POA geth here
```
vimukthi-git: I will do this in a later pr

cmd/common.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+	canc()
+	db := ctx[storage.BootstrappedLevelDB].(*leveldb.DB)
+	db.Close()
+	return nil
+}
+
+// RunBootstrap bootstraps the node for running
+func RunBootstrap(cfgFile string) {
```
manuelpolzhofer: I think we could move the Bootstraps func in another package.

For me the`cmd` package should only take care of parsing commands etc. The actual execution logic/startup logic of a command should be in another place.

node/bootstrapper.go
```diff
@@ -15,8 +17,9 @@ type Bootstrapper struct{}
 // Bootstrap runs the severs.
```
manuelpolzhofer:
```suggestion
// Bootstrap runs the servers.
```

testworld/park.go
```diff
@@ -0,0 +1,248 @@
...
+	// network settings
+	ethNodeUrl, accountKeyPath, accountPassword, network string
+
+	txPoolAccess bool
+
+	// contractAddresses are the addresses of centrifuge contracts on Ethereum
+	contractAddresses *config.SmartContractAddresses
+
+	// bernard is the bootnode for all the hosts
+	bernard *host
```
manuelpolzhofer: Why not call it `mainBootNode` or similar?

testworld/park.go
```diff
@@ -0,0 +1,248 @@
...
+	name             string
+	apiPort, p2pPort int64
+}{
+	{"Alice", 8084, 38204},
+	{"Bob", 8085, 38205},
+	{"Charlie", 8086, 38206},
+}
+
+// manager is the manager of the hosts at Testworld (Robert)
+type manager struct {
```
manuelpolzhofer: Could we come up with a better name?

https://stackoverflow.com/questions/1866794/naming-classes-how-to-avoid-calling-everything-a-whatevermanager

testworld/util.go
```diff
@@ -0,0 +1,95 @@
...
+
+func getOpForContract(selector string) jq.Op {
+	addrOp, err := jq.Parse(selector)
+	if err != nil {
+		panic(err)
+	}
+	return addrOp
+}
+
+func getProjectDir() string {
```
manuelpolzhofer: Does golang not know which is the current directory?

testworld/park.go
```diff
@@ -0,0 +1,248 @@
...
+		network:           network,
+		txPoolAccess:      txPoolAccess,
+		contractAddresses: smartContractAddrs,
+		niceHosts:         make(map[string]*host),
+	}
+}
+
+func (r *manager) init() error {
+	cancCtx, canc := context.WithCancel(context.Background())
+	r.bernard = r.createHost("Bernard", 8081, 38201, nil)
```
manuelpolzhofer: Maybe own config variables for the different ports used for the bootNodes and other nodes?

testworld/util_test.go.test
```diff
@@ -0,0 +1,21 @@
+package tests
```
manuelpolzhofer: `util_test.go.test` ?

testworld/util.go
```diff
@@ -0,0 +1,95 @@
...
+	gp := os.Getenv("GOPATH")
+	projDir := path.Join(gp, "src", "github.com", "centrifuge", "go-centrifuge")
+	return projDir
+}
+
+// ToMap converts a struct to a map using the struct's tags.
+//
+// ToMap uses tags on struct fields to decide which fields to add to the
```
manuelpolzhofer:
```suggestion
// toMap uses tags on struct fields to decide which fields to add to the
```

testworld/util.go
```diff
@@ -0,0 +1,95 @@
...
+	return addrOp
+}
+
+func getProjectDir() string {
+	gp := os.Getenv("GOPATH")
+	projDir := path.Join(gp, "src", "github.com", "centrifuge", "go-centrifuge")
+	return projDir
+}
+
+// ToMap converts a struct to a map using the struct's tags.
```
manuelpolzhofer:
```suggestion

```

testworld/util.go
```diff
@@ -0,0 +1,95 @@
...
+//
```
manuelpolzhofer:
```suggestion

```

testworld/util.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+func getOpAddr(addrOp jq.Op, dat []byte) string {
+	addr, err := addrOp.Apply(dat)
+	if err != nil {
+		panic(err)
+	}
+
+	// remove annoying quotes
```
manuelpolzhofer: I would prefer more objective comments without adjectives related to feelings.

testworld/park.go
```diff
@@ -0,0 +1,248 @@
...
+	err := r.bernard.init()
+	if err != nil {
+		return err
+	}
+
+	// start and wait for Bernard since other hosts depend on him
+	go r.bernard.start(cancCtx)
+	time.Sleep(10 * time.Second)
```
manuelpolzhofer: I am not so deep into it but is there a better way to avoid a `time.sleep` using a channel inside the go routine for example?

testworld/keystore
```diff
@@ -0,0 +1 @@
+{"address":"89b0a86583c4444acfd71b463e0d3c55ae1412a5","crypto":{"cipher":"aes-128-ctr","ciphertext":"c779f8379d770d92cfc1ddd4a8f31d5a0adc8f2a0b2a1401370d3630f38c0c8a","cipherparams":{"iv":"36c168e73bf980fe75b0727f890a71ad"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"de1be16e3c981944d1eca2b8b27e4e6e0b5bfb43be0376a5d8889fa679a28122"},"mac":"cc128b815555ba1ead7cae9060e8842afca8356d06455bd9a5752ba6fcc092ef"},"id":"45e060a6-d2ae-43b8-922f-44829499d37d","version":3}
```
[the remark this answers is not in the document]
vimukthi-git: This is for now, next pr will include
- Loading this from a common location within the code, I don't want to have any implicit dependancy on env flags on tests as its hard to figure out whats going on. Although a further discussion is needed to agree on this, I like to remove all env variable dependancies in test suites, it will help us a lot in debugging issues going forward
- Running localgeth node if it hasn't started already
- Loading testworld configs from a actual config json so that we can for example switch networks at will

build/scripts/tests/run_testworld.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+echo "Running Testworld"
+
+status=$?
+for d in $(go list -tags=testworld ./... | grep -v vendor); do
+    output="go test -race -coverprofile=profile.out -covermode=atomic -tags=testworld $d 2>&1"
```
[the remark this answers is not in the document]
vimukthi-git: yup, the problem is that this runs on travis and verbose logging will make it harder to find actual issues

build/configs/default_config.yaml
```diff
@@ -21,8 +21,8 @@ networks:
     id: 51
     # Bootstrap list of nodes that Centrifuge provides to the russianhill testnet
     bootstrapPeers:
-      - "/ip4/35.225.200.42/tcp/38202/ipfs/12D3KooWLiicQVwThTBY6xKcPoLf6RQYJFpwf1r75wLx2ZR3pCd1"
-      - "/ip4/35.225.86.210/tcp/38202/ipfs/12D3KooWQZMA8GPHrvEZB9wdkoUcAAmCZHp9eyyZ4SE8gFr3hTNX"
+    - "/ip4/35.225.200.42/tcp/38202/ipfs/12D3KooWLiicQVwThTBY6xKcPoLf6RQYJFpwf1r75wLx2ZR3pCd1"
```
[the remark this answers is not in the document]
vimukthi-git: sure, I like that change, just that its should be a different pr.

build/scripts/migrate.sh
```diff
@@ -0,0 +1,40 @@
...
+# Let's not count on it and be clear instead
+if [ -z ${CENT_ETHEREUM_CONTRACTS_DIR} ]; then
+    CENT_ETHEREUM_CONTRACTS_DIR=${PARENT_DIR}/vendor/github.com/centrifuge/centrifuge-ethereum-contracts
+fi
+
+# Assure that all the dependencies for the contracts are installed
+npm install --pwd ${CENT_ETHEREUM_CONTRACTS_DIR} --prefix=${CENT_ETHEREUM_CONTRACTS_DIR}
+
+# `truffle migrate` will fail if not executed in the sub-dir
+cd ${CENT_ETHEREUM_CONTRACTS_DIR}
```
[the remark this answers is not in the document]
vimukthi-git: hmm, again a change that is not necessarily related to this pr, but you should create a ticket for this for sure.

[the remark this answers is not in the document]
vimukthi-git: this is just a refactoring of the existing script (if you didn't notice)

cmd/common.go
vimukthi-git: There is a constraint on where we can put this because, it can create cyclic dependancies very easily while creating many packages just because we like to see some functions in it seems kind of too much for my taste, but a pr is welcome if you find a suitable way to change this. Right now I'm just moving things so that I can use them in the `testworld`.

testworld/util.go
vimukthi-git: gonna remove this func

testworld/park.go
vimukthi-git: there is but its much more complicated that.. requires a major refactoring.

testworld/util.go
vimukthi-git: haha ok :)

testworld/park.go
vimukthi-git: this follows `Westworld` analogy ;)

testworld/park.go
vimukthi-git: I can make it `hostManager`

testworld/util_test.go.test
vimukthi-git: yup as the comment below explains

testworld/util.go
vimukthi-git: thats exactly what this function does. using `GOPATH` there is no other way as its specific to project structure and the location of compilation.

testworld/park.go
vimukthi-git: yup as I explained to @mikiquantum in another comment, thats in a next pr.

testworld/util.go
manuelpolzhofer: I thought of something simliar to `pwd`command.

In case the project name changes we would have to change this method as well

build/scripts/migrate.sh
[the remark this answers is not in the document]
manuelpolzhofer: I noticed only some overall ideas

PR #524, opened 2018-11-29
Remove storage.levelDBInstance global

vimukthi-git:


vimukthi-git:
@vedhavyas please note that the test coverage reduction is reported for removed code :P

PR #532, opened 2018-12-04
NFT: fixed incorrect tokenID size

manuelpolzhofer:
The tokenID we passed to the NFT contract was too big and just cut off.

max value of a uint256 is **2^256-1** we generated randomly **2^(256*8)-1**
We are generating 256 random bytes but we actually need 256 random bits.

It is interesting that we didn't get an error from the EVM or from the bindings before sending the transaction.

Links:
https://ethereum.stackexchange.com/questions/39321/what-is-the-maximum-input-value-for-function-uint256-parameter

### DIFF
--- a/nft/ethereum_payment_obligation.go
+++ b/nft/ethereum_payment_obligation.go
@@ -235,7 +235,11 @@ type MintRequest struct {
 
 // NewMintRequest converts the parameters and returns a struct with needed parameter for minting
 func NewMintRequest(to common.Address, anchorID anchors.AnchorID, proofs []*proofspb.Proof, rootHash [32]byte) (*MintRequest, error) {
-	tokenID := utils.ByteSliceToBigInt(utils.RandomSlice(256))
+
+	// tokenID is uint256 in Solidity (max value is 2^256-1)
+	// tokenID has 77 digits
+	// 1 byte = 8 bits
+	tokenID := utils.ByteSliceToBigInt(utils.RandomSlice(256/8))
 	tokenURI := "http:=//www.centrifuge.io/DUMMY_URI_SERVICE"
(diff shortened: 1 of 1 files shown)


nft/ethereum_payment_obligation.go
```diff
@@ -235,7 +235,11 @@ type MintRequest struct {
 
 // NewMintRequest converts the parameters and returns a struct with needed parameter for minting
 func NewMintRequest(to common.Address, anchorID anchors.AnchorID, proofs []*proofspb.Proof, rootHash [32]byte) (*MintRequest, error) {
-	tokenID := utils.ByteSliceToBigInt(utils.RandomSlice(256))
+
+	// tokenID is uint256 in Solidity (max value is 2^256-1)
+	// tokenID has 77 digits
+	// 1 byte = 8 bits
+	tokenID := utils.ByteSliceToBigInt(utils.RandomSlice(256/8))
```
mikiquantum: Why do we need to keep the `256` byte reference? If we know the type in solidity is uint256 (256 bits). Is there any reason why we cannot use `32` here?

mikiquantum: I found it a bit more confusing 😄 as the function name says: `ByteSliceToBigInt` already. Maybe `utils.RandomSlice` should be `utils.RandomByteSlice` instead.

manuelpolzhofer: I changed it to 32 and added a better comment

vimukthi-git:
@manuelpolzhofer is there no unit test that breaks because of this update, just curious?

manuelpolzhofer:
@vimukthi-git in go-centrifuge we are using bigInt and that didn't change.

PR #533, opened 2018-12-04
Fixing unnecessary for loops that I have written around selects

vimukthi-git:


vimukthi-git:
yup its a mistake that I kept it since earlier the idea was to have a continous stream of feedback. But I changed it and forgot to remove and copied all over the place 🤣

PR #535, opened 2018-12-05
Add update collaborator test

mikiquantum:


testworld/park_test.go
```diff
@@ -46,3 +46,70 @@ func TestHost_Happy(t *testing.T) {
...
+
+func TestHost_AddExternalCollaborator(t *testing.T) {
+	alice := doctorFord.getHost("Alice")
+	bob := doctorFord.getHost("Bob")
+	charlie := doctorFord.getHost("Charlie")
+	eAlice := alice.createHttpExpectation(t)
+	eBob := bob.createHttpExpectation(t)
+	eCharlie := charlie.createHttpExpectation(t)
+
+	a, err := alice.id()
```
vedhavyas: I just checked the definition

```
return h.identity.CentID(), nil
```

Why have `error` return when we are returning nil always @vimukthi-git ?

testworld/park_test.go
```diff
@@ -46,3 +46,70 @@ func TestHost_Happy(t *testing.T) {
...
+		"collaborators": []string{b.String()},
+	})
+	if err != nil {
+		t.Error(err)
+	}
+	docIdentifier := res.Value("header").Path("$.document_id").String().NotEmpty().Raw()
+	if docIdentifier == "" {
+		t.Error("docIdentifier empty")
+	}
+	getInvoiceAndCheck(eAlice, docIdentifier, "GBP")
```
vedhavyas: Can we pass in a map of keys and values instead of just currency here?

testworld/park_test.go
```diff
@@ -46,3 +46,70 @@ func TestHost_Happy(t *testing.T) {
 	getInvoiceAndCheck(eCharlie, docIdentifier, "GBP")
 	fmt.Println("Host test success")
 }
+
+func TestHost_AddExternalCollaborator(t *testing.T) {
```
vimukthi-git: `park_test.go` is a sanity check and an example on the test framework it self, please move the test to a more relevant file eg: `document_update_test.go` (just a suggestion)

testworld/park_test.go
vimukthi-git: its just to comply to the interface as it might change in the future inside the function, if will avoid nasty surprises.

testworld/park_test.go
mikiquantum: good to know

testworld/park_test.go
mikiquantum: I agree, I will add that to be more flexible in the future.

testworld/park_test.go
vimukthi-git: I should have added a comment, I know 😅

PR #539, opened 2018-12-06
integration test: added node time out test & improved test framework

manuelpolzhofer:
**test cases** 
- added node timout test
- improved existing test cases

**testworld**
- added status to create and update methods
- added canc func to hosts (to kill individual hosts)
- added hostTestSuite to manage all needed variables for a host

### DIFF
--- a/testworld/README.md
+++ b/testworld/README.md
@@ -17,5 +17,9 @@ Here you can create, run and test nodes with various behaviours to observe how t
 - Plus points if you write a test with a scenario that matches a scene in Westworld with node names matching the characters ;)
 
 
+### Speed improvements for testing
+- set `runMigrations` to `false` after the contracts are deployed once at the local geth node.
+- remove `cmd.CreateConfig(...)` in `park.go` after identities are already deployed once at the local geth node.
+
 
 
--- a/testworld/document_consensus_test.go
+++ b/testworld/document_consensus_test.go
@@ -75,3 +75,71 @@ func TestHost_AddExternalCollaborator(t *testing.T) {
 	getInvoiceAndCheck(eBob, params)
 	getInvoiceAndCheck(eCharlie, params)
 }
+
+func TestHost_CollaboratorTimeOut(t *testing.T) {
+
+	alice := doctorFord.getHost("Alice")
+	bob := doctorFord.getHost("Bob")
+	eAlice := alice.createHttpExpectation(t)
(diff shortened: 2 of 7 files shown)


testworld/park.go
```diff
@@ -248,3 +253,21 @@ func (h *host) p2pURL() (string, error) {
...
 }
+
+type hostTestSuite struct {
+	name   string
+	host   *host
+	id     identity.CentID
+	expect *httpexpect.Expect
+}
+
+func getHostTestSuite(t *testing.T, name string) hostTestSuite {
```
vimukthi-git: I would like to keep these kind of util functions that deal with host information as part of `hostManager`

testworld/payloads.go
```diff
@@ -0,0 +1,32 @@
+// +build testworld
```
vimukthi-git: 👍

testworld/document_consensus_test.go
```diff
@@ -2,76 +2,89 @@
...
+	// check if bob saved the updated document
+	params = map[string]interface{}{
+		"document_id": docIdentifier,
+		"currency":    "EUR",
 	}
-	params["currency"] = "USD"
-	getInvoiceAndCheck(eAlice, params)
-	getInvoiceAndCheck(eBob, params)
-	getInvoiceAndCheck(eCharlie, params)
+	getInvoiceAndCheck(bob.expect, params)
```
vimukthi-git: You need to restart any host that you bring down at the end of the test, otherwise other tests are affected.

vimukthi-git: Also lets create a new host that is explicitly meant to be brought down for tests if required. Since I want to make sure that we can run all tests concurrently in the future.

manuelpolzhofer: yes, you are right. I will add a own host for that

testworld/README.md
```diff
@@ -17,5 +17,9 @@ Here you can create, run and test nodes with various behaviours to observe how t
 - Plus points if you write a test with a scenario that matches a scene in Westworld with node names matching the characters ;)
 
 
+### Speed improvements for testing
+- set `runMigrations` to `false` after the contracts are deployed once at the local geth node.
+- remove `cmd.CreateConfig(...)` in `park.go` after identities are already deployed once at the local geth node.
```
vimukthi-git: needs to emphasise that this is only for local testing ;)

PR #543, opened 2018-12-06
Errors/internal

### LINKED ISSUE #537: Create `errors` library so that we can start using it for internal error handling
Refer - Spec section https://hackmd.io/f0WqseIITIqaqRrmYBgy1w#Internal-Node-errors

## Acceptance criteria
- Implement following set of functions as outlined by the spec
```
func New(err error, args ...interface{}) error
func NewListError(olderr, newerr) error
func Len(err) int 
func NewTypedError(type, err error) error
func OfType(type, err error) bool
```
- Unit tests to make sure that each functions works

vedhavyas:
Fixes #537 

Note: This only adds internal errors. We need to add client and peerError later

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -61,12 +61,11 @@ func getErrs(err error) []error {
 // append errn errors to err and return err
 func AppendError(err, errn error) error {
 	var errs []error
-	if serrs := getErrs(err); len(serrs) > 0 {
-		errs = append(errs, serrs...)
-	}
 
-	if serrs := getErrs(errn); len(serrs) > 0 {
-		errs = append(errs, serrs...)
+	for _, e := range []error{err, errn} {
+		if serrs := getErrs(e); len(serrs) > 0 {
+			errs = append(errs, serrs...)
+		}
 	}
 
 	if len(errs) < 1 {
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -1,7 +1,6 @@
 package errors
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 )
@@ -10,7 +9,7 @@ import (
 const ErrUnknown = Error("unknown error")
 
 // Error is a string that implements error
-// this will have interesting side effects of having a constant errors
+// this will have interesting side effects of having constant errors
 type Error string
 
 // Error returns error message
@@ -25,10 +24,10 @@ func (e Error) Error() string {
 // New("some error: %v", "some context") returns error with message "some error: some context"
 func New(msg string, args ...interface{}) error {
 	if len(args) > 0 {
-		return fmt.Errorf(msg, args...)
+		return Error(fmt.Sprintf(msg, args...))
 	}
 
-	return errors.New(msg)
+	return Error(msg)
 }
 
 // listError holds a list of errors
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -22,12 +22,12 @@ func (e Error) Error() string {
 // Example:
 // New("some error") returns error with exact string passed
 // New("some error: %v", "some context") returns error with message "some error: some context"
-func New(msg string, args ...interface{}) error {
+func New(format string, args ...interface{}) error {
 	if len(args) > 0 {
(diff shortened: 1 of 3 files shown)


utils/events.go
```diff
@@ -1,9 +1,11 @@
 package utils
 
-import "fmt"
+import (
+	"github.com/centrifuge/go-centrifuge/errors"
+)
 
 // ErrEventNotFound when event is not found and need to retry
-var ErrEventNotFound = fmt.Errorf("event not found")
+var ErrEventNotFound = errors.New("event not found")
```
vimukthi-git: `const` please

errors/errors.go
```diff
@@ -0,0 +1,139 @@
+package errors
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// ErrUnknown is an unknown error type
+var ErrUnknown = New("unknown error")
```
vimukthi-git: `const` please

errors/errors_test.go
```diff
@@ -0,0 +1,113 @@
+// +build unit
+
+package errors
```
vimukthi-git: 👍

errors/errors.go
```diff
@@ -0,0 +1,139 @@
...
+	return &typeError{terr: terr, ctxErr: err}
+}
+
+// TypeError can be implemented by any type error
+type TypeError interface {
+	IsOfType(terr error) bool
+}
+
+// isOfType returns if the err t is of type terr
+func (t *typeError) IsOfType(terr error) bool {
```
vimukthi-git: isOfType => IsOfType

errors/errors.go
vedhavyas: `const` wont work for error types :)

vedhavyas: It should be a type not an interface for constants. We can work this around but i'm thinking a better solution

vedhavyas: Pushed `const` changes

vimukthi-git: i had the solution to this in the spec, as you may see

errors/errors.go
```diff
@@ -0,0 +1,147 @@
...
+func (e Error) Error() string {
+	return string(e)
+}
+
+// New returns a new error with message passed.
+// if args are passed, we will format the message with args
+// Example:
+// New("some error") returns error with exact string passed
+// New("some error: %v", "some context") returns error with message "some error: some context"
+func New(msg string, args ...interface{}) error {
```
mikiquantum: Can we enforce here the format? Meaning:
```
err := New("some error", "some context")
```
And it is the `New` function that formats it:
```
...
if len(args) > 0 {
		return Error(fmt.Sprintf("%s: %s", msg, args...))
}
```
That way we have the standard built in.

errors/errors.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Error returns the error in string
+func (t *typeError) Error() string {
+	return fmt.Sprintf("%v: %v", t.terr, t.ctxErr)
+}
+
+// NewTypeError returns a new error of type typeError
+func NewTypeError(terr, err error) error {
+	if terr == nil {
```
mikiquantum: `terror` lol

errors/errors.go
vedhavyas: Just noticed! awesome

errors/errors.go
vedhavyas: We can enforce it but leaves with multiple functions like std library. Thinking having a single function can be easier.

mikiquantum: But we are trying to standardize our usage of errors right? 
In addition, looking at the `msg string` argument, doesn't tell me if it is just a string or a format string, so either we make it clear in the arg name and doc or we enforce it, IMHO.

vedhavyas: Doc says it already. Not sure what could be the name of arg can be. Do you have a suggestion for arg name?

vedhavyas: Will update the arg to format as per the std lib

PR #546, opened 2018-12-07
Testworld: detect when host is live instead of waiting arbitrarily

vimukthi-git:
Fixing flakyness

### DIFF
--- a/testworld/park.go
+++ b/testworld/park.go
@@ -34,6 +34,14 @@ var hostConfig = []struct {
 	{"Kenny", 8087, 38207},
 }
 
+// hostTestSuite encapsulates test utilities on top of each host
+type hostTestSuite struct {
+	name       string
+	host       *host
+	id         identity.CentID
+	httpExpect *httpexpect.Expect
+}
+
 // hostManager is the hostManager of the hosts at Testworld (Robert)
 type hostManager struct {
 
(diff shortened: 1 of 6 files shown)


testworld/park.go
```diff
@@ -264,6 +282,42 @@ func (h *host) start(c context.Context) error {
 
 }
 
+func (h *host) kill() {
+	h.canc()
+}
+
+func (h *host) isLive(softTimeOut time.Duration) (bool, error) {
```
vimukthi-git: this is the main change

manuelpolzhofer:
```suggestion
func (h *host) isLive(timeOut int) (bool, error) {
```

manuelpolzhofer: Why a soft and a hard time out? Why not just one hard time out passed by the user?

manuelpolzhofer: something like:
```golang
func (h *host) isLive(timeOut int) (bool, error) {	
sig := make(chan bool)
	c := createInsecureClient()
	go func(sig chan<- bool) {
		var fErr error
		for i := 0; i < timeOut; i++ {
			res, err := c.Get(fmt.Sprintf("https://localhost:%d/ping", h.config.GetServerPort()))
			fErr = err
			if err != nil {
				time.Sleep(time.Second)
				continue
			}
			if res.StatusCode == http.StatusOK {
				sig <- true
				return
			}
		}
		sig <- false
	}(sig)
	return <-sig:
	
```

manuelpolzhofer: In current implementation the soft timeout is higher than the hard timeout :)

vimukthi-git: Only reason the soft timeout exists is to give the person who writes the test to exit earlier since he doesn't care about host being alive. Right now the only place we use it I have set the soft timeout higher so that hard timeout is used. But someone else can use it in a different way.

PR #551, opened 2018-12-10
ISSUE and PR templates proposal

vimukthi-git:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+## Shiny new `go-centrifuge` issue !!!
```
vedhavyas: Let's not add `Shiny` :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+## Shiny new `go-centrifuge` PR !!!
```
vedhavyas: Same as above :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
...
+
+This is intended as a guide when describing/reviewing pull requests for `go-centrifuge` or related projects
+
+- [ ] Issue link
+- [ ] The solution implemented
+- [ ] If this is only part of the solution what is that part?
```
vedhavyas: Not sure if we really need point 2 and 3 IMO

.github/ISSUE_TEMPLATE.md
vimukthi-git: This is not something you should use, just a replaceable title.

.github/PULL_REQUEST_TEMPLATE.md
vimukthi-git: same here

.github/PULL_REQUEST_TEMPLATE.md
vimukthi-git: its needed as its sometimes not obvious what the fix is and also which part of the fix it is. As described this is just a guide not a requirement.

.github/ISSUE_TEMPLATE.md
vimukthi-git: I will put a <replace> tag

manuelpolzhofer: I would prefer maybe a simpler template at the current stage.

Are there any existing templates, we could simple re-use?

vimukthi-git: Its the simplest I could think of, and its based on some existing templates I found on the internet :) 
`Simpler` is subjective, please make specific suggestions on how you see it can be simplified.

Issue #504, opened 2018-11-28
Update the Invoice Service to use new Repository


vedhavyas:
1. Invoice service should register Model and use new Repository

PR #554, opened 2018-12-11
 Migrate invoice service to new repo

vimukthi-git:
#504 pr 2


documents/bootstrapper_test.go
```diff
@@ -5,11 +5,14 @@ package documents
...
 	"testing"
 
+	"github.com/centrifuge/go-centrifuge/storage"
 	"github.com/stretchr/testify/assert"
+	"github.com/syndtr/goleveldb/leveldb"
 )
 
 func TestBootstrapper_Bootstrap(t *testing.T) {
 	ctx := make(map[string]interface{})
+	ctx[storage.BootstrappedLevelDB] = &leveldb.DB{}
```
vedhavyas:
```suggestion
	ctx[storage.BootstrappedLevelDB] = new(leveldb.DB)
```

documents/error.go
```diff
@@ -8,6 +8,12 @@ import (
 
 const (
 
+	// ErrDocumentConfigTenantID must be used for errors related to tenantID operations
+	ErrDocumentConfigTenantID = errors.Error("error with tenantID operations")
```
vedhavyas: This seems very generic IMO

documents/error.go
```diff
@@ -49,6 +55,26 @@ const (
...
+	ErrDocumentRepositorySerialisation = errors.Error("document repository encountered a marshalling error")
+
+	// ErrDocumentRepositoryModelNotFound must be used when document repository can not locate the given model
+	ErrDocumentRepositoryModelNotFound = errors.Error("document repository could not locate the given model")
+
+	// ErrDocumentRepositoryModelSave must be used when document repository can not save the given model
+	ErrDocumentRepositoryModelSave = errors.Error("document repository could not save the given model")
+
+	// ErrDocumentRepositoryModelAllReadyExists must be used when document repository finds an already existing model when saving
+	ErrDocumentRepositoryModelAllReadyExists = errors.Error("document repository found an already existing model when saving")
```
vedhavyas: `ErrDocumentRepositoryModelAllReadyExists` => `ErrRepositoryModelExists `

vedhavyas: `document repository found an already existing model when saving` => `model already exists in the repository`

documents/invoice/service.go
```diff
@@ -150,8 +151,14 @@ func (s service) calculateDataRoot(old, new documents.Model, validator documents
 		return nil, errors.NewTypedError(documents.ErrDocumentInvalid, err)
 	}
 
+	// get tenant ID
+	tenantID, err := s.config.GetIdentityID()
```
vedhavyas: lets explicitly pass tenentID through function args. Makes it easier later for multi tenency

documents/invoice/service.go
```diff
@@ -166,14 +173,24 @@ func (s service) Create(ctx *header.ContextHeader, inv documents.Model) (documen
...
 		return nil, err
 	}
 
 	return inv, nil
 }
 
+// updater wraps logic related to updating documents so that it can be executed as a closure
+func (s service) updater(id []byte, model documents.Model) error {
+	// get tenant ID
+	tenantID, err := s.config.GetIdentityID()
```
vedhavyas: Same as above. Use a closure instead pls

documents/invoice/service.go
```diff
@@ -380,7 +407,13 @@ func (s service) ReceiveAnchoredDocument(model documents.Model, headers *p2ppb.C
 		return errors.NewTypedError(documents.ErrDocumentPackingCoreDocument, err)
 	}
 
-	err = s.repo.Update(doc.CurrentVersion, model)
+	// get tenant ID
+	tenantID, err := s.config.GetIdentityID()
```
vedhavyas: fetch the tenentID from the `header.ContextHeader`

documents/service.go
```diff
@@ -7,6 +7,13 @@ import (
 	"github.com/centrifuge/precise-proofs/proofs/proto"
 )
 
+// Config specified configs required by documents package
+type Config interface {
```
vedhavyas: We dont need this once we fetch the tenetID from `header.ContextHeader`

p2p/handler_integration_test.go
```diff
@@ -74,7 +74,7 @@ func TestHandler_RequestDocumentSignature_AlreadyExists(t *testing.T) {
 	req = getSignatureRequest(doc)
 	resp, err = handler.RequestDocumentSignature(context.Background(), req)
 	assert.NotNil(t, err, "must not be nil")
-	assert.Contains(t, err.Error(), "document already exists")
+	assert.Contains(t, err.Error(), documents.ErrDocumentRepositoryModelAllReadyExists.Error())
```
vedhavyas: Use `errors.IsTypeOf`

p2p/handler_integration_test.go
```diff
@@ -147,7 +147,7 @@ func TestHandler_SendAnchoredDocument_update_fail(t *testing.T) {
 	anchorReq := getAnchoredRequest(doc)
 	anchorResp, err := handler.SendAnchoredDocument(context.Background(), anchorReq)
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "document doesn't exists")
+	assert.Contains(t, err.Error(), documents.ErrDocumentRepositoryModelDoesntExist.Error())
```
vedhavyas: same as above

vimukthi-git:
@vedhavyas 
- Based on the explanation given offline the model we assume in multi tenancy is based on the config per tenant, so the implementation is according to the spec.
- About errors I think lets not make it subjective how we decide how to implement errors, you will never know which error needs to be handled in which way, I think we can go with the list of error types here and decide later if we need to remove them(this is a compromise that I don't like, just so that we can move on)
I put a 👍 on stuff I will change.

documents/bootstrapper.go
```diff
@@ -1,13 +1,28 @@
...
+import (
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/storage"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
 // BootstrappedRegistry is the key to ServiceRegistry in Bootstrap context
 const BootstrappedRegistry = "BootstrappedRegistry"
 
+// BootstrappedDocumentRepository is the key to the database repository of documents
```
mikiquantum: Can we bundle them in same block:
```
const (
  BootstrappedRegistry = "BootstrappedRegistry"
  BootstrappedDocumentRepository = "BootstrappedDocumentRepository"
)
```

documents/error.go
```diff
@@ -8,6 +8,12 @@ import (
 
 const (
 
+	// ErrDocumentConfigTenantID must be used for errors related to tenantID operations
```
mikiquantum: Aiming to have this PR merged asap, is it possible to take this error part out into another one? As it will bring up more questions as shown below. Or @vedhavyas we are ok keeping this err definition, and change it later?

documents/invoice/bootstrapper.go
```diff
@@ -51,8 +44,14 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return fmt.Errorf("identity service not initialised")
 	}
 
+	repo, ok := ctx[documents.BootstrappedDocumentRepository].(documents.Repository)
+	if !ok {
+		return fmt.Errorf("document db repository not initialised")
+	}
+	repo.Register(&Invoice{})
```
mikiquantum: Is it then recommended to use `new` to reserve memory for an object? Instead of above. Just trying to get clarity for further work and set up an standard.

vimukthi-git: I'm agnostic to this at least in this occasion as registry only uses this object to derive the type using reflection. Its basically a type representation.

documents/error.go
vimukthi-git: - We discussed error types
- Agreed on an implementation
- Now we have the implementation
- I'm using it, and now its about which place to use it. My preference is not to make things so subjective and flexible so that everyone chooses whether to add an error type or not based on personal judgement. By default we go with error types and we decide later when we see a particular error type not being used much and remove, its an easy refactoring.

mikiquantum: @vedhavyas ^^ I am ok either way.

vimukthi-git:
I will address the agreed 👍 comments on a different smaller pr so as to not block anything else.

PR #555, opened 2018-12-11
added integration tests for NFT

manuelpolzhofer:
- added integration tests for NFT
- added new utils method to testworld


testworld/nft_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const tokenIdLength = 77
+
+func TestPaymentObligationMint_successful(t *testing.T) {
```
vimukthi-git: 💯

vimukthi-git:
One thing we should build a utility for is to verify transactions on geth it self and see the transaction actually went through and the data is correct, but thats for another time.

PR #556, opened 2018-12-11
added testworld tests for proofs

manuelpolzhofer:


testworld/proof_test.go
```diff
@@ -0,0 +1,38 @@
...
+		t.Error(err)
+	}
+
+	docIdentifier := getDocumentIdentifier(t, res)
+	if docIdentifier == "" {
+		t.Error("docIdentifier empty")
+	}
+
+	// Alice want's to get a proof
+	proofPayload := map[string]interface{}{
```
vimukthi-git: please move this to the payloads.go you created.
also get a proof from bobs node and match with Alices.

PR #559, opened 2018-12-11
 Remove fmt.Errorf in favour of errors.New

vimukthi-git:


ethereum/geth_client_test.go
```diff
@@ -46,12 +45,12 @@ type MockTransactionRequest struct {
 func (transactionRequest *MockTransactionRequest) RegisterTransaction(opts *bind.TransactOpts, transactionName string, anotherVar string) (tx *types.Transaction, err error) {
 	transactionRequest.count++
 	if transactionName == "otherError" {
-		err = errors.Wrap("Some other error", 1)
+		err = errors.New("Some other error")
 	} else if transactionName == "optimisticLockingTimeout" {
-		err = errors.Wrap(transactionUnderpriced, 1)
+		err = transactionUnderpriced
```
manuelpolzhofer: Why not `error.New` here?

vimukthi-git: its already an error constant, no need to do errors.New if you are using error constants, in most cases.

PR #570, opened 2018-12-13
Skip timeout tests on CI to avoid resource problems

vimukthi-git:


build/scripts/tests/run_testworld.sh
```diff
@@ -4,7 +4,7 @@ echo "Running Testworld"
 
 status=$?
 
-output="go test -race -parallel 4 -coverprofile=profile.out -covermode=atomic -tags=testworld github.com/centrifuge/go-centrifuge/testworld 2>&1"
```
vimukthi-git: Number of parallel tests at a time will be decided by go automatically based on resource availability.

PR #571, opened 2018-12-14
go version 1.11.3 bug, lock to 1.11.2

mikiquantum:


mikiquantum:
https://github.com/golang/go/issues/29241

PR #572, opened 2018-12-14
NFT: using new repository

manuelpolzhofer:
**changes:**
- removed find service from the registry
- added repository to NFT service
- adjusted unit tests
- testworld corrected PO payloads

**old flow:**
- ask registry to search service based on docIdentifier

**new flow:**
- repository return correct model
- get coredocument from model
- use coredocument type to recieve correct service

**question:**
the service is currently needed for the proof generation.
Maybe we should refactor it? The current flow is a bit too complicated in my opinion.


nft/ethereum_payment_obligation.go
```diff
@@ -66,12 +70,13 @@ func newEthereumPaymentObligation(registry *documents.ServiceRegistry, identityS
 }
 
 func (s *ethereumPaymentObligation) prepareMintRequest(documentID []byte, depositAddress string, proofFields []string) (*MintRequest, error) {
-	docService, err := s.registry.FindService(documentID)
+
```
vimukthi-git: This function looks very weird to me, 
typically in layered architecture you need to execute biz logic you go in this call sequence
`controller->service->(biz logic+ model)-> repo` (note repo is accessible only through service)
here its like 
`controller->repo->model->service->repo`

There has to be a better option than directly accessing the repository from another package, which creates issue in design for future, someone can argue for not having any services at all and say "let's just deal with repositories everywhere". There is a clear reason why we have service a layer which is to encapsulate all business logic(such as model updates, permissions, anchor logic etc.). We must not forgo that convention.

The solution I see here is to pull up the proof creation logic into a base service in the `documents` package and call that service here. That service can be passed around instead of the repo.
So the only call here would be,
`proofs, err := baseDocService.CreateProofs(documentID, proofFields)`

manuelpolzhofer: I completely agree. I am still thinking of a better way.

The problem is the following:
- we having specific abi endpoints for invoice and po obviously, you know the service. service can handle all the biz logic
- we designed our entire node based on that assumption
- nft is generic and we only require a documentID. maybe we should still pass the documentType for NFT do not break the logic

vimukthi-git: As explained the `baseDocService` would solve the problem. But we can discuss more if we have to :)

PR #573, opened 2018-12-17
move to 1.11.4

vedhavyas:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.11.2
+  - 1.11.x
```
mikiquantum: I am a bit reluctant on using the "latest" minor here, as in the previous release 1.11.3 a bug was added that affected us directly and it took some time to figure out what the problem was. I would lock the version to 1.11.4, and have periodic upgrades on demand

PR #574, opened 2018-12-17
Consolidate DB Repository

mikiquantum:


storage/repository.go
```diff
@@ -0,0 +1,32 @@
+package storage
+
+import (
+	"reflect"
+)
+
+// Model is an interface to abstract away storage model specificness
+type Model interface {
+	// Get the ID of the document represented by this model
+	ID() ([]byte, error)
```
vedhavyas: Dont think we need `ID` as part of Model Interface IMO

PR #575, opened 2018-12-17
HTTP header for tenantID

### LINKED ISSUE #488: [API Server] Include HTTP/gRPC Header for TenantID
- Read auth header from request (make sure it works as well for gRPC)

vedhavyas:
Fixes #488 

1. Add interceptor that require header 'authorization' which holds tenantID
2. Add basic version if HTTPError and http Error response interceptor since we ned it for point 1


api/server.go
```diff
@@ -11,16 +12,27 @@ import (
...
 	logging "github.com/ipfs/go-log"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
 )
 
-var log = logging.Logger("api-server")
+// ErrNoAuthHeader used for requests when header is not passed.
+const ErrNoAuthHeader = errors.Error("'authorization' header missing")
```
mikiquantum: Should we move this to its own error.go file within the api package?

api/server.go
```diff
@@ -165,3 +182,66 @@ func loadKeyPair() (keyPair tls.Certificate, err error) {
 	}
 	return pair, nil
 }
+
+// grpcInterceptor returns a GRPC UnaryInterceptor for all grpc/http requests.
+func grpcInterceptor() grpc.ServerOption {
+	return grpc.UnaryInterceptor(auth)
+}
+
+// auth is the grpc unary interceptor to to check if the tenent ID is passed in the header.
```
mikiquantum: `tenant ID`

api/server.go
```diff
@@ -11,16 +12,27 @@ import (
...
+const ErrNoAuthHeader = errors.Error("'authorization' header missing")
+
+var (
+	log = logging.Logger("api-server")
+
+	// noAuthPaths holds the paths that doesn't require header to be passed.
+	noAuthPaths = [...]string{"/health.HealthCheckService/Ping"}
+
+	// TenantKey represents the key used to fetch the tenant id from context
+	TenantKey struct{}
```
mikiquantum: Wondering if the TenantKey should go to an `auth` package. Reason is, the ctx will be pass through to each handler, which will load the tenant info.

api/server.go
```diff
@@ -59,10 +71,15 @@ func (c apiServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr cha
...
 	})
 
-	opts := []grpc.ServerOption{grpc.Creds(creds)}
+	// set http error interceptor
+	runtime.HTTPError = httpResponseInterceptor
 
-	grpcServer := grpc.NewServer(opts...)
+	opts := []grpc.ServerOption{
+		grpc.Creds(creds),
+		// grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
```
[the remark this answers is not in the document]
vimukthi-git: I would add it rather than have a many commented lines of code in the code base.
These are called feature flags. you can read about it at https://martinfowler.com/articles/feature-toggles.html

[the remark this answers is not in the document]
vedhavyas: Feature flags are used to enable features, not for integrations which is what i'm talking about here. Feature flags which enable/disable a feature should not break the application flow, whereas this one does. I would rather keep it commented rather than putting it behind a flag which enables makes the application error out for every API call. Hence, I would like to keep this commented as is and enable it when our API require tenantID to be passed.

We have used feature flags earlier like `pprof` but this one doesn't belong under feature flag IMO.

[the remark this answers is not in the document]
vedhavyas: Let's take it offline if you have any more concerns 👍

api/server.go
mikiquantum: ok

Issue #577, opened 2018-12-18
adding document service implementation for proofs


manuelpolzhofer:
the NFT service should not access the repository directly instead an implementation of the document service should offer the needed methods for the proof generation.

See discussion in PR #572

PR #579, opened 2018-12-18
added a generic document service

manuelpolzhofer:
Issue Link: #577 

This is a first PR to implement a document service.
Document service will not be used yet.

currently implemented + tests:
`GetVersion` and `GetCurrentVersion`

next PR: `CreateProofs` and `CreateProofForVersion`

questions: I think some methods of the document.Service interface will be moved to the specific ones.


documents/service.go
```diff
@@ -49,3 +53,107 @@ type Service interface {
...
+	config Config
+	repo   Repository
+}
+
+// DefaultService returns the default implementation of the service
+func DefaultService(config config.Configuration, repo Repository) Service {
+	return service{repo: repo, config: config}
+}
+
+func getIds(model Model) ([]byte, []byte, error) {
```
vedhavyas: `getIDs`

documents/service.go
```diff
@@ -49,3 +53,107 @@ type Service interface {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return cd.DocumentIdentifier, cd.NextVersion, nil
+}
+
+func (s service) searchVersion(m Model) (Model, error) {
+
+	id, next, err := getIds(m)
```
vedhavyas: handle err here pls

documents/service.go
```diff
@@ -49,3 +53,107 @@ type Service interface {
...
+
+	nm, err := s.getVersion(id, next)
+
+	if err != nil {
+		// here the err is returned as nil because it is expected that the nextVersion
+		// is not available in the db at some stage of the iteration
+		return m, nil
+	}
+
+	if next != nil {
```
vedhavyas: Next is always set, we dont need this check IMO

documents/service.go
```diff
@@ -49,3 +53,107 @@ type Service interface {
...
+
+	return nm, nil
+}
+
+func (s service) GetCurrentVersion(documentID []byte) (Model, error) {
+	model, err := s.getVersion(documentID, documentID)
+	if err != nil {
+		return nil, errors.NewTypedError(ErrDocumentNotFound, err)
+	}
+	return s.searchVersion(model)
```
vedhavyas: I think getversion and searchVersion names are used interchangeably here.

documents/service.go
```diff
@@ -49,3 +53,107 @@ type Service interface {
...
+
+func (s service) RequestDocumentSignature(contextHeader *header.ContextHeader, model Model) (*coredocumentpb.Signature, error) {
+	return nil, nil
+}
+
+func (s service) ReceiveAnchoredDocument(model Model, headers *p2ppb.CentrifugeHeader) error {
+	return nil
+}
+
+func (s service) Exists(documentID []byte) bool {
```
vedhavyas: We dont need this anymore, right ?

manuelpolzhofer: Yes, right. We can remove it or leave it since we already have it. 
Maybe we need it in the future.
If we keep it, i will add some tests

documents/service.go
manuelpolzhofer: That's true. Otherwise, it would not be possible to create a nextVersion without modifying the older version. 

Currently, it is just some additional error handling. Something is strange with nextVersion let's  return the actual.

documents/service.go
manuelpolzhofer: searchVersion -> recursive linked list get last element
getVersion -> read document from DB.

I will rename it.

documents/service.go
manuelpolzhofer: I am using it now for getting the currentVersion.
It's more elegant to use an exist instead of waiting for a db error to find the latest version

PR #582, opened 2018-12-19
Transactions

### LINKED ISSUE #495: Define and Implement TransactionRepository
Define TransactionRepository interface and a LevelDB implementation.

vedhavyas:
Fixes #495 

1. Add Transaction Repository
2. Add Transaction Definitions

### DIFF
--- a/transactions/repository.go
+++ b/transactions/repository.go
@@ -1,0 +1,78 @@
+package transactions
+
+import (
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/storage"
+	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/satori/go.uuid"
+)
+
+const (
+	// ErrTransactionMissing error when transaction doesn't exist in Repository.
+	ErrTransactionMissing = errors.Error("transaction doesn't exist")
+
+	// ErrKeyConstructionFailed error when the key construction failed.
+	ErrKeyConstructionFailed = errors.Error("failed to construct transaction key")
+)
+
+// Repository can be implemented by a type that handles storage for transactions.
+type Repository interface {
+	Get(identity common.Address, id uuid.UUID) (*Transaction, error)
+	Save(transaction *Transaction) error
+}
+
+// txRepository implements Repository.
+type txRepository struct {
+	repo storage.Repository
+}
+
+// NewRepository registers the the transaction model and returns the an implementation
+// of the Repository.
+func NewRepository(repo storage.Repository) Repository {
+	repo.Register(&Transaction{})
+	return &txRepository{repo: repo}
+}
+
+// getKey appends identity with id.
+// With identity coming at first, we can even fetch transactions belonging to specific identity through prefix.
+func getKey(identity common.Address, id uuid.UUID) ([]byte, error) {
+	if utils.IsEmptyAddress(identity) {
+		return nil, errors.New("identity cannot be empty")
+	}
+
+	if uuid.Equal(uuid.Nil, id) {
+		return nil, errors.New("transaction ID is not valid")
+	}
+
+	return append(identity[:], id.Bytes()...), nil
+}
+
+// Get returns the transaction associated with identity and id.
+func (r *txRepository) Get(identity common.Address, id uuid.UUID) (*Transaction, error) {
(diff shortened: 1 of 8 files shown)


transactions/repository.go
```diff
@@ -0,0 +1,78 @@
...
+// NewRepository registers the the transaction model and returns the an implementation
+// of the Repository.
+func NewRepository(repo storage.Repository) Repository {
+	repo.Register(&Transaction{})
+	return &txRepository{repo: repo}
+}
+
+// getKey appends identity with id.
+// With identity coming at first, we can even fetch transactions belonging to specific identity through prefix.
+func getKey(identity common.Address, id uuid.UUID) ([]byte, error) {
```
vimukthi-git: I know we would use Ethereum addresses for identities everywhere soon, but I prefer to make the change 
from
`type CentID [len]byte`
to
`type CentID common.Address` in `identities` package
than use `common.Address` in these places, it's a helpful abstraction and this kind of change is exactly the reason why we have it. Let's just use `CentID` type here for all function signatures.

transactions/transaction.go
```diff
@@ -0,0 +1,72 @@
...
+		Action:  action,
+		Message: message,
+		Time:    time.Now().UTC(),
+	}
+}
+
+// Transaction contains details of transaction.
+type Transaction struct {
+	ID          uuid.UUID
+	Identity    common.Address
```
vimukthi-git: Same here

vedhavyas:
> exposing a repository directly is also a concern but for transations it doesn't matter so much as it currently doesn't have any business logic. But my personal preference is to have a wrapped service for allowing any future evolution.

@vimukthi-git the general idea where the repository is used are 2 places, one in side the services such as invoice and PO services and other through a `Status` API. We will be using a handler and service pattern for that API as well. So adding a service to this is just another wrapper.

transactions/repository.go
vedhavyas: We are purposefully not calling it CentID but rather identity in our epics so that it is not specific. Maybe @lucasvo or @manuelpolzhofer can explain more?

manuelpolzhofer: We decided to name it identityAddress or identity instead.

I would suggest: DID (Decentralized Identifier)

```golang
type DID common.Address
```
I think keeping a specific type a good idea.

transactions/repository.go
```diff
@@ -0,0 +1,78 @@
+package transactions
+
+import (
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/storage"
+	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/satori/go.uuid"
```
manuelpolzhofer: What is the benefit of using a generic uuid?

Why not something specific with addtitional information.

For example: `concat(identityAddress,sha3(transactionCounter))`

transactions/repository.go
```diff
@@ -0,0 +1,78 @@
...
+	ErrTransactionMissing = errors.Error("transaction doesn't exist")
+
+	// ErrKeyConstructionFailed error when the key construction failed.
+	ErrKeyConstructionFailed = errors.Error("failed to construct transaction key")
+)
+
+// Repository can be implemented by a type that handles storage for transactions.
+type Repository interface {
+	Get(identity common.Address, id uuid.UUID) (*Transaction, error)
+	Save(transaction *Transaction) error
```
manuelpolzhofer: The db supports only a `Put(key,value)`.  we build a `create` and `update` on the lowest level around it.
Here we build a new `put(key, value)` around on a higher level.

transactions/repository.go
manuelpolzhofer: I just notice on the db level we storing the uuid anyway together with the identityAddress

transactions/repository.go
vimukthi-git: lets do the rename later, but lets use CentID for now.

transactions/repository.go
vedhavyas: Just to make things easier, We dont need to support create and update specifically here since we will using it to update logs

transactions/repository.go
vedhavyas: Storing uuid alone is preferable but if we want tenant specific transactions, its going to be inefficient

Hence, we prefix it with identity so that its easier to fetch tenant specific txns

PR #585, opened 2018-12-20
Tenant Struct

vimukthi-git:
Issue Link: <Github link>

### DIFF
--- a/context/bootstrapper.go
+++ b/context/bootstrapper.go
@@ -63,3 +63,37 @@ func (m *MainBootstrapper) Bootstrap(context map[string]interface{}) error {
 	}
 	return nil
 }
+
+// TenantBootstrapper ensures all dependencies for a tenant is initialised correctly
+type TenantBootstrapper struct {
+	IOCContext    map[string]interface{}
+	Bootstrappers []bootstrap.Bootstrapper
+}
+
+func (t *TenantBootstrapper) Populate() {
+	t.Bootstrappers = []bootstrap.Bootstrapper{
+		documents.Bootstrapper{},
+		p2p.Bootstrapper{},
+		&invoice.Bootstrapper{},
+		&purchaseorder.Bootstrapper{},
+		&nft.Bootstrapper{},
+	}
+}
+
+// Bootstrap runs all the loaded bootstrapper implementations.
+// The context passed has to be the same one used for the MainBootstrapper otherwise this will error out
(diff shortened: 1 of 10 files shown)


context/bootstrapper.go
```diff
@@ -63,3 +63,37 @@ func (m *MainBootstrapper) Bootstrap(context map[string]interface{}) error {
...
 	return nil
 }
+
+// TenantBootstrapper ensures all dependencies for a tenant is initialised correctly
+type TenantBootstrapper struct {
+	IOCContext    map[string]interface{}
+	Bootstrappers []bootstrap.Bootstrapper
+}
+
+func (t *TenantBootstrapper) Populate() {
```
mikiquantum: Why don't keep them static?

multitenant/bootstrapper.go
```diff
@@ -0,0 +1,53 @@
+package multitenant
```
mikiquantum: I rather not have a different package for multitenancy, as it will be the default setup either if there is 1 tenant or 100+. We can leave it this way for now, but I believe it will go "potentially" to node package or similar.

multitenant/tenant.go
```diff
@@ -0,0 +1,103 @@
...
+	// Name is the unique name given to the service within the Cent Node
+	Name() string
+
+	// Start starts the service, expectation is that this would always be called in a separate go routine.
+	// WaitGroup contract must always be honoured by calling `defer wg.Done()`
+	Start(ctx context.Context, wg *sync.WaitGroup, startupErr chan<- error)
+}
+
+// tenant is a tenant in the multi tenant centrifuge node
+type tenant struct {
```
mikiquantum: Here we will have to add the services exposed in future PRs right?

multitenant/bootstrapper.go
vimukthi-git: Its a limitation in bootstrapping stuff, otherwise it creates dependancy cycles, I also would have preferred to have this in the node.

context/bootstrapper.go
vimukthi-git: since we might need to change the list dynamically and call bootstrap. Also it follows the same style as main bootstrapper just to keep things consistent. We can refactor later if needed.

multitenant/tenant.go
vimukthi-git: the services are in the bootstrapped context per tenant. so its already in there as you may see.

PR #596, opened 2018-12-21
Transaction Status Check API

### LINKED ISSUE #496: Add Status Check Endpoint
Transaction status check endpoint of the form `/transaction/{centrifugeID}/{txID}` should pick the transaction from the repository and return the latest status of the Transaction.

vedhavyas:
Fixes #496 

- Add Transaction Status check endpoint


common/common.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "github.com/ethereum/go-ethereum/common"
+
+var (
+	// TenantKey is used as key for the tenant identity in the context.ContextWithValue.
+	TenantKey struct{}
+
+	// DummyIdentity to be used until we have identity coming from auth header
+	// TODO(ved): get rid of this once we have multitenancy enabled/
+	DummyIdentity = common.Address([20]byte{1})
```
vedhavyas: This will be used until we start fetching identity from HTTP calls. This should not break any existing APIs and helps us continue to add multi-tenancy code

vimukthi-git: I hope we would remove this package with it as well..

PR #597, opened 2018-12-23
Remove P2P GRPC

vimukthi-git:
Issue Link: <Github link>

today I tried something and it was easier than I thought, removing grpc
- It makes it much easier to build clients, we can also easily build a JS based browser client just based on libp2p and protobufs.
- greatly simplifies centrifuge wireprotocol,
- Weird hacks are not needed to multiplex streams to individual tenants - check p2p/receiver/handler.go - which is what triggered me to think of this in the first place


p2p/messenger.go
```diff
@@ -0,0 +1,374 @@
+package p2p
```
vimukthi-git: this is what replaces GRPC stuff

vimukthi-git: adapted from https://github.com/libp2p/go-libp2p-kad-dht

protobufs/protocol/protocol.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
```
vimukthi-git: this is an envelope for all p2p protobufs so that they can be identified and decoded.

vimukthi-git: needs to be moved to centrifuge-protobufs , its here for the sake of hackability.

p2p/receiver/handler.go
```diff
@@ -84,8 +109,28 @@ func (srv Handler) RequestDocumentSignature(ctx context.Context, sigReq *p2ppb.S
 	}, nil
 }
 
+// HandleSendAnchoredDocument handles the SendAnchoredDocument message
+func (srv *Handler) HandleSendAnchoredDocument(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *pb.P2PEnvelope) (*pb.P2PEnvelope, error) {
```
vimukthi-git: note the protocol id access :)

p2p/client.go
```diff
@@ -16,36 +19,63 @@ import (
...
+func (s *p2pServer) SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error) {
+	pid, err := s.getPeerID(id)
+	if err != nil {
+		return nil, err
+	}
+	adr, err := proto.Marshal(in)
+	if err != nil {
+		return nil, err
+	}
+	recv, err := s.mes.sendRequest(ctx, pid, &protocolpb.P2PEnvelope{Type: protocolpb.MessageType_MESSAGE_TYPE_SEND_ANCHORED_DOC, Body: adr}, CentrifugeProtocol)
```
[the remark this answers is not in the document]
vimukthi-git: About RPC:
yes it looks like an RPC system for ease of use(You could also call it distributed event handling :D ), but we have the freedom to do the wire protocol the way we want instead of dealing with http2 framing and other stuff that GRPC does.
I will fix the naming its confusing here, adr is not an address.

p2p/client.go
```diff
@@ -16,36 +19,63 @@ import (
...
+	// after all signatures are collected the sender sends the document including the signatures
+	SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error)
+}
+
+func (s *p2pServer) SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error) {
+	pid, err := s.getPeerID(id)
+	if err != nil {
+		return nil, err
+	}
+	adr, err := proto.Marshal(in)
```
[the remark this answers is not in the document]
vimukthi-git: its the marshalled anchor response, I will fix naming

p2p/client.go
```diff
@@ -16,36 +19,63 @@ import (
...
-	OpenClient(id identity.Identity) (p2ppb.P2PServiceClient, error)
+
+	// GetSignaturesForDocument gets the signatures for document
 	GetSignaturesForDocument(ctx *header.ContextHeader, identityService identity.Service, doc *coredocumentpb.CoreDocument) error
+
+	// after all signatures are collected the sender sends the document including the signatures
+	SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error)
+}
+
+func (s *p2pServer) SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error) {
```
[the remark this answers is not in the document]
vimukthi-git: `Client` is not a struct, it's an interface of ease of use, testing and mocking.

protobufs/gen/swagger/protocol/protocol.swagger.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
vimukthi-git: @lucasvo do you know a way to set swagger generation to `false` for `protocol.proto`?

p2p/client.go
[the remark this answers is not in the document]
vimukthi-git: its defined in the p2p/server.go file, its an existing struct, it implements the `Client` interface so that we can reuse the same struct with its internal values to support both p2p client and server modes. Which makes a lot of sense in a p2p node IMO. But this is not something I wrote, it was there before.

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+	"sync"
+	"time"
+
+	"github.com/centrifuge/go-centrifuge/errors"
+
+	"github.com/libp2p/go-libp2p-host"
+	"github.com/libp2p/go-libp2p-protocol"
+
+	pb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/protocol"
+	ggio "github.com/gogo/protobuf/io"
```
vedhavyas: Let's stick one protobuf implementation else this will cause unneeded compatibility issues later

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+
+	timout time.Duration
+	ctx    context.Context
+
+	strmap map[peer.ID]map[protocol.ID]*messageSender
+	smlk   sync.Mutex
+
+	plk sync.Mutex
+
+	msgHandlers map[pb.MessageType]func(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *pb.P2PEnvelope) (*pb.P2PEnvelope, error)
```
vedhavyas: Lets go with simple handler instead. I can see all handlers having the same boilerplate code of marshalling and unmarshalling. Instead, this can be done by p2p with all marshaling and unmarshalling

protobufs/protocol/protocol.proto
```diff
@@ -0,0 +1,24 @@
...
+option java_outer_classname = "ProtocolProto";
+option java_package = "com.protocol";
+
+enum MessageType {
+  MESSAGE_TYPE_INVALID = 0;
+  MESSAGE_TYPE_REQUEST_SIGNATURE = 1;
+  MESSAGE_TYPE_REQUEST_SIGNATURE_REP = 2;
+  MESSAGE_TYPE_SEND_ANCHORED_DOC = 3;
+  MESSAGE_TYPE_SEND_ANCHORED_DOC_REP = 4;
+  MESSAGE_TYPE_ERROR = 5;
```
vedhavyas: Invalid and Error shouldn't be part of the Type. They are unexpected errors.

IMO, we should not tie the Type in the protocol itself.

p2p/messenger.go
lucasvo: Just talked to Ved, to clarify I think this makes sense:

Unmarshal the embedded message in `P2PEnvelope` to the right protobuf message and pass it in. This would remove a lot of redundant code for unmarshaling & error checking and would allow the handler to just do one type assertion.

```suggestion
 	msgHandlers map[pb.MessageType]func(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *proto.Message) (*pb.P2PEnvelope, error)
```

p2p/client.go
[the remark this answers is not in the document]
lucasvo: maybe we should call it something other than p2pServer but p2pNode?

coredocument/processor_test.go
```diff
@@ -49,7 +49,7 @@ func TestDefaultProcessor_PrepareForSignatureRequests(t *testing.T) {
 	// pack failed
 	model := mockModel{}
 	model.On("PackCoreDocument").Return(nil, errors.New("error")).Once()
-	ctxh, err := context2.NewContextHeader(context.Background(), cfg)
+	ctxh, err := context2.NewHeader(context.Background(), cfg)
```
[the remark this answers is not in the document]
vimukthi-git: this has been changed already in the develop branch :)

p2p/client.go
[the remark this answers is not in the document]
vimukthi-git: sure, we could do that.

vimukthi-git:
@lucasvo @mikiquantum told me he would look at this today at the latest on DM.

p2p/messenger.go
vedhavyas: What I'm suggesting is not centrifuge specific but rather generic one as we did for Repository interface. Once we integrate this and we decide to make it a library and add the boiler plate code to the p2p itself, we have to refactor all the handlers which can be completely avoided IMO

protobufs/protocol/protocol.proto
vedhavyas: That is where we use error in the as part of the Message. IMO Error or invalid Type are not same as `MESSAGE_TYPE_SEND_ANCHORED_DOC`.

p2p/messenger.go
vimukthi-git: I clarified my comment above with a diagram @vedhavyas @lucasvo since you guys don't seem to have understood what I said.

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func (mes *p2pMessenger) handleNewMessage(s inet.Stream) {
+	ctx := mes.ctx
+	// ok to use. we defer close stream in this func
+	cr := ctxio.NewReader(ctx, s)
+	cw := ctxio.NewWriter(ctx, s)
+
+	// delimited readers and writers to set length of the protobuf messages to the stream
+	r := ggio.NewDelimitedReader(cr, inet.MessageSizeMax)
```
mikiquantum: This value `MessageSizeMax` I assume is the batch reading size and does not mean that any part of a proto object bigger than that will be discarded otherwise it is a problem.

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+	w := newBufferedDelimitedWriter(cw)
+	mPeer := s.Conn().RemotePeer()
```
mikiquantum: How are you planning on resolving the receiver/local peer to identify the request?

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+	msgHandlers map[pb.MessageType]func(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *pb.P2PEnvelope) (*pb.P2PEnvelope, error)
+}
+
+func newP2PMessenger(ctx context.Context, host host.Host, p2pTimeout time.Duration) *p2pMessenger {
+	return &p2pMessenger{
+		ctx:         ctx,
+		host:        host,
+		self:        host.ID(),
+		timout:      p2pTimeout,
+		strmap:      make(map[peer.ID]map[protocol.ID]*messageSender),
```
mikiquantum: I believe keeping streams open is interesting but uncertain when scaling up. I like the concept, but I would  have loved to simplify this approach at the moment. Let's keep it now that is implemented.

p2p/messenger.go
vimukthi-git: that would not get to the messenger, it would be through the `p2p/client.go`. We check if the centID is a tenant in the system using configstore and call local handler functions with the centID in the context.

p2p/client.go
```diff
@@ -4,9 +4,13 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/protocol"
```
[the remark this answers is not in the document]
vimukthi-git: it ran through the linter without an issue

p2p/client.go
```diff
@@ -17,36 +21,69 @@ import (
...
+	err = proto.Unmarshal(recv.Body, r)
+	if err != nil {
+		return nil, err
+	}
+	return r, nil
 }
 
 // OpenClient returns P2PServiceClient to contact the remote peer
-func (s *p2pServer) OpenClient(id identity.Identity) (p2ppb.P2PServiceClient, error) {
+func (s *p2pServer) getPeerID(id identity.Identity) (peer.ID, error) {
```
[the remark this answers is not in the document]
vimukthi-git: I would like to avoid that kind of dependancies between layers, but we can think of how to optimise.

p2p/client.go
```diff
@@ -58,54 +95,39 @@ func (s *p2pServer) OpenClient(id identity.Identity) (p2ppb.P2PServiceClient, er
...
-		return nil, errors.New("failed to dial peer [%s]: %v", peerID.Pretty(), err)
-	}
-
-	return p2ppb.NewP2PServiceClient(g), nil
+	return peerID, nil
 }
 
 // getSignatureForDocument requests the target node to sign the document
-func (s *p2pServer) getSignatureForDocument(ctx context.Context, identityService identity.Service, doc coredocumentpb.CoreDocument, client p2ppb.P2PServiceClient, receiverCentID identity.CentID) (*p2ppb.SignatureResponse, error) {
+func (s *p2pServer) getSignatureForDocument(ctx context.Context, identityService identity.Service, doc coredocumentpb.CoreDocument, receiverPeer peer.ID, receiverCentID identity.CentID) (*p2ppb.SignatureResponse, error) {
```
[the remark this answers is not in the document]
vimukthi-git: subjective :) I like receiver since its more network/communication terminology

p2p/messenger.go
```diff
@@ -0,0 +1,342 @@
...
+	"time"
+
+	"github.com/centrifuge/go-centrifuge/errors"
+
+	"github.com/libp2p/go-libp2p-host"
+	"github.com/libp2p/go-libp2p-protocol"
+
+	pb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/protocol"
+	ggio "github.com/gogo/protobuf/io"
+	"github.com/jbenet/go-context/io"
```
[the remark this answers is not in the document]
vimukthi-git: I think its useful, even if we create it we would anyway have a separate package for the IO structs.

p2p/messenger.go
vimukthi-git: I also thought about this, it does discard right now, but we can increase it arbitrarily or even remove it by forking ggio reader and writer, so it should be fine.

vimukthi-git:
@mikiquantum I don't think we should get rid of grpc-gateway right now, that doesn't affect our protocol and it works. But I do think its just too much baggage to carry just to add API endpoints. 
The idea was to move on as well as fix structural changes. I think the limitations in GRPC particularly for multi tenancy was a blocker for me. I also had to fix merge conflicts against my self multiple times when I work on other stuff. It just seemed the right thing to merge this and move on.
Lets discuss and create issues for stuff that is left in the next grooming when you are here.
But I do believe as any other change/feature this would have some bugs and issues to be ironed out overtime. I don't expect this to be the last pr in making the p2p layer nicer to work with.

Issue #499, opened 2018-11-28
Add NFT Task to use the new Transaction model


vedhavyas:
NFT process should be moved to a new tasks that will follow Transaction model

Issue #498, opened 2018-11-28
Move Invoice and PO anchoring to New Anchor Tasks


vedhavyas:
Update invoice and service to use the new anchoring tasks and deprecate the old anchoring process

PR #598, opened 2018-12-24
Async APIs

### LINKED ISSUE #497: Add Anchoring Task with the Transaction model
Define a new tasks with multi step anchoring and use the Transaction repository.

vedhavyas:
Fixes #497 #498 #499 

I started using commom.Address for tenant id since that is a requirement for transactions.
Note: This will break the functional tests.


transactions/base_task.go
```diff
@@ -0,0 +1,53 @@
...
+	b.TxID, err = uuid.FromString(txID)
+	if err != nil {
+		return errors.New("invalid transaction ID")
+	}
+
+	return nil
+}
+
+// UpdateTransaction add a new log and updates the status of the transaction based on the error.
+func (b *BaseTask) UpdateTransaction(tenantID common.Address, name string, err error) error {
```
vimukthi-git: Transaction service function

testworld/httputils.go
```diff
@@ -115,3 +124,19 @@ func createInsecureClient() *http.Client {
 	}
 	return &http.Client{Transport: tr}
 }
+
+func waitTillSuccess(t *testing.T, e *httpexpect.Expect, txID string) {
```
vimukthi-git: busy loop, lets add a timer

queue/server.go
```diff
@@ -120,3 +120,8 @@ func ParseBlockHeight(valMap map[string]interface{}) (uint64, error) {
 	}
 	return 0, errors.New("value can not be parsed")
 }
+
+// TaskQueuer can be implemented by any queueing system
+type TaskQueuer interface {
```
vimukthi-git: 👍

nft/payment_obligation_integration_test.go
```diff
@@ -66,21 +69,42 @@ func TestPaymentObligationService_mint(t *testing.T) {
...
 	assert.Nil(t, err, "should not error out when minting an invoice")
-	tokenConfirm := <-confirmations
-	assert.Nil(t, tokenConfirm.Err, "should not error out when minting an invoice")
-	assert.NotNil(t, tokenConfirm.TokenID, "token id should be present")
+	assert.NotNil(t, resp.TokenID, "token id should be present")
+}
+
+func waitTillSuccessOrError(txID uuid.UUID, txService transactions.Service) (bool, error) {
+	status := transactions.Pending
+	for status == transactions.Pending {
```
vimukthi-git: this should be a function in the `txService`
also please add a timer instead of a busy loop

nft/ethereum_payment_obligation.go
```diff
@@ -122,51 +130,46 @@ func (s *ethereumPaymentObligation) MintNFT(documentID []byte, registryAddress,
...
 	if err != nil {
-		return nil, err
+		return txID, err
 	}
-	asyncRes, err := qs.EnqueueJob(mintingConfirmationTaskName, map[string]interface{}{
-		tokenIDParam:           hex.EncodeToString(tokenID.Bytes()),
-		queue.BlockHeightParam: h.Number.Uint64(),
-		registryAddressParam:   registryAddress,
-	})
+	tx := transactions.NewTransaction(tenantID, "Mint NFT")
```
vimukthi-git: following should be part of `txService.createTransaction`,
```
tx := transactions.NewTransaction(tenantID, "Mint NFT")
err = s.txRepository.Save(tx)
 		queue.BlockHeightParam: h.Number.Uint64(),	
 		registryAddressParam:   registryAddress,	
 	})	
 	if err != nil {	 		 		
           return txID, err
 	}	 	
```

documents/purchaseorder/bootstrapper.go
```diff
@@ -49,8 +43,18 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	}
 	repo.Register(&PurchaseOrder{})
 
+	queueSrv, ok := ctx[bootstrap.BootstrappedQueueServer].(*queue.Server)
+	if !ok {
+		return errors.New("queue server not initialised")
+	}
+
+	txRepo, ok := ctx[transactions.BootstrappedRepo].(transactions.Repository)
```
vimukthi-git: lets just use the txService interface on other packages instead of repo, please. This just breaks that pattern we are creating else where plus there are multiple places in this pr where txService is useful.

protobufs/transactions/service.proto
```diff
@@ -25,7 +25,7 @@ message TransactionStatusResponse {
 service TransactionService {
   rpc GetTransactionStatus(TransactionStatusRequest) returns (TransactionStatusResponse) {
     option (google.api.http) = {
-      get: "/transaction/{transaction_id}"
+      get: "/transactions/{transaction_id}"
```
vimukthi-git: 👍

documents/anchor_task.go
```diff
@@ -0,0 +1,138 @@
...
+
+const (
+	modelIDParam           = "modelID"
+	tenantIDParam          = "tenantID"
+	documentAnchorTaskName = "Document Anchoring"
+)
+
+var log = logging.Logger("anchor_task")
+
+type documentAnchorTask struct {
```
vimukthi-git: 👍

testworld/httputils.go
lucasvo: This is for testworld, right?

vedhavyas: Yeap for testworld. but can add timer

vimukthi-git: yes

documents/purchaseorder/bootstrapper.go
vedhavyas: I think calling it txRepo was a bad idea. but the idea was that we have too many wrappers services which will be behind the other services. Hence, I decided to use it as it is. This will not be breaking any patterns because of following reasons
1. tx repository is behind invoice/po service like the rest of the repositories
2. We still follow transactions handler -> service pattern for rest apis

I would like to move the transaction wait state to transactions packages so that we can re-use the code

nft/payment_obligation_integration_test.go
vedhavyas: This is test and shouldn't be of an issue in prod IMO

PR #601, opened 2018-12-27
 Convert ContextHeader to Context and make a util

vimukthi-git:
The ContextHeader structure might lead to race conditions in following way,

Thread 0: `origin := NewContextHeader(typedVal: val, ctx : someStandardCTX)`

Thread 1: Uses `origin` and wants to set a timeout to the context, but also want to keep using the `origin` because of the `typeVal`, so it creates a context using `origin.ctx` and does `origin.ctx = ctxwithtimeout`

Thread2: Uses `origin` and wants to set a timeout to the context, but also want to keep using the `origin` because of the `typeVal`, so it creates a context using `origin.ctx` and does `origin.ctx = ctxwithtimeout`

So both `Thread 1` and `Thread 2` endsup having to modify the `origin`, but this leads to a race condition. We can of course use a mutex and solve the issue, but that would lead to us rewriting the go `context` package with some typedVals we want, IMO this is not worth the effort, when we can go with the simple approach of defining a standard context with our own values and parsing them using some standard util functions, which is what I'm doing in this pr.


vedhavyas:
@vimukthi-git context header was supposed to be used in a single request context and multiple goroutines should not have access to the same context header. I dont think we used shared it between multiple go routines.

The problem with using context.Context is that we are losing static typing and end up bloating handlers/services with uneeded error checks IMO

vimukthi-git:
@vedhavyas Thanks for the reply
I'm aware of the type safety issue and I read about this much more after our discussions before on context package and this implementation is a careful compromise between having type safety(runtime time type safety) vs thread(go routine) safety. Following are the benefits,
1. All function signatures would be consistent with most other go projects and standard errors would be returned when the contexts passes don't contain values needed. I value consistency.
2. All functions that use the context can be assumed thread(go routine) safe vs making them all unsafe by usage of the context header
3. Since we changed our multi tenancy approach to one where we use context value to decide the tenant(one you suggested as well) instead of tenant struct having this context passable everywhere even across go routine boundaries becomes important eg: for p2p requests, although you could argue we can use queuing for that, I would say having the freedom to choose among standard language features such as go routines is important.
4. Error check bloat can't be avoided unfortunately but it also makes sure that the code is safe, and error checks are converted into much nicer standard `contextutil` functions here.

vedhavyas:
Sure, let's be consistent. but I'm not sure what you mean by context header not being thread safe since we will not be mutating it but rather reading from it @vimukthi

vimukthi-git:
I already explained above how it could happen and this exact scenario would happen once multi tenancy is implemented in p2p requests. Eg: When we want to set a timeout to the embedded context while also passing the header. Which you might argue that we can duplicate the header, but it would mean that chain of contexts is ignored. You could also argue that we can set a mutex for the embedded context but then which context should the receiving threads honour, also more importantly it means that we would reinvent context package bit by bit. You could also argue we wouldn't need to pass the header to the go routines just pass the embedded context but then how would we pass the tenant configs if needed? then we might need to add the configs to embedded context and pass it then whats the use of having the header in the first place? Like this we can go on arguing but as pointed out above best to use the available tool which `context.Context` and make do while we can unless we face an extreme scenario where we can't.

contextutil/context.go
```diff
@@ -0,0 +1,35 @@
+package contextutil
```
mikiquantum: I believe adding upper layer packages dependencies here will lead to an overhead of cyclic dependencies. Instead of having each package "know what look for". 
We can discuss more about it next week.

vimukthi-git: yes, Im already away of this issue, particular the identity package has an issue, because of the use of identity here. My idea is we should use only the ID bytes in the context.

Issue #508, opened 2018-11-28
[Config DB] Migrate to DB Config Model


mikiquantum:
Migration Plan:
1. ~Read from config file and store in DB (tenant + node). Keep config file with current structure.~
2. ~When loading tenant, read from DB instead.~
3. ~When reading node config, read from DB instead~

PR #602, opened 2018-12-27
[MultiTenancy] Enable config database and api

vimukthi-git:
Issue Link: #508

### DIFF
--- a/api/service.go
+++ b/api/service.go
@@ -3,12 +3,14 @@ package api
 import (
 	"github.com/centrifuge/go-centrifuge/bootstrap"
 	"github.com/centrifuge/go-centrifuge/config"
+	"github.com/centrifuge/go-centrifuge/config/configstore"
 	"github.com/centrifuge/go-centrifuge/documents"
 	"github.com/centrifuge/go-centrifuge/documents/invoice"
 	"github.com/centrifuge/go-centrifuge/documents/purchaseorder"
 	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-centrifuge/healthcheck"
 	"github.com/centrifuge/go-centrifuge/nft"
+	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/config"
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/documents"
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/health"
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/invoice"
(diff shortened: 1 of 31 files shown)


api/service.go
```diff
@@ -79,12 +81,24 @@ func registerServices(ctx context.Context, cfg Config, grpcServer *grpc.Server,
...
 
+	// nft api
 	nftpb.RegisterNFTServiceServer(grpcServer, nft.GRPCHandler(payObService))
 	err = nftpb.RegisterNFTServiceHandlerFromEndpoint(ctx, gwmux, addr, dopts)
 	if err != nil {
 		return err
 	}
 
+	// config api
+	configService, ok := nodeObjReg[configstore.BootstrappedConfigStorage].(configstore.Service)
```
vimukthi-git: main change

bootstrap/bootstrappers/bootstrapper.go
```diff
@@ -33,6 +34,7 @@ func (m *MainBootstrapper) PopulateBaseBootstrappers() {
 		&version.Bootstrapper{},
 		&config.Bootstrapper{},
 		&storage.Bootstrapper{},
+		&configstore.Bootstrapper{},
```
vimukthi-git: main change

bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
```diff
@@ -26,6 +27,7 @@ var bootstappers = []bootstrap.TestBootstrapper{
 	&testlogging.TestLoggingBootstrapper{},
 	&config.Bootstrapper{},
 	&storage.Bootstrapper{},
+	&configstore.Bootstrapper{},
```
vimukthi-git: main change

config/configstore/bootstrapper.go
```diff
@@ -0,0 +1,54 @@
+package configstore
```
vimukthi-git: main change

protobufs/config/service.proto
```diff
@@ -32,7 +32,7 @@ service ConfigService {
   }
   rpc GetAllTenants(google.protobuf.Empty) returns (GetAllTenantResponse) {
     option (google.api.http) = {
-      get: "/config/tenant/list"
+      get: "/config/tenant"
```
vimukthi-git: bugfix

testworld/config_test.go
```diff
@@ -0,0 +1,29 @@
+// +build testworld
```
mikiquantum: From my point of view, this doesn't belong to testworld. Regular integration test and/or functional tests.

protobufs/config/service.proto
mikiquantum: IMO we need a plural verb here, either what it was or `/config/tenants`

config/configstore/bootstrapper.go
mikiquantum: Looks good, thanks!

testworld/config_test.go
vimukthi-git: regular tests wouldn't be sufficient, considering the bug I fixed with urls
functional tests can be done, but this is much more convenient and also verifies things for each pr.

vimukthi-git: also the utils created here are useful for Testworld tests that we do for multi tenancy in the future

protobufs/config/service.proto
vimukthi-git: agree, it needs to change for all endpoints per rest conventions.

PR #603, opened 2018-12-28
[MultiTenancy] Use db stored config at runtime

vimukthi-git:
Issue Link: #508 
Review #602 first


anchors/anchor.go
```diff
@@ -34,7 +36,7 @@ type AnchorID [AnchorIDLength]byte
 type Config interface {
 	GetEthereumDefaultAccountName() string
 	GetEthereumContextWaitTimeout() time.Duration
-	GetContractAddress(address string) common.Address
+	GetContractAddress(contractName config.ContractName) common.Address
```
vedhavyas: do we need a new type for this?

build/scripts/tests/run_testworld.sh
```diff
@@ -8,7 +8,7 @@ eval "$cleanup"
 
 status=$?
 
-output="go test -race -coverprofile=profile.out -covermode=atomic -tags=testworld github.com/centrifuge/go-centrifuge/testworld 2>&1"
+output="go test -coverprofile=profile.out -covermode=atomic -tags=testworld github.com/centrifuge/go-centrifuge/testworld 2>&1"
```
vedhavyas: not running with race conditions anymore ?

anchors/anchor.go
vimukthi-git: I kept running into problems where the contracts weren't initialized properly, this prevents that.

build/scripts/tests/run_testworld.sh
vimukthi-git: I would love to have this, just that it requires around 3gb memory and travis just keeps timing out randomly because of that.

vimukthi-git: reverted, it worked this time.

vedhavyas: Nice!

PR #609, opened 2019-01-06
Multi tenancy service parts complete without api handlers and p2p handlers

vimukthi-git:
Issue Link: <Github link>


documents/anchor_task.go
```diff
@@ -32,10 +32,10 @@ type documentAnchorTask struct {
 	transactions.BaseTask
 
 	id       []byte
-	tenantID common.Address
+	tenantID identity.CentID
 
 	// state
-	config        config.Configuration
+	config        configstore.Service
```
mikiquantum: For asynchronous tasks I rather pass static values instead of loading them dynamically when the message reaches the consumer. It adds up more predictability to the queueing/dequeuing flow.

documents/purchaseorder/handler.go
```diff
@@ -25,12 +20,11 @@ var apiLog = logging.Logger("purchaseorder-api")
...
 type grpcHandler struct {
 	service Service
-	// TODO [multi-tenancy] replace this with config service
-	config config.Configuration
+	config  configstore.Service
 }
 
 // GRPCHandler returns an implementation of the purchaseorder DocumentServiceServer
-func GRPCHandler(config config.Configuration, registry *documents.ServiceRegistry) (clientpurchaseorderpb.DocumentServiceServer, error) {
+func GRPCHandler(config configstore.Service, registry *documents.ServiceRegistry) (clientpurchaseorderpb.DocumentServiceServer, error) {
```
mikiquantum: We should start thinking on renaming this 😄

contextutil/context.go
```diff
@@ -32,3 +36,26 @@ func Self(ctx context.Context) (*identity.IDConfig, error) {
...
+func Tenant(ctx context.Context) (*configstore.TenantConfig, error) {
+	tc, ok := ctx.Value(self).(*configstore.TenantConfig)
+	if !ok {
+		return nil, ErrSelfNotFound
+	}
+	return tc, nil
+}
+
+// CentContext updates a context with tenant info using the configstore, must only be used for api handlers
+func CentContext(ctx context.Context, config configstore.Service) (context.Context, error) {
```
mikiquantum: I believe that the contextutil/context.go should only be a helper wrapper for our logic. On those lines, I'd like more to extract the `config.GetTenant(id)` to the invoker (handlers) and just use the `NewCentrifugeContext(ctx, tc)` 
It is a bit more overhead to every handler call but keeps context simpler, What do u think?

documents/purchaseorder/handler.go
vimukthi-git: oh I realise this is not a p2p handler so I believe it still makes sense to keep this until we remove grpc from API.

mikiquantum: yep, not asking to do it now.

PR #610, opened 2019-01-07
genericDocumentService: added interface and bootstrappers

manuelpolzhofer:


documents/genericdoc/service.go
```diff
@@ -22,6 +22,31 @@ import (
 	logging "github.com/ipfs/go-log"
 )
 
+// Service provides an interface for generic document methods
+type Service interface {
```
manuelpolzhofer: We can't use the 'documents.Service' interface.

Speaking in set-theory the need common methods (document.Service) are not identical with the method which can be implemented in a generic way.

manuelpolzhofer:
@vedhavyas Yes. 
here is the one for invoice: https://github.com/centrifuge/go-centrifuge/pull/620

PR #611, opened 2019-01-07
WIP: define read ACL protos

### LINKED ISSUE #587: Update core document  Protobuf with Role, Action, ReadRule
Follow the spec to implement the protobufs

#### Testing

#### Acceptance Criteria

vedhavyas:
fixes #587

### DIFF
--- a/coredocument/coredocument.go
+++ b/coredocument/coredocument.go
@@ -100,7 +100,7 @@ func GetDocumentRootTree(document *coredocumentpb.CoreDocument) (tree *proofs.Do
 		leaf := proofs.LeafNode{
 			Hash:     payload[:],
 			Hashed:   true,
-			Property: sigProperty.ElemProp(proofs.FieldNum(i)),
+			Property: sigProperty.SliceElemProp(proofs.FieldNum(i)),
 		}
 		leaf.HashNode(h, false)
 		sigLeafList[i+1] = leaf
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -86,14 +86,14 @@
   revision = "80f35a18b9a0c632d7843e7e536873b0536f0040"
 
 [[projects]]
-  digest = "1:f84321c62605dab82f35a0795dfe30981199e28ea8a0652e5e7ba9af66575fef"
+  digest = "1:7564ab9af5eead6e9a7f4935ada1b44336b6e0b3645ed0b9e258660d04d81a73"
   name = "github.com/centrifuge/precise-proofs"
   packages = [
     "proofs",
     "proofs/proto",
   ]
   pruneopts = "UT"
-  revision = "d0df6e2ed059de47375ef8281529a500cd920261"
+  revision = "e6376eab28a9a7da6c94f8d23099932a65b4e2a2"
 
 [[projects]]
   digest = "1:0ef770954bca104ee99b3b6b7f9b240605ac03517d9f98cbc1893daa03f3c038"
@@ -1519,7 +1519,6 @@
     "github.com/gavv/httpexpect",
     "github.com/go-errors/errors",
     "github.com/gogo/protobuf/io",
-    "github.com/golang/protobuf/jsonpb",
     "github.com/golang/protobuf/proto",
     "github.com/golang/protobuf/protoc-gen-go",
     "github.com/golang/protobuf/ptypes",
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -40,7 +40,7 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 
 [[override]]
   name = "github.com/centrifuge/precise-proofs"
-  revision = "d0df6e2ed059de47375ef8281529a500cd920261"
+  revision = "e6376eab28a9a7da6c94f8d23099932a65b4e2a2"
 
 [[constraint]]
   name = "github.com/centrifuge/gocelery"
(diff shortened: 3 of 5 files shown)


coredocument/coredocument.go
```diff
@@ -150,33 +151,52 @@ func GetDocumentSigningTree(document *coredocumentpb.CoreDocument) (tree *proofs
...
-	newCD, err := NewWithCollaborators(collaborators)
+	ncd := New()
+	ucs, err := fetchUniqueCollaborators(oldCD.Collaborators, collaborators)
+	if err != nil {
+		return nil, errors.New("failed to decode collaborator: %v", err)
+	}
+
+	cs := oldCD.Collaborators
+	for _, c := range ucs {
+		c := c
```
mikiquantum: beautiful syntax 😉

coredocument/coredocument.go
```diff
@@ -295,3 +320,30 @@ func CreateProofs(dataTree *proofs.DocumentTree, coreDoc *coredocumentpb.CoreDoc
 
 	return proofs, nil
 }
+
+func fetchUniqueCollaborators(oldCollabs [][]byte, newCollabs []string) (ids []identity.CentID, err error) {
+	ocsm := make(map[string]struct{})
```
mikiquantum: Maybe we can move this logic as a util to create "sets" of strings. I think it will be useful in other places.

coredocument/coredocument_test.go
```diff
@@ -101,7 +100,7 @@ func TestGetDataProofHashes(t *testing.T) {
 
 	hashes, err := getDataProofHashes(cd)
 	assert.Nil(t, err)
-	assert.Equal(t, 4, len(hashes))
+	assert.Equal(t, 5, len(hashes))
```
mikiquantum: out of curiosity, which change triggered this extra hash?

coredocument/read_acls.go
```diff
@@ -0,0 +1,56 @@
+package coredocument
+
+import (
+	"github.com/centrifuge/centrifuge-protobufs/gen/go/coredocument"
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/identity"
+)
+
+// ErrZeroCollaborators error when no collaborators are passed
+const ErrZeroCollaborators = errors.Error("require atleast one collaborator.")
```
mikiquantum: `at least` typo. Do we need the `.` at the end?

coredocument/read_acls.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if len(collabs) < 1 {
+		return ErrZeroCollaborators
+	}
+
+	addCollaboratorsToReadRules(cd, collabs)
+	return nil
+}
+
+func addCollaboratorsToReadRules(cd *coredocumentpb.CoreDocument, collabs []identity.CentID) {
```
mikiquantum: Shouldn't be `addCollaboratorsToReadSignRules` instead?

coredocument/read_acls.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// create a role for given collaborators
+	role := new(coredocumentpb.Role)
+	for _, c := range collabs {
+		c := c
+		role.Collaborators = append(role.Collaborators, c[:])
+	}
+
+	// add the role to Roles
+	cd.Roles = appendRole(cd.Roles, role)
```
mikiquantum: If this is an update or new version of the document wouldn't this appendRole append duplicates to the existing Roles of the previous version? I guess that depends on how we "prepare" the new version.
I can see that the initReadRules returns if there are rules/roles already. Will the `addCollaboratorsToReadRules` be only used from the `initReadRules` call? Otherwise I think it will add duplicates.

coredocument/coredocument.go
```diff
@@ -150,33 +151,52 @@ func GetDocumentSigningTree(document *coredocumentpb.CoreDocument) (tree *proofs
 
 // PrepareNewVersion creates a copy of the passed coreDocument with the version fields updated
 // Adds collaborators and fills salts
-// Note: ignores any collaborators in the oldCD
+// Note: new collaborators are added to the list with old collaborators.
 func PrepareNewVersion(oldCD coredocumentpb.CoreDocument, collaborators []string) (*coredocumentpb.CoreDocument, error) {
-	newCD, err := NewWithCollaborators(collaborators)
+	ncd := New()
```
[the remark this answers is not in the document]
vedhavyas: Those should come as a separate field in the APIs IMO

coredocument/coredocument.go
vedhavyas: This is to avoid the slice reference of the previous element

coredocument/coredocument.go
vedhavyas: We can move this when we need it maybe ?

coredocument/coredocument.go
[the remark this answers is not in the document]
mikiquantum: Meaning an "override" flag in the API?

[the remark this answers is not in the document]
vedhavyas: something like that, yes :)

coredocument/coredocument.go
lucasvo: I wouldn't touch the collaborators stuff too much. We will get rid of that part of the document when implementing write ACL. Then it will be completely replaced by ACLs.

PR #620, opened 2019-01-08
using generic document service inside invoice service

manuelpolzhofer:
Please review #610 first.


manuelpolzhofer:
I removed all tests if the method only calls the generic document service. 
The removed tests are anyway more or less identical to the tests in the generic document service

Issue #616, opened 2019-01-07
P2P - add CID to protocol ID and load tenant from stream


vimukthi-git:
<This is intended as a guide when thinking about or creating issues for go-centrifuge>

#### Testing
- Unit tests

#### Acceptance Criteria
- Tenant should load correctly into context for p2p requests

PR #626, opened 2019-01-09
P2P - tenant <--> protocol mapping

vimukthi-git:
Issue Link: #616

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/centrifuge/go-centrifuge/p2p/receiver"
+
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/protocol"
 	"github.com/golang/protobuf/proto"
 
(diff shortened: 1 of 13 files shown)


p2p/client.go
```diff
@@ -102,18 +106,18 @@ func (s *peer) getPeerID(id identity.Identity) (libp2pPeer.ID, error) {
 
 // getSignatureForDocument requests the target node to sign the document
 func (s *peer) getSignatureForDocument(ctx context.Context, identityService identity.Service, doc coredocumentpb.CoreDocument, receiverPeer libp2pPeer.ID, receiverCentID identity.CentID) (*p2ppb.SignatureResponse, error) {
-	senderID, err := s.config.GetIdentityID()
+	self, err := contextutil.Self(ctx)
```
mikiquantum: I was making these changes as well, thanks.

p2p/receiver/protocol.go
```diff
@@ -0,0 +1,24 @@
+package receiver
```
mikiquantum: I created a different package in my pr for some common sender/receiver utils, I will move that there as part of my work.

p2p/server.go
```diff
@@ -71,39 +60,60 @@ func (*peer) Name() string {
 func (s *peer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr chan<- error) {
 	defer wg.Done()
 
-	if s.config.GetP2PPort() == 0 {
+	nc, err := s.config.GetConfig()
```
mikiquantum: Cool

p2p/server.go
```diff
@@ -71,39 +60,60 @@ func (*peer) Name() string {
...
+	var protocols []protocol.ID
+	for _, t := range tcs {
+		CID, err := identity.ToCentID(t.IdentityID)
+		if err != nil {
+			startupErr <- err
+			return
+		}
+		protocols = append(protocols, receiver.ProtocolForCID(CID))
+	}
+	s.mes.init(protocols...)
```
mikiquantum: We briefly talked about adding the protocol list to the messenger struct and let the init iterate over them? I am ok leaving it as is. The only question for me is if we see a use case on having them as part of the struct.

vimukthi-git: Thats part of the messenger isn't it? so there won't be a change here even if we implement your suggested change.

vimukthi-git:
@mikiquantum I don't know. It seems there were some changes.

p2p/server.go
mikiquantum: If we have the protocols within the messenger then we do not need to pass protocol list to the `init` method. Just `s.mes.init()`.

Issue #550, opened 2018-12-10
Support localhost loops when opening connection between tenants in same host


mikiquantum:
Can we leverage current p2p dialing to call to itself? 
Does p2p support receiving connections from self as well?

 Otherwise implement an internal  call model.

vimukthi-git:
Tried locally and with a test project and encountered error 
`dial to self attempted` 

so its not allowed right now.

https://github.com/libp2p/go-libp2p/issues/328
This is an open issue, I'm attempting to circumvent this problem by other means.

vimukthi-git:
One very non invasive solution I found for the problem is to construct two libp2p hosts, one for external peer communication and one for local. They would run on different ports. The p2p/client/connect method would check the tenant is registered in the registry and decide to use the libp2p "local host" that we created for a local tenant or else the standard one. 

This approach I tested and it works.

vimukthi-git:
@mikiquantum I think it doesn't make sense to have two libp2p hosts as suggested above all though it would have been a much more non invasive approach given if we still had p2p GRPC. But because I'm removing P2P GRPC and also using the context based approach, I will just make this implementation so that we call the tenant p2p event handlers(which replace grpc calls) with the tenantID in the context.
cc @lucasvo

PR #627, opened 2019-01-09
 Support local tenant loops

vimukthi-git:
Issue Link: #550

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -36,17 +36,23 @@ type Client interface {
 }
 
 func (s *peer) SendAnchoredDocument(ctx context.Context, id identity.Identity, in *p2ppb.AnchorDocumentRequest) (*p2ppb.AnchorDocumentResponse, error) {
+	nc, err := s.config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	peerCtx, _ := context.WithTimeout(ctx, nc.GetP2PConnectionTimeout())
 	cid := id.CentID()
 	tc, err := s.config.GetTenant(cid[:])
 	if err == nil {
 		// this is a local tenant
 		h := s.handlerCreator()
 		// the following context has to be different from the parent context since its initiating a local peer call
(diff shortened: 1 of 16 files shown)


p2p/client.go
```diff
@@ -194,22 +209,46 @@ func (s *peer) GetSignaturesForDocument(ctx context.Context, identityService ide
...
-		if err != nil {
-			return centerrors.Wrap(err, "error fetching collaborator identity")
-		}
-
-		receiverPeer, err := s.getPeerID(id)
-		if err != nil {
-			log.Error(centerrors.Wrap(err, "failed to connect to target"))
-			continue
+		tc, err := s.config.GetTenant(collaboratorID[:])
+		if err == nil {
```
mikiquantum: I think the concept is good, although I would rather have this switch done way down the pipeline, so we do not have to copy this code elsewhere. Meaning when invoking `s.mes.sendMessage()` chose there if is local or remote, should be transparent to the business logic (as much as possible). Let's sit down tomorrow and draft out a solution.

p2p/client.go
```diff
@@ -189,27 +238,14 @@ func (s *peer) GetSignaturesForDocument(ctx context.Context, identityService ide
 	}
 
 	var count int
+	peerCtx, _ := context.WithTimeout(ctx, nc.GetP2PConnectionTimeout())
 	for _, collaborator := range extCollaborators {
 		collaboratorID, err := identity.ToCentID(collaborator)
 		if err != nil {
 			return centerrors.Wrap(err, "failed to convert to CentID")
 		}
-		id, err := identityService.LookupIdentityForID(collaboratorID)
```
vimukthi-git: These two expensive calls happen inside collaborator specific go routines now because of your comment @mikiquantum :)

vimukthi-git:
@mikiquantum created https://app.zenhub.com/workspaces/centrifuge-5ba350114b5806bc2be90978/issues/centrifuge/go-centrifuge/630 to track our discussion

PR #632, opened 2019-01-11
identity: gocelery task for successful transaction

manuelpolzhofer:
- added a gocelery task
- nft minting using new task instead of confirm event task


manuelpolzhofer:
test and old code remove will be in a separate PR

testworld/start_test.go
```diff
@@ -15,13 +15,13 @@ var isRunningOnCI = len(os.Getenv("TRAVIS")) != 0
 
 // Adjust these based on local testing requirments, please revert for CI server
 var configFile = "configs/local.json"
-var runPOAGeth = !isRunningOnCI
+var runPOAGeth = false
```
vimukthi-git: revert ;)

testworld/httputils.go
```diff
@@ -130,6 +131,11 @@ func waitTillSuccess(t *testing.T, e *httpexpect.Expect, txID string) {
 	for {
 		resp := e.GET("/transactions/" + txID).Expect().Status(200).JSON().Object()
 		status := resp.Path("$.status").String().Raw()
+
+		fmt.Println("Centrifuge Transaction Status")
```
vimukthi-git: revert ;)

ethereum/transaction_status_task.go
```diff
@@ -0,0 +1,156 @@
...
+
+type transactionStatusTask struct {
+	transactions.BaseTask
+	timeout time.Duration
+	//state
+	ethContextInitializer func(d time.Duration) (ctx context.Context, cancelFunc context.CancelFunc)
+
+	//task parameter
+	txHash   string
+	tenantID identity.CentID
```
vedhavyas: `accountID`

ethereum/transaction_status_task.go
```diff
@@ -0,0 +1,156 @@
...
+		} else {
+			log.Infof("Transaction successful:%v\n", nftc.txHash)
+		}
+
+		err = nftc.UpdateTransaction(nftc.tenantID, nftc.TaskTypeName(), err)
+	}()
+
+	isPending := true
+	client := GetClient()
+	for isPending {
```
vimukthi-git: please use `retry` feature of gocelery instead of for loop

ethereum/transaction_status_task.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return errors.New("malformed kwarg [%s] because [%s]", queue.TimeoutParam, err.Error())
+		}
+		nftc.timeout = td
+	}
+
+	return nil
+}
+
+func getTransactionStatus(ctx context.Context, txHash string) (bool, error) {
```
vedhavyas: I think bool/err is redundant here. If we can change this to `func isTransactionSuccessful(...) bool {}`, if ok {}, then it is successful else failed
What do you think ?

ethereum/transaction_status_task.go
```diff
@@ -0,0 +1,156 @@
...
+	client := GetClient()
```
vedhavyas: lets take the client as function parameter. easier to test code

ethereum/transaction_status_task.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if err != nil {
+		return false, err
+	}
+
+	if receipt.Status == TransactionStatusSuccess {
+		return true, nil
+	}
+
+	return false, errors.New("Transaction failed")
```
vedhavyas: Once we change the definition, we can make return much easier.

testworld/start_test.go
vedhavyas: We need better solution than this soon :)

vedhavyas:
@manuelpolzhofer I'll integrate the retry logic along with other breaking changes from gocelery in the next PR. Lets push this once the tests are successful

cc: @vimukthi-git

vimukthi-git:
@vedhavyas why do we need to change this?

testworld/start_test.go
vimukthi-git: we will, its on the way

ethereum/transaction_status_task.go
vimukthi-git: we need to always return the error, otherwise it would be a swallowed error ;)

ethereum/transaction_status_task.go
vimukthi-git: As commented

ethereum/transaction_status_task.go
vimukthi-git: please dont do these changes now, it would be done in a single pr by me..

vimukthi-git: https://app.zenhub.com/workspaces/centrifuge-5ba350114b5806bc2be90978/issues/centrifuge/go-centrifuge/617

vedhavyas:
This will help us support infura fully especially kovan issues. @vimukthi-git

ethereum/transaction_status_task.go
vedhavyas: in that return an error instead of returning both is what I'm hinting at :)

vimukthi-git:
@manuelpolzhofer I just realise there are no tests at all for the `transaction_status_task.go`,
please add.

PR #633, opened 2019-01-11
Fix storage <--> config cycles

vimukthi-git:
This is required because for my next prs where we create tenant keys through API.


vedhavyas:
@vimukthi-git I do not have any context on this PR. What does this PR do?

vimukthi-git:
@vedhavyas the change here is to extract interfaces into base packages in both `configstore` and `storage` so that we follow a standard go package structure of parent containing all interface types and implementations in the child. If you look at `identity` package you can already see this implemented.

config/configuration.go
```diff
@@ -100,6 +107,38 @@ type Configuration interface {
...
+	GetSigningKeyPair() (pub, priv string)
+	GetEthAuthKeyPair() (pub, priv string)
+	GetEthereumContextWaitTimeout() time.Duration
+
+	// CreateProtobuf creates protobuf
+	CreateProtobuf() *configpb.TenantData
+}
+
+// Service exposes functions over the config objects
+type Service interface {
```
vimukthi-git: @vedhavyas moving these interfaces is the main change 1

storage/leveldb/leveldb.go
```diff
@@ -1,10 +1,12 @@
-package storage
+package leveldb
```
vimukthi-git: @vedhavyas moving these interfaces is the main change 2

vimukthi-git:
@vedhavyas @mikiquantum I'm merging this as this is just small refactoring(but causes many changes) and I have to resolve conflicts in two other branches which are based on this when someone uses config services. Please feel free to comment if there are any concerns, I will make any change required ASAP. Thanks!

Issue #613, opened 2019-01-07
Generate Keys as part of CreateTenant API Endpoint


mikiquantum:
For ease of use, if Signing and/or ethauth key not provided, generate a random key for the new tenant

#### Testing
- Create testworld or functional tests to cover new tenants with non-provided keys

#### Acceptance Criteria
- Tests pass

PR #637, opened 2019-01-13
API /config/tenants/generate endpoint

vimukthi-git:
Issue Link: #613


vimukthi-git:
@mikiquantum this would create a new endpoint to generate a new tenant with keys instead of changing existing create endpoint. I think create endpoint can be removed soon as it doesn't make sense to take in parameters from the API caller, it can be kind of a security problem as well. 

This pr also creates a new config parameter `tenants.keystore` to store the generated keys for tenants. They would be store in a location like `tenants.keystore`/<centID>/

testworld/start_test.go
```diff
@@ -18,10 +18,10 @@ var configFile = "configs/local.json"
 var runPOAGeth = !isRunningOnCI
 
 // make this true this when running for the first time in local env
-var createHostConfigs = isRunningOnCI
+var createHostConfigs = true
```
mikiquantum: revert

Issue #612, opened 2019-01-07
Add Auth/Signature Header to P2P Envelope


mikiquantum:
Add signature to validate payload was not tampered during the transmission.
Use P2P Signing key from node.

#### Testing
- Unit/Integration tests for signature creation and validation
- Include bad actors

#### Acceptance Criteria
- Auth Header is available in all P2P Messages
- Auth Header is not empty
- Auth Header is validated by the receiver peer

PR #640, opened 2019-01-14
Added signature to header + handshake validation

mikiquantum:
Issue Link: #612


p2p/receiver/validator.go
```diff
@@ -31,34 +34,63 @@ func (group ValidatorGroup) Validate(header *p2ppb.Header) (errs error) {
...
+		return nil
+	})
+}
+
+func signatureValidator() Validator {
+	return ValidatorFunc(func(envelope *p2ppb.Envelope) error {
+		if envelope == nil || envelope.Header == nil {
+			return errors.New("nil envelope/header")
+		}
+		envData := &p2ppb.Envelope{
```
mikiquantum: Any better ideas to deep copy structs?

p2p/receiver/validator.go
```diff
@@ -31,33 +34,65 @@ func (group ValidatorGroup) Validate(header *p2ppb.Header) (errs error) {
...
+			},
+			Body: envelope.Body,
+		}
+
+		data, err := proto.Marshal(envData)
+		if err != nil {
+			return err
+		}
+
+		if envelope.Header.Signature == nil {
```
vedhavyas: maybe we could move this up so that we can avoid copies and marshalling ?

p2p/receiver/validator.go
vedhavyas: This should help:
https://github.com/golang/protobuf/blob/master/proto/clone.go#L45

vedhavyas:
@mikiquantum can we add a test case for this - https://codecov.io/gh/centrifuge/go-centrifuge/pull/640/diff#D3-72 ?

p2p/receiver/validator.go
mikiquantum: Thanks!

Issue #565, opened 2018-12-12
Add tenantConfig to context in API 


mikiquantum:
- Execute requested service with tenant in the context

PR #642, opened 2019-01-14
Multitenant(Account) api

vimukthi-git:
Issue Link: #565

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -75,7 +75,7 @@ func (c apiServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr cha
 
 	opts := []grpc.ServerOption{
 		grpc.Creds(creds),
-		grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
+		grpcInterceptor(),
 	}
 
 	grpcServer := grpc.NewServer(opts...)
--- a/api/server.go
+++ b/api/server.go
@@ -75,7 +75,7 @@ func (c apiServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr cha
 
 	opts := []grpc.ServerOption{
 		grpc.Creds(creds),
-		// grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
+		grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
 	}
 
 	grpcServer := grpc.NewServer(opts...)
--- a/config/configstore/service.go
+++ b/config/configstore/service.go
@@ -21,6 +21,7 @@ const (
 	ethAuthPrivKeyName = "ethauth.key.pem"
 )
 
+// ProtocolSetter sets the protocol on host for the centID
 type ProtocolSetter interface {
 	InitProtocolForCID(CID identity.CentID)
 }
--- a/config/configstore/service.go
+++ b/config/configstore/service.go
@@ -21,14 +21,14 @@ const (
 	ethAuthPrivKeyName = "ethauth.key.pem"
 )
 
-type protocolSetter interface {
+type ProtocolSetter interface {
 	InitProtocolForCID(CID identity.CentID)
 }
 
 type service struct {
 	repo                 repository
 	idService            identity.Service
-	protocolSetterFinder func() protocolSetter
+	protocolSetterFinder func() ProtocolSetter
 }
 
 // DefaultService returns an implementation of the config.Service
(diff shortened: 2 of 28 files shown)


testworld/start_test.go
```diff
@@ -11,6 +11,14 @@ import (
 	"github.com/centrifuge/go-centrifuge/config"
 )
 
+type testType string
+
+const (
+	withinHost            testType = "withinHost"
+	multiHost             testType = "multiHost"
+	multiHostMultiAccount testType = "multiHostMultiAccount"
```
vimukthi-git: all covered :)

api/server.go
```diff
@@ -75,7 +75,7 @@ func (c apiServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr cha
 
 	opts := []grpc.ServerOption{
 		grpc.Creds(creds),
-		// grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
+		grpcInterceptor(), // enable this once we start requiring the tenant id to passed as header
```
vimukthi-git: main change

testworld/document_consensus_test.go
```diff
@@ -12,37 +12,153 @@ import (
...
+		},
+		{
+			"Invoice_withinhost_AddExternalCollaborator",
+			typeInvoice,
+			withinHost,
+		},
+		{
+			"Invoice_multiHostMultiAccount_AddExternalCollaborator",
+			typeInvoice,
+			multiHostMultiAccount,
```
vimukthi-git: testworld covered

api/server.go
mikiquantum: Can remove the comment now

config/configstore/service.go
```diff
@@ -89,6 +95,14 @@ func (s service) GenerateTenant() (config.TenantConfiguration, error) {
 		return nil, err
 	}
 
+	// minor hack to set same p2p keys as node to tenant: Set the new tenant ID to copy of main tenant and create p2p keys
```
mikiquantum: I do not understand this, why not using the same `tc` ?

contextutil/context.go
```diff
@@ -47,13 +49,19 @@ func Tenant(ctx context.Context) (config.TenantConfiguration, error) {
 }
 
 // CentContext updates a context with tenant info using the configstore, must only be used for api handlers
-func CentContext(ctx context.Context, config config.Service) (context.Context, error) {
-	// TODO [multi-tenancy] remove following and read the tenantID from the context
-	tc, err := config.GetAllTenants()
+func CentContext(ctx context.Context, cs config.Service) (context.Context, error) {
+	tcIDHex := ctx.Value(config.TenantKey).(string)
```
mikiquantum: Can we build a wrapper function (same as above) for loading the TenantKey?

config/configstore/service.go
vimukthi-git: because I'm using main tenant p2p keys with new tenants ID, so its not a consistent config. Should not be saved to db.

contextutil/context.go
vimukthi-git: Sure we can but I'm curious what the use case would be? since this function is the only one needed by handlers to extract auth header.

mikiquantum: Just want to be consistent with every value in the context has a helper function to be retrieved abstracting the type check to the function itself and not the consumer.

vimukthi-git: Just don't want that to be used by anything other than the handlers, having that kind of function gives the meaning that it's a common value that should be used, which is not the case as it comes from outside of the node, should be used with care, and any validation has to go here in this function in the future.

contextutil/context.go
```diff
@@ -47,13 +49,19 @@ func Tenant(ctx context.Context) (config.TenantConfiguration, error) {
 }
 
 // CentContext updates a context with tenant info using the configstore, must only be used for api handlers
-func CentContext(ctx context.Context, config config.Service) (context.Context, error) {
-	// TODO [multi-tenancy] remove following and read the tenantID from the context
-	tc, err := config.GetAllTenants()
+func CentContext(ctx context.Context, cs config.Service) (context.Context, error) {
```
lucasvo: NodeContext instead of CentContext?

contextutil/context.go
mikiquantum: I do not see a problem on that (for that particular case), if they have access to the TenantConfig they have access to the TenantID and the context is immutable.

contextutil/context.go
vimukthi-git: Sure, will rename soon in the other pr I have for config stuff. I want to have this in dev and use it in my other prs.

PR #644, opened 2019-01-15
Update version.go

Remi-Tuy:
Maybe this should be made automatic somehow rather then a string, the current version is: v.0.0.2-alpha3.
/ping also returns the wrong version based on this I assume.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -10,7 +10,7 @@ import (
 var gitCommit = "master"
 
 // CentrifugeNodeVersion is the current version of the app
-const CentrifugeNodeVersion = "0.0.1-alpha"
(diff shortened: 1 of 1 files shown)


vedhavyas:
@Remi-Tuy agree `Ping` is returning the wrong version. We will add the automation to our pipeline.

mikiquantum:
@Remi-Tuy I believe we are overriding the `gitCommit` when compiling the binary using LD_Flags when we release the official binary already.
This is either something we need to have in mind when we release always update here or (better in my opinion) use same approach as the gitCommit and override the value when compiling.

vedhavyas:
Closing this as the version is updated through a release
cc: @Remi-Tuy

Issue #646, opened 2019-01-15
Stop storing node config in the db


vimukthi-git:
<This is intended as a guide when thinking about or creating issues for go-centrifuge>

#### Testing
- existing Tests should pass

#### Acceptance Criteria
- Node should not depend on the db for node config

PR #649, opened 2019-01-15
Import main node config file everytime node starts(simpler than removing config service usage)

vimukthi-git:
Issue Link: #646


config/configstore/bootstrapper.go
```diff
@@ -33,14 +33,12 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
 
 	nc := NewNodeConfig(cfg)
 	configdb.Register(nc)
-	_, err := service.GetConfig()
-	// if node config doesn't exist in the db, add it
+	// install the file based config everytime so that file updates are reflected in the db, direct updates to db are not allowed
+	nc, err := service.CreateConfig(NewNodeConfig(cfg))
```
vedhavyas: So, this basically, overwrites the previous config ?

vimukthi-git: yup

contextutil/context.go
```diff
@@ -48,8 +48,8 @@ func Tenant(ctx context.Context) (config.TenantConfiguration, error) {
 	return tc, nil
 }
 
-// CentContext updates a context with tenant info using the configstore, must only be used for api handlers
-func CentContext(ctx context.Context, cs config.Service) (context.Context, error) {
+// NodeContext updates a context with tenant info using the configstore, must only be used for api handlers
+func NodeContext(ctx context.Context, cs config.Service) (context.Context, error) {
```
mikiquantum: I am not certain about this renaming, as Node sounds like a static context to me, and the purpose of this context is on request basis. I think it makes sense to just use `Context(...)`.

contextutil/context.go
```diff
@@ -48,8 +48,8 @@ func Tenant(ctx context.Context) (config.TenantConfiguration, error) {
...
 	tcIDHex := ctx.Value(config.TenantKey).(string)
```
mikiquantum: Can you create here the wrapping function you said you were going to do earlier?

vimukthi-git: I think I didn't agree with that change since its not something that should be used commonly in the node.

PR #650, opened 2019-01-15
Feat/read acl nfts

### LINKED ISSUE #590: Implement NFT read ACL
#### Testing

#### Acceptance Criteria

vedhavyas:
Fixes #590 

This would add the NFT validator.


coredocument/read_acls.go
```diff
@@ -110,7 +152,101 @@ func isPeerInRole(role *coredocumentpb.Role, peer identity.CentID) bool {
...
+	ch := roleIterator(cd, coredocumentpb.Action_ACTION_READ)
+	found := false
+	for role := range ch {
+		if isNFTInRole(role, registry, tokenID) {
+			found = true
+		}
+	}
+
+	if !found {
+		return errors.New("nft missing")
```
mikiquantum: Is this a suitable place for an error constant?

coredocument/read_acls.go
```diff
@@ -4,15 +4,29 @@ import (
...
 
 const (
 	// ErrZeroCollaborators error when no collaborators are passed
 	ErrZeroCollaborators = errors.Error("require at least one collaborator")
+
+	// ErrPeerNotFound error when peer is not found in the read rules
+	ErrPeerNotFound = errors.Error("peer not found")
+
+	// nftByteCount is the length of combined bytes of registry and tokenID
+	nftByteCount = 52
```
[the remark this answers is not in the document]
vedhavyas: Not planning on exposing this. Moved to constant instead of a magic number

coredocument/read_acls.go
vedhavyas: I feel like we are overusing constant errors. If we were to check the type, I would have but I dont see any special scenario this would be required at the moment

coredocument/read_acls.go
```diff
@@ -41,52 +55,80 @@ func addCollaboratorsToReadSignRules(cd *coredocumentpb.CoreDocument, collabs []
...
+	}
+
+	role := new(coredocumentpb.Role)
+	role.Nfts = append(role.Nfts, nft)
+	addNewRule(cd, role, coredocumentpb.Action_ACTION_READ)
+	return nil
+}
+
+// constructNFT appends registry and tokenID to byte slice
+func constructNFT(registry common.Address, tokenID []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
vedhavyas: That would be redundant since the function returns the bytes already. What do you think?

coredocument/read_acls.go
```diff
@@ -110,7 +144,100 @@ func isPeerInRole(role *coredocumentpb.Role, peer identity.CentID) bool {
...
+	if err != nil {
+		return errors.New("failed to get NFT owner: %v", err)
+	}
+
+	msg, err := constructNFT(registry, tokenID)
+	if err != nil {
+		return err
+	}
+
+	if !secp256k1.VerifySignatureWithAddress(owner.String(), signature, msg) {
```
lucasvo: If I understand this correctly, you check that the signature matches the owner.String(). Owner.String is the Ethereum address of the token owner.

What do you sign? (registry+tokenID) 

In the end, the only thing you check here, is that the actual token owner signed a statement (registry+tokenID). Here's a way to exploit it:

Let's assume Alice and Eve collaborate on a document together. 
1) Alice mint's an NFT for Bob. Bob signs the registry+tokenID creating `sig_bob`. 
2) Bob sends Eve a request to read the document with `sig_bob`
3) Alice removes Eve from the collaborators and updates the document.
    *Eve should not have access to the updated document anymore but Bob still has access*
4) Eve can send an NFT access request to Alice and impersonate Bob just by sending `sig_bob` 👿 

The problem with the signature is that it doesn't scope the permission to one key or identity. If we really want to support this access scheme, the signature needs to be scoped to one identity:

Signature = registry + tokenId + identity

coredocument/read_acls.go
```diff
@@ -110,7 +144,100 @@ func isPeerInRole(role *coredocumentpb.Role, peer identity.CentID) bool {
...
+	if !found {
+		return errors.New("nft missing")
+	}
+
+	// get the owner of the NFT
+	owner, err := r.tokenRegistry.OwnerOf(registry, tokenID)
+	if err != nil {
+		return errors.New("failed to get NFT owner: %v", err)
+	}
+
```
lucasvo: By default we should check if token owner == sender and approve if true. This is the minimum requirement. I would say more important than the signature below.

lucasvo: Thinking more about this actually:

If the NFT owner is not a centrifuge identity, then what is the likelihood that they could easily create the signature. Let's assume you do run a centrifuge node and want to read a document, you would also have to make sure that the ethereum private key that owns the document is also configured on the node as a private key so the node that makes the request can generate the signature. I don't know how likely that is or what the benefit would be in this over just having an identity own the NFT. If the node does not have the private key of the account, then you would have to manually construct the signature and provide the signature as an additional argument to the rest api when making the read request. The usability of this to do this manually is definitely less than ideal and even in a fully integrated situation it's still questionable how often this is practical.

I do think this might be something we want to support in the future, but I really don't see this as a very convenient way for users to read documents in the beginning. Just checking that it is a centrifuge identity should definitely be supported.

Issue #631, opened 2019-01-11
Support getting CoreDocuments


charleenfei:
- [x] New message types on the [Centrifuge-Protobuf Library/related PR](https://github.com/centrifuge/centrifuge-protobufs/pull/47)
- [x] Method on the p2p receiver handler to handle these requests and return the requested CoreDocument

#### Testing
- [x] Testing coverage for these methods

#### Acceptance Criteria
simple implementation for now, will update with ACLs

Issue #651, opened 2019-01-15
Refactor handling of requests in the p2p receiver


charleenfei:
<This is intended as a guide when thinking about or creating issues for go-centrifuge>

Currently, the HandleInterceptor routes incoming requests to the correct handler, depending on the MessageType. However, there is not so much functional differentiation between the handlers, they just unmarshal the message into the correct protobuf format before passing it to the relevant method. This should be refactored so there is not so much code repetition.

This could be done either by moving the unmarshal functionality to the HandleInterceptor method or maybe a separate util function which is called and contains switch cases each MessageType.

A larger refactor could streamline this entire p2p receiver logic to be more efficient.

#### Testing
<Describe required tests: Unit, Integration, Testworld or Functional>

#### Acceptance Criteria
<What is the acceptance criteria of this ticket?>

PR #652, opened 2019-01-16
support getting core documents

charleenfei:
Issue Link: #631


p2p/common/protocol.go
```diff
@@ -61,6 +65,10 @@ func MessageTypeFromString(mt string) MessageType {
 		found = MessageTypeSendAnchoredDoc
 	} else if MessageTypeSendAnchoredDocRep.Equals(mt) {
 		found = MessageTypeSendAnchoredDocRep
+	} else if MessageTypeGetDoc.Equals(mt) {
+		found = MessageTypeGetDoc
+	} else if MessageTypeGetDocRep.Equals(mt) {
+		found = MessageTypeGetDocRep
```
lucasvo: @charleenfei could you change this entire block to use a switch statement instead?

charleenfei: will do!

p2p/receiver/handler.go
```diff
@@ -190,6 +193,47 @@ func (srv *Handler) SendAnchoredDocument(ctx context.Context, docReq *p2ppb.Anch
...
+	p2pEnv, err := p2pcommon.PrepareP2PEnvelope(ctx, nc.GetNetworkID(), p2pcommon.MessageTypeGetDocRep, res)
+	if err != nil {
+		return convertToErrorEnvelop(err)
+	}
+
+	return p2pEnv, nil
+}
+
+// GetDocument receives document identifier and retrieves the corresponding CoreDocument from the repository
+func (srv *Handler) GetDocument(ctx context.Context, docReq *p2ppb.GetDocumentRequest) (*p2ppb.GetDocumentResponse, error) {
```
vedhavyas: Let's move this to the above function itself and rename `HandleGetDocument` -> `GetDocuemnt` ?

charleenfei: That logic would differ from how the other methods on the handlers work (SendAnchoredDoc, ReqDocumentSignature). I agree there is redundancy here, I have noted this in a separate issue #651. Do you think it makes sense to address that in a separate PR?

Issue #489, opened 2018-11-28
[Webhook] Add from/to Peer to the Protobuf message and adapt code


mikiquantum:
Protobuf Message:
```
message NotificationMessage {
  uint32 event_type = 1;
  string from = 6;
  string to = 8;
  google.protobuf.Timestamp recorded = 3;
  string document_type = 4;
  string document_id = 7;
}
```
Fill new fields accordingly in the code.

lucasvo:
I don't think the 'from/to' is the right naming. It's not entirely clear which field defines the tenant. I would just add one field called `identity`. Given not all notifications will have a destination identity I would remove that from the standard notifications.

vedhavyas:
Why not we rename then as sender(who triggered the notification) and receiver instead ?

mikiquantum:
Final agreement:
```
{
  account_id string,
  from_id string,
  to_id string
}
```

PR #653, opened 2019-01-16
add account fields to notification

mikiquantum:
Issue Link: #489


notification/notification_test.go
```diff
@@ -50,7 +50,8 @@ func (m mockConfig) GetReceiveEventNotificationEndpoint() string {
 
 func TestWebhookSender_Send(t *testing.T) {
 	docID := utils.RandomSlice(32)
-	cid := utils.RandomSlice(32)
+	accountID := utils.RandomSlice(6)
```
lucasvo: Why is the accountid so short?

vimukthi-git: its a centID

mikiquantum: the account_id is the centID and as of today it is 6 bytes long.

vimukthi-git: btw @mikiquantum could you use the `identity.CentIDLength` constant for this or use `identity.RandomCentID` function? better for readability and understanding.

Issue #617, opened 2019-01-07
Rename `tenant` to `account`


vimukthi-git:
With this change we introduce the concept of a Centrifuge Account to the system.

#### Testing
No new tests needed

#### Acceptance Criteria
All references to Tenant or multi tenancy should be removed and replaced with Account.

PR #654, opened 2019-01-16
Account rename 1: move API /config/tenants => /accounts

vimukthi-git:
Issue Link: #617


protobufs/accounts/service.proto
```diff
@@ -0,0 +1,91 @@
...
+option go_package = "accountpb";
+option java_multiple_files = true;
+option java_outer_classname = "ServiceProto";
+option java_package = "com.account";
+
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+import "protoc-gen-swagger/options/annotations.proto";
+
+// AccountService allows to manage the node and tenant configuration of the node
```
vimukthi-git: shit, this needs to change :)

config/configstore/handler.go
```diff
@@ -23,8 +22,13 @@ func GRPCHandler(svc config.Service) configpb.ConfigServiceServer {
 	return &grpcHandler{service: svc}
 }
 
-func (h grpcHandler) deriveAllTenantResponse(cfgs []config.TenantConfiguration) (*configpb.GetAllTenantResponse, error) {
-	response := new(configpb.GetAllTenantResponse)
+// GRPCAccountHandler returns an implementation of configpb.ConfigServiceServer
```
vimukthi-git: change

api/service.go
```diff
@@ -99,6 +100,13 @@ func registerServices(ctx context.Context, cfg Config, grpcServer *grpc.Server,
 		return err
 	}
 
+	// account api
```
mikiquantum: Does it mean we do not need the above `config api` registration right?

protobufs/accounts/service.proto
```diff
@@ -0,0 +1,91 @@
+syntax = "proto3";
```
mikiquantum: folder name is more standard in singular `protobufs/account/service.proto`

api/service.go
vimukthi-git: we have two apis since `GET /config` endpoint is not removed and its a separate API from now on.

mikiquantum: why do we need the `GET` endpoint exposed if we read from the file always?

vimukthi-git: I felt its nicer to be able to see the config using an API, since we had the API thought of not leaving it as it is.

PR #655, opened 2019-01-16
Add Auth Header to Swagger UI

mikiquantum:
Issue Link: <Github link>


build/scripts/build_swagger.js
```diff
@@ -6,6 +6,14 @@ const swaggerFilesPath = path.resolve(__dirname, '../../protobufs/gen/swagger');
 const swaggerJsonPath = path.resolve(__dirname, '../../protobufs/gen/swagger.json');
 const swaggerConfig = require(path.resolve(__dirname, '../swagger_config'));
 
+let authHeader = {
+  "name": "authorization",
+  "in": "header",
+  "description": "Authorization Header",
```
vimukthi-git: may be add more info like whats intended to be sent
eg: `Authorization Header is the centrifuge ID of the account for the intended API action`

vimukthi-git: ..`hex encoded centrifuge ID`.. might be better above

Issue #656, opened 2019-01-16
Refactor handling of MessageTypes in p2pcommon


charleenfei:
The if-else statements block could be refactored to a map format, something like this:

```go

func MessageTypeFromString(ht string) MessageType {

	var messageType MessageType

	messageTypes := map[string]MessageType{
		// MessageTypeError defines any protocol error
		"MessageTypeError": "MessageTypeErrorSTRING",
		// MessageTypeInvalid defines invalid protocol type
		"MessageTypeInvalid": "MessageTypeInvalid",
		// MessageTypeRequestSignature defines RequestSignature type
		"MessageTypeRequestSignature": "MessageTypeRequestSignature",
		// MessageTypeRequestSignatureRep defines RequestSignature response type
		"MessageTypeRequestSignatureRep": "MessageTypeRequestSignatureRep",
		// MessageTypeSendAnchoredDoc defines SendAnchored type
		"MessageTypeSendAnchoredDoc": "MessageTypeSendAnchoredDoc",
		// MessageTypeSendAnchoredDocRep defines SendAnchored response type
		"MessageTypeSendAnchoredDocRep": "MessageTypeSendAnchoredDocRep",
		//MessageTypeGetAnchoredDoc defines GetAnchoredDoc type
		"MessageTypeGetDoc": "MessageTypeGetDoc",
		//MessageTypeGetAnchoredDocRep defines GetAnchoredDoc response type
		"MessageTypeGetDocRep": "MessageTypeGetDocRep",
	}

	if mt, exists := messageTypes[ht]; exists {
		messageType = mt
	}
	return messageType
}

```

PR #662, opened 2019-01-17
Refactor/message types to map

charleenfei:
A small refactor addressing #656, larger refactor of p2p handler/messagetypes to follow per #651 .

### DIFF
--- a/p2p/common/protocol.go
+++ b/p2p/common/protocol.go
@@ -56,14 +56,14 @@ func MessageTypeFromString(ht string) MessageType {
 	var messageType MessageType
 
 	messageTypes := map[string]MessageType{
-		"MessageTypeError": "MessageTypeError",
-		"MessageTypeInvalid": "MessageTypeInvalid",
-		"MessageTypeRequestSignature": "MessageTypeRequestSignature",
+		"MessageTypeError":               "MessageTypeError",
+		"MessageTypeInvalid":             "MessageTypeInvalid",
+		"MessageTypeRequestSignature":    "MessageTypeRequestSignature",
 		"MessageTypeRequestSignatureRep": "MessageTypeRequestSignatureRep",
-		"MessageTypeSendAnchoredDoc": "MessageTypeSendAnchoredDoc",
-		"MessageTypeSendAnchoredDocRep": "MessageTypeSendAnchoredDocRep",
-		"MessageTypeGetDoc": "MessageTypeGetDoc",
-		"MessageTypeGetDocRep": "MessageTypeGetDocRep",
+		"MessageTypeSendAnchoredDoc":     "MessageTypeSendAnchoredDoc",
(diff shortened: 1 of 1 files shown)


p2p/common/protocol.go
```diff
@@ -51,26 +63,14 @@ func (mt MessageType) String() string {
...
-	} else if MessageTypeSendAnchoredDoc.Equals(mt) {
-		found = MessageTypeSendAnchoredDoc
-	} else if MessageTypeSendAnchoredDocRep.Equals(mt) {
-		found = MessageTypeSendAnchoredDocRep
-	} else if MessageTypeGetDoc.Equals(mt) {
-		found = MessageTypeGetDoc
-	} else if MessageTypeGetDocRep.Equals(mt) {
-		found = MessageTypeGetDocRep
+func MessageTypeFromString(ht string) MessageType {
+
```
vimukthi-git: these extra newlines are unnecessary IMO

p2p/common/protocol.go
```diff
@@ -40,6 +40,18 @@ const (
...
+	"MessageTypeRequestSignature":    "MessageTypeRequestSignature",
+	"MessageTypeRequestSignatureRep": "MessageTypeRequestSignatureRep",
+	"MessageTypeSendAnchoredDoc":     "MessageTypeSendAnchoredDoc",
+	"MessageTypeSendAnchoredDocRep":  "MessageTypeSendAnchoredDocRep",
+	"MessageTypeGetDoc":              "MessageTypeGetDoc",
+	"MessageTypeGetDocRep":           "MessageTypeGetDocRep",
+}
+
 // Equals compares if string is of a particular MessageType
 func (mt MessageType) Equals(mt2 string) bool {
```
vimukthi-git: please remove this if not used

vimukthi-git:
@mikiquantum I agree this is not much of an improvement in terms of adding messages, the great thing here is that adding message types is now a more declarative exercise than a logic change, I feel that to be a good thing for protocol extensibility and readability.

mikiquantum:
@vimukthi-git couldn't we achieve both by transforming the type to int and use iota enum?

vimukthi-git:
Sure, I think that would work as well, 
this whole discussion started because @lucasvo and I thought having `switch` case here was more idiomatic than `if-else`. But map seemed to me was even better as we can remove logic. If you have a better way to map these I'm sure I would agree as well.
 This is just the minimum change required to remove the `if-else`. IMO lets just do this now and make those changes you suggest in the future.

charleenfei:
agree with @vimukthi-git about the temp/quick and dirty nature of this refactor. I think that it would make sense to have a larger refactor of how MessageTypes are used, including the way that HandleInterceptor routes requests to the relevant handlers, there is a lot of code redundancy there (see #651 housekeeping).

PR #663, opened 2019-01-17
Feat/nft document update

vedhavyas:
Once the NFT is created, we need to update the document with nft data added to read rules.


nft/ethereum_payment_obligation.go
```diff
@@ -160,18 +161,41 @@ func (s *ethereumPaymentObligation) OwnerOf(registry common.Address, tokenID []b
...
+		return uuid.Nil, nil, err
+	}
+
+	var txID uuid.UUID
+	txID, err = ethereum.QueueTransactionStatusTask(cid, tx.Hash(), s.txService, s.queue, func(status transactions.Status) error {
+		if status == transactions.Failed {
+			return nil
+		}
+
+		_, err := documents.InitNFTCreatedTask(
```
vimukthi-git: I thought we are not going to do callbacks ;) 
(I totally agree with this change as I suggested :)

transactions/service.go
```diff
@@ -18,37 +19,68 @@ type Service interface {
...
+		return err
+	}
+
+	if len(tx.Logs) < 1 {
+		return nil
+	}
+
+	s.mu.Lock()
+
+	for _, h := range s.handlers[tx.ID] {
```
manuelpolzhofer: Maybe we can move the call of the handlers in an own function together with a comment?

ethereum/geth_client.go
```diff
@@ -206,37 +200,33 @@ func (gc *gethClient) getGethTxOpts(accountName string) (*bind.TransactOpts, err
...
+	queuer queue.TaskQueuer, updateHandler func(status transactions.Status) error) (txID uuid.UUID, err error) {
+	tx, err := txService.CreateTransaction(accountID, "Polling Ethereum transaction status")
 	if err != nil {
 		return txID, err
 	}
-	_, err = gc.queue.EnqueueJob(TransactionStatusTaskName, map[string]interface{}{
+
+	var next bool
+	if updateHandler != nil {
+		next = true
```
manuelpolzhofer: Why is a next flag needed?

I understood it in a way that I can register a handler closure in the transactions and after the 
transaction status is updated the handler func will be called?

The actual task isn't aware that he might have a successor. The update transaction is taking care.
 So why the next flag?

documents/anchor_task.go
```diff
@@ -116,16 +120,23 @@ func (d *documentAnchorTask) RunTask() (res interface{}, err error) {
 }
 
 // InitDocumentAnchorTask enqueues a new document anchor task and returns the txID.
-func InitDocumentAnchorTask(tq queue.TaskQueuer, txService transactions.Service, accountID identity.CentID, modelID []byte) (uuid.UUID, error) {
-	tx, err := txService.CreateTransaction(accountID, documentAnchorTaskName)
+func InitDocumentAnchorTask(tq queue.TaskQueuer, txService transactions.Service, accountID identity.CentID, modelID []byte, txID uuid.UUID) (uuid.UUID, error) {
+	var tx *transactions.Transaction
+	var err error
+	if txID != uuid.Nil {
```
vimukthi-git: we need a utility here `GetOrCreateTx(**)`

transactions/base_task.go
```diff
@@ -46,22 +57,29 @@ func (b *BaseTask) UpdateTransaction(accountID identity.CentID, name string, err
...
-	} else {
-		log.Infof("Transaction successful:%v\n", b.TxID.String())
+		return errors.AppendError(err, b.updateStatus(accountID, Failed, NewLog(name, err.Error())))
 	}
 
-	tx, erri := b.TxService.GetTransaction(accountID, b.TxID)
-	if erri != nil {
-		return errors.AppendError(err, erri)
+	if b.Next {
+		err = b.updateStatus(accountID, Pending, NewLog(name, ""))
```
manuelpolzhofer:
```suggestion
		return b.updateStatus(accountID, Pending, NewLog(name, ""))
```

transactions/base_task.go
```diff
@@ -46,22 +57,29 @@ func (b *BaseTask) UpdateTransaction(accountID identity.CentID, name string, err
...
+		if err != nil {
```
manuelpolzhofer: I think you can remove the IF. instead just return the updateStatus return.

documents/nft_created_task.go
```diff
@@ -0,0 +1,164 @@
...
+	return nil, nil
+}
+
+func (t *nftCreatedTask) TaskTypeName() string {
+	return nftCreatedTaskName
+}
+
+// InitNFTCreatedTask initiates a new nft created task
+func InitNFTCreatedTask(
+	queuer queue.TaskQueuer,
```
manuelpolzhofer: Why are you not creating an own txID for the next task in the queue?

Multiply queuing task are using the same txID?

transactions/service.go
manuelpolzhofer: Is it correct that multiple queuing task are using the same tx.ID?

If that is the case it might cause a problem here:

Let's say task A registers a handler to trigger task B.
B has an own handler as well.

After finishing task A this loop will be executed.
The handler for A would register the handler for task B and queue the task.

Now we have two entries in this list. 
Could it be that the second handler func is now executed before task B is finished?
(Because we are still in the loop)

ethereum/geth_client.go
manuelpolzhofer: Is the reason because we are using the same tx.ID for multiply task to update to transactionStatus to pending instead of success?

vimukthi-git: @manuelpolzhofer you came to the right conclusion, to track the status of a certain network action (transaction) all queued tasks for that action needs to have the same unique txID and also have a status field.

transactions/service.go
vimukthi-git: Also this handler lock is kind of scary,
as this can lead to deadlocks very easily since as someone might create a dependancy between a downstream task and an upstream handler and since handler is arbitrary code it can hold the lock indefinitely I think current solution is very risky in the long term. I have doubts whether we should fix it now when it's easier to fix in this pr.

vimukthi-git:
@manuelpolzhofer @mikiquantum @charleenfei lets have a discussion on this as I have serious doubts about the handler lock, lets decide if we want to merge afterwards.

@manuelpolzhofer thanks for pointing towards the lock situation with your comment.

documents/invoice/service.go
```diff
@@ -122,7 +122,7 @@ func (s service) calculateDataRoot(ctx context.Context, old, new documents.Model
 }
 
 // Create takes and invoice model and does required validation checks, tries to persist to DB
-func (s service) Create(ctx context.Context, inv documents.Model) (documents.Model, uuid.UUID, error) {
+func (s service) Create(ctx context.Context, inv documents.Model, txID uuid.UUID) (documents.Model, uuid.UUID, error) {
```
mikiquantum: This is a great example of a parameter that needs to be available down the stream. I would add it as part of the `ctx`

transactions/service.go
mikiquantum: The way I would approach this is:
- Enable nested transactions out of a parent tx
- Create a tx for each task that is done async.
- Pass through the context the parent tx
- Update the parent tx with every update of the child ones
- The only place where a new child tx is created is by the previous task service (no handlers, or callbacks)
- Only the last task should update the latest status of the parent tx

Let's discuss this tomorrow

transactions/base_task.go
vedhavyas: This was something we have discussed about having same transaction I'd for multiple tasks. Else we will not be able to communicate the next yxid to the client

documents/nft_created_task.go
vedhavyas: This was something we have discussed about having same transaction I'd for multiple tasks. Else we will not be able to communicate the next yxid to the client

documents/invoice/service.go
vedhavyas: I agree!

transactions/service.go
vimukthi-git: @vedhavyas 
A global lock that serialises all transactions not even related to each other is a problem in any system. Since that means a bug or a problems caused by a single handler can cause all other non related transactions to block. You probably have heard about python GIL, its a similar situation here.

I agree that concern that @manuelpolzhofer is could indeed be rare, and thats not the reason I don't agree with this change as you may observe.

vimukthi-git:
@mikiquantum I have take into account your suggestion as well as our discussion and my implementation is going to be through a `TrasactionManager`. Which among other things,
- Would avoid the handlers
- Would relieve individual tasks from the responsibility of updating status
- Would remove global locks
- Would make sure that we have a way to restart TXs after a restart of node
cc @manuelpolzhofer @vedhavyas @lucasvo

nft/ethereum_payment_obligation.go
```diff
@@ -160,18 +168,54 @@ func (s *ethereumPaymentObligation) OwnerOf(registry common.Address, tokenID []b
...
+	opts *bind.TransactOpts,
+	requestData MintRequest,
+	registry common.Address,
+	documentID []byte) (uuid.UUID, error) {
+
+	r := requestData.copy()
+
+	// Run within transaction
+	// We use context.Background() for now so that the transaction is only limited by ethereum timeouts
+	txID, _, err := s.txService.ExecuteWithinTX(context.Background(), cid, uuid.Nil, "Minting NFT", func(accountID identity.CentID, txID uuid.UUID, txMan transactions.Manager, errOut chan<- error) {
```
vimukthi-git: Main update

nft/ethereum_payment_obligation_test.go
```diff
@@ -36,6 +36,35 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+type MockTxManager struct {
+	mock.Mock
+}
+
+func (m MockTxManager) ExecuteWithinTX(ctx context.Context, accountID identity.CentID, existingTxID uuid.UUID, desc string, work func(accountID identity.CentID, txID uuid.UUID, txMan transactions.Manager, err chan<- error)) (txID uuid.UUID, done chan bool, err error) {
```
vimukthi-git: Main update

coredocument/read_acls.go
```diff
@@ -72,21 +72,26 @@ func addNewRule(cd *coredocumentpb.CoreDocument, role *coredocumentpb.Role, acti
...
-	nft, err := constructNFT(registry, tokenID)
+// AddNFTToReadRules adds NFT token to the read rules of core document.
+func AddNFTToReadRules(cd *coredocumentpb.CoreDocument, registry common.Address, tokenID []byte) error {
+	nft, err := ConstructNFT(registry, tokenID)
 	if err != nil {
 		return errors.New("failed to construct NFT: %v", err)
 	}
 
 	role := new(coredocumentpb.Role)
+	rk, err := utils.ConvertIntToByte32(len(cd.Roles))
```
mikiquantum: Thanks!

documents/anchor_task.go
```diff
@@ -116,16 +120,24 @@ func (d *documentAnchorTask) RunTask() (res interface{}, err error) {
 }
 
 // InitDocumentAnchorTask enqueues a new document anchor task and returns the txID.
-func InitDocumentAnchorTask(tq queue.TaskQueuer, txService transactions.Service, accountID identity.CentID, modelID []byte) (uuid.UUID, error) {
-	tx, err := txService.CreateTransaction(accountID, documentAnchorTaskName)
+// TODO [TXManager] migrate this to use TxManager
+func InitDocumentAnchorTask(tq queue.TaskQueuer, txService transactions.Manager, accountID identity.CentID, modelID []byte, txID uuid.UUID) (uuid.UUID, error) {
+	var tx *transactions.Transaction
+	var err error
+	if txID != uuid.Nil {
```
mikiquantum: Will it never be the case that txID here is `nil`? or uuid.Nil is an alias as well of go nil?

documents/anchor_task.go
```diff
@@ -116,16 +120,24 @@ func (d *documentAnchorTask) RunTask() (res interface{}, err error) {
...
+		tx, err = txService.GetTransaction(accountID, txID)
```
mikiquantum: As this is an Init function, in which scenario do we expect a txID to be passed and exists before hand?

documents/bootstrapper.go
```diff
@@ -91,15 +91,25 @@ func (PostBootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return errors.New("p2p client not initialised")
 	}
 
-	task := &documentAnchorTask{
+	txMan := ctx[transactions.BootstrappedService].(transactions.Manager)
+	anchorTask := &documentAnchorTask{
 		BaseTask: transactions.BaseTask{
-			TxService: ctx[transactions.BootstrappedService].(transactions.Service),
+			TxService: txMan,
```
mikiquantum: Why not call the property `TxManager` ?

documents/invoice/model.go
```diff
@@ -319,13 +317,18 @@ func (i *Invoice) UnpackCoreDocument(coreDoc *coredocumentpb.CoreDocument) error
...
 
 	i.loadFromP2PProtobuf(invoiceData)
-	invoiceSalts := &invoicepb.InvoiceDataSalts{}
-	err = proto.Unmarshal(coreDoc.EmbeddedDataSalts.Value, invoiceSalts)
-	if err != nil {
-		return err
-	}
 
-	i.InvoiceSalts = invoiceSalts
+	if coreDoc.EmbeddedDataSalts == nil {
```
mikiquantum: we are changing the policy here to generate salts if empty vs failing. This could cause silent failures when documents get out of sync. Agree that this problem should have been caught way before initially saving it, but I believe this approach will potentially make data inconsistent across the network. What is the main reason for this change?

nft/ethereum_payment_obligation.go
```diff
@@ -65,38 +73,37 @@ func newEthereumPaymentObligation(
...
-		return nil, err
+		return MintRequest{}, err
 	}
 
 	toAddress := common.HexToAddress(depositAddress)
 
 	anchorID, err := anchors.ToAnchorID(corDoc.CurrentVersion)
 	if err != nil {
-		return nil, nil
+		return MintRequest{}, nil
```
mikiquantum: Shouldn't we return here the `err`?

nft/ethereum_payment_obligation.go
```diff
@@ -65,38 +73,37 @@ func newEthereumPaymentObligation(
...
 	}
 
 	rootHash, err := anchors.ToDocumentRoot(corDoc.DocumentRoot)
```
mikiquantum: same here

nft/ethereum_payment_obligation.go
```diff
@@ -65,38 +73,37 @@ func newEthereumPaymentObligation(
...
 	if err != nil {
-		return nil, nil
+		return MintRequest{}, nil
```
mikiquantum: And here

transactions/manager.go
```diff
@@ -0,0 +1,143 @@
...
+		t = newTransaction(accountID, desc)
+		err := s.SaveTransaction(t)
+		if err != nil {
+			return uuid.Nil, nil, err
+		}
+	}
+	done = make(chan bool)
+	go func(ctx context.Context) {
+		err := make(chan error)
+		go work(accountID, t.ID, s, err)
```
mikiquantum: This call will set up the tx.Task["smtg"] status? Or will it be done as part of the below status update?

documents/anchor_task.go
vimukthi-git: As explained offline this can't be nil

documents/anchor_task.go
vimukthi-git: This is a possibility right now because of the some transactions are implemented but would be removed later

documents/invoice/model.go
vimukthi-git: I will check, as agreed

PR #666, opened 2019-01-18
Document service refactor

vedhavyas:
Issue Link: <Github link>


documents/service.go
```diff
@@ -1,10 +1,25 @@
 package documents
```
vimukthi-git: I agree with this change
but I have doubts whether we are making this service interface too specific for our current usecases, anyway we would never know until we have those uses.

PR #668, opened 2019-01-18
Added CMD tests

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/cmd/centrifuge_cmd_test.go
+++ b/cmd/centrifuge_cmd_test.go
@@ -1,0 +1,38 @@
+// +build cmd
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+
+	"github.com/centrifuge/go-centrifuge/testingutils"
+	"github.com/centrifuge/go-centrifuge/version"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


vedhavyas:
@mikiquantum Amazing!!
Can we use go tests instead of shell scripts? Useful for debugging

mikiquantum:
@vedhavyas I want to call the binary explicitly (such as functional test) as a black box test, I can convert the shell script into a go file, but I am still interested on invoking the cmd binary. 
In any case there will not be a debugging option, for that we have (or should have) unit/integration testing.

mikiquantum:
@vedhavyas I actually moved it to a go file, mainly because I noticed I was duplicating the way that we read smart contract addresses.

cmd/centrifuge_cmd_test.go
```diff
@@ -0,0 +1,38 @@
+// +build cmd
```
vedhavyas: Any specific reason why new build tag instead of unit test. I think these belong to unit test IMO

mikiquantum:
@vedhavyas this test suite requires an ethereum node running, that belongs to the integration tests domain. On the other hand, this is supposed to be a functional one as there is no access to code base during the test, so that is why I decided to create own tag.

Issue #682, opened 2019-01-23
[POW testing] Ropsten


vimukthi-git:
<This is intended as a guide when thinking about or creating issues for go-centrifuge>

#### Testing
Testworld

#### Acceptance Criteria
go-centrifuge works without a problem on Ropsten

PR #683, opened 2019-01-23
Ropsten

vimukthi-git:
Issue Link: #682

### DIFF
--- a/build/configs/default_config.yaml
+++ b/build/configs/default_config.yaml
@@ -15,7 +15,7 @@ networks:
       anchorRepository: ""
       paymentObligation: ""
 
-  # Main development testnet network
+  # Main development testnet network (Rinkeby)
   russianhill:
     # Numeric ID of the Centrifuge network
     id: 51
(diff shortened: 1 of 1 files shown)


build/configs/default_config.yaml
```diff
@@ -55,6 +56,23 @@ networks:
       anchorRepository: "0x444f649e307442e76ccf737466e52f1609b98260"
       paymentObligation: "0x0417eb37941164368401D666984cED7694ABcBb1"
 
+  # Ropsten test network
+  doloresheights:
```
mikiquantum: This neighborhood is not as well known 😄 I would use something like `dogpatch` instead

vimukthi-git: I liked the name .. but I can change

rdinicut:
AnchorRepository: 0x31f760e704F37755d1Ca4fB5eDA6D755E107Ae11
IdentityRegistry: 0x55160B12091c41E7C41f0BA5Ae925a0426c4aAEA
IdentityFactory: 0x546EF82C503ef306e1fA4775f30068b4d9e20814
PaymentObligation: 0x3Ea32C24A525E7e8a6191C923793052F77b65936

PR #685, opened 2019-01-23
identity: initial commit for new identity

manuelpolzhofer:
**Initial Commit**
the initial commit includes the following:
- go-bindings for new factory and identity contract
- importing two versions of our smart contracts
  - used fork in my repo
- bootstrapper
    - intermediated solution to deploy new contracts together with the old ones
- identity service
- create identity with new service
- calculate identity based on address and nonce

**Next Commit**
using transactionStatusTask for Create Identity


identity/did/service_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func TestCreateIdentity_successful(t *testing.T) {
+	service := ctx[BootstrappedDIDService].(Service)
+	did, _, err := service.CreateIdentity(context.Background())
+	assert.Nil(t, err, "create identity should be successful")
+
+	client := ctx[ethereum.BootstrappedEthereumClient].(ethereum.Client)
+
+	time.Sleep(2000 * time.Millisecond)
```
manuelpolzhofer: will be removed after Create Identity using a transaction status task

mikiquantum: When is that going to be added? As you see, this is error prompt as tx can take longer than 2 secs to be confirmed (if we are willing to point integration tests to rinkeby). Just want to make sure it is not going to linger around for long.

identity/did/bootstrapper.go
```diff
@@ -0,0 +1,147 @@
...
+	gp := os.Getenv("GOPATH")
+	projDir := path.Join(gp, "src", "github.com", "centrifuge", "go-centrifuge")
+
+	smartContractDir := path.Join(projDir, "vendor", "github.com", "manuelpolzhofer", "centrifuge-ethereum-contracts")
+	smartContractDirStandard := path.Join(projDir, "vendor", "github.com", "centrifuge", "centrifuge-ethereum-contracts")
+
+	os.Setenv("CENT_ETHEREUM_CONTRACTS_DIR", smartContractDir)
+
+	migrationScript := path.Join(projDir, "build", "scripts", "migrate.sh")
+	_, err := exec.Command(migrationScript, projDir).Output()
```
mikiquantum: Can we pass it instead as part of the envs of the `exec.Command`? Instead of modifying the global env. Look at: https://github.com/centrifuge/go-centrifuge/blob/develop/cmd/centrifuge_cmd_test.go#L29-L33

manuelpolzhofer: As discussed. The registry is not existing any more and the sleep func will be removed as soon as #663 is merged.

manuelpolzhofer: I would need to modify the testingutils package to make it work. 

I will keep it as it is now, because the code will be removed anyway

PR #690, opened 2019-01-23
identity: AddKey and GetKey support added

manuelpolzhofer:
- AddKey implemented
- GetKey implemented
- additional tests implemented


identity/did/identity.go
```diff
@@ -18,3 +29,128 @@ func NewDID(address common.Address) DID {
...
+
+}
+
+// NewIdentity creates a instance of an identity
+func NewIdentity(config id.Config, client ethereum.Client) Identity {
+	return identity{config: config, client: client}
+}
+
+// TODO: will be replaced with statusTask
+func waitForTransaction(client ethereum.Client, txHash common.Hash, txStatus chan *ethereum.WatchTransaction) {
```
manuelpolzhofer: waiting for #663 to get merged :)

identity/did/service.go
```diff
@@ -14,17 +14,17 @@ var log = logging.Logger("identity")
 
 // Service is the interface for identity related interactions
 type Service interface {
-	CreateIdentity(ctx context.Context) (id *DID, confirmations chan *identity.WatchIdentity, err error)
+	CreateIdentity(ctx context.Context) (id *DID, confirmations chan *id.WatchIdentity, err error)
```
mikiquantum: Shouldn't we implement as well `FindIdentity` function? Or that would be done after? I think it makes sense to have added it when creating the identity was done, or now.

ethereum/transaction_status_task.go
```diff
@@ -21,12 +21,20 @@ const (
 	TransactionTxHashParam string = "TxHashParam"
 	// TransactionAccountParam contains the name  of the account
 	TransactionAccountParam  string = "Account ID"
-	transactionStatusSuccess uint64 = 1
+	// TransactionStatusSuccess contains the flag for a successful receipt.status
+	TransactionStatusSuccess uint64 = 1
```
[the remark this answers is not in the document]
manuelpolzhofer: The transactions is the package for centrifuge transactions. The flag is for a successful Ethereum transaction.

identity/did/service.go
manuelpolzhofer: I implemented this method:

```golang
func (i identity) bindContract(did DID) (contract, error) {
	contract, err := NewIdentityContract(did.toAddress(), i.client.GetEthClient())
	if err != nil {
		return nil, errors.New("Could not bind identity contract: %v", err)
	}

 	return contract, nil

 }
```

The user needs to provide the DID for every identity interaction. The first thing every method will do is to bind the provided DID with the contract. I could add an additional check by getting the byteCode of the contract to check if it is an identity contract.

mikiquantum: That is fine for now, Can we then create a public method called FindIdentity that invokes `bindContract` internally and returns the identity object?

ethereum/transaction_status_task.go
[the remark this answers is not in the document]
vimukthi-git: Lets not make it public then

identity/did/identity.go
vimukthi-git: @manuelpolzhofer You can merge this without that PR using the older way of doing async once comments are addressed. I have a design for that that changes things for this and anchoring anyways. The design basically introduces the concept of a TX manager. Instead of tasks managing transactions.

PR #698, opened 2019-01-28
identity: added execute method

manuelpolzhofer:
- added execute method
  - abi encoding of parameters
  - check if parameters and method name is correct
- added raw execute method

- added go-bindings for new anchor contract
- tested execute method with new anchor contract

I used an own file for execute tests because i needed to include information about another contract.


identity/did/identity.go
```diff
@@ -34,8 +37,15 @@ func NewDIDFromString(address string) DID {
 type Identity interface {
 	// AddKey adds a key to identity contract
 	AddKey(did *DID, key Key) (chan *ethereum.WatchTransaction, error)
+
 	// GetKey return a key from the identity contract
 	GetKey(did *DID, key [32]byte) (*KeyResponse, error)
+
+	// RawExecute calls the execute method on the identity contract
```
manuelpolzhofer: Should we keep RawExecute public?

mikiquantum: Is there a use case for it? If not let's make it private.

identity/did/bootstrapper.go
```diff
@@ -63,6 +63,10 @@ func getFactoryAddress() common.Address {
 
 }
 
+func getAnchorAddress() common.Address {
```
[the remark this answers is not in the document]
manuelpolzhofer: Yes. I will remove it. (They are currently only here because we are using two versions of the smart contracts)

identity/did/identity.go
manuelpolzhofer: I will make it private.

mikiquantum: ok!

PR #701, opened 2019-01-29
Fixed generate keys + more

mikiquantum:
Issue Link: <Github link>


build/scripts/tests/run_testworld.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Running Testworld"
 
-cleanup="ls testworld/peerconfigs/* | grep testworld | grep -v README.md | tr -d : | xargs rm -rf"
```
vimukthi-git: this was a mistake I also realised yesterday. Its fixed in `develop`

testworld/document_consensus_test.go
```diff
@@ -75,10 +75,10 @@ func addExternalCollaborator_withinHost(t *testing.T, documentType string) {
 	getDocumentAndCheck(bob.httpExpect, b, documentType, params)
 	nonExistingDocumentCheck(bob.httpExpect, c, documentType, params)
 
-	// Bob updates invoice and shares with Charlie as well
-	res = updateDocument(bob.httpExpect, bob.id.String(), documentType, http.StatusOK, docIdentifier, updatedDocumentPayload(documentType, []string{a, c}))
+	// b updates invoice and shares with c as well
```
vimukthi-git: 🥇 
My bad

testworld/park_test.go
```diff
@@ -33,3 +36,42 @@ func TestHost_Happy(t *testing.T) {
...
 }
+
+func TestHost_RestartWithAccounts(t *testing.T) {
+	t.Parallel()
+	// Name can be randomly generated
+	tempHostName := "Sleepy"
+	bootnode, err := doctorFord.bernard.p2pURL()
+	assert.NoError(t, err)
+	sleepyHost := doctorFord.createTempHost(tempHostName, doctorFord.twConfigName, defaultP2PTimeout, 8088, 38208, true, true, []string{bootnode})
+	doctorFord.addNiceHost(tempHostName, sleepyHost)
```
vimukthi-git: why do you need to add it here?

vimukthi-git: you can just not add it, since its not useful to be managed by doctorFord, 
the test can create, start and then kill it at the end.

mikiquantum: right now is needed as if there is no nice hosts then `getHostTestSuite` will fail and I do not want to modify more code here.

Issue #678, opened 2019-01-22
Fill CoreDocument with access token + Sign origin


mikiquantum:
When provided on create/update api payload create token and fill coredocument + signature

#### Testing
<Describe required tests: Unit, Integration, Testworld or Functional>

#### Acceptance Criteria
<What is the acceptance criteria of this ticket?>

#### Dependencies and affected external components
<Outline any dependencies and components that could be affected by this>

PR #709, opened 2019-01-31
[WIP] Fill CoreDoc with AccessToken

charleenfei:
Issue Link: #678


coredocument/accesstoken.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return nil, err
+	}
+	docId, err := hexutil.Decode(payload.GetDocumentIdentifier())
+	if err != nil {
+		return nil, err
+	}
+	// assemble access token message to be signed
+	tm := append(tokenIdentifier, granterId...,)
+	tm = append(grantee, roleId)
```
vedhavyas: You are replacing

`tm := append(tokenIdentifier, granterId...,)`

with 

`tm = append(grantee, roleId)`

first param of append should be tm since you want to append new data to tm

identity/identity.go
```diff
@@ -74,6 +74,11 @@ func (c CentID) BigInt() *big.Int {
 	return utils.ByteSliceToBigInt(c[:])
 }
 
+// Bytes returns CentID in bytes
+func (c CentID) Bytes() []byte {
```
vedhavyas: You can always use `c[:]` to get the bytes here. We are using this multiple places.

check out
`coredocument.go#PrepareNewVersion` function

charleenfei:
closing this WIP PR due to coredoc refactoring.

Issue #625, opened 2019-01-09
createconfig: missing p2p keys usage and creation


vedhavyas:
Created config:

```
...
keys:
  ethauth:
    privatekey: /Users/vedhavyas/datadir/test/ethauth.key.pem
    publickey: /Users/vedhavyas/datadir/test/ethauth.pub.pem
  p2p:
    privatekey: /Users/vedhavyas/datadir/test/p2p.key.pem
    publickey: /Users/vedhavyas/datadir/test/p2p.pub.pem
  signing:
    privatekey: /Users/vedhavyas/datadir/test/signing.key.pem
    publickey: /Users/vedhavyas/datadir/test/signing.pub.pem
...
```

but the p2p.*.pem keys are not created.

Solution: 
Seems like we are not using p2p keys anymore but rather using signing keys.  So, we either remove p2p keys from the node and let the libp2p create keys or we create p2p keys during the `createconfig` and use them for basic host.

I'm leaning towards first one

vimukthi-git:
done

PR #710, opened 2019-01-31
Key separation

mikiquantum:
Issue Link: <Github link>


cmd/common.go
```diff
@@ -34,10 +34,12 @@ func createIdentity(ctx context.Context, idService identity.Service) (identity.C
 }
 
 func generateKeys(config config.Configuration) {
-	p2pPub, p2pPvt := config.GetSigningKeyPair()
+	p2pPub, p2pPvt := config.GetP2PKeyPair()
+	signPub, signPvt := config.GetSigningKeyPair()
 	ethAuthPub, ethAuthPvt := config.GetEthAuthKeyPair()
-	crypto.GenerateSigningKeyPair(p2pPub, p2pPvt, "ed25519")
-	crypto.GenerateSigningKeyPair(ethAuthPub, ethAuthPvt, "secp256k1")
+	crypto.GenerateCryptoKeyPair(p2pPub, p2pPvt, "ed25519")
```
lucasvo: Actually the name signing key is accurate because these are signature keys. The algorithm secp256k1 and ed25519 are sig schemes (as opposed to encryption schemes).

I would not rename it or just get rid of "Crypto" -> "GenerateKeyPair"

cmd/common.go
```diff
@@ -34,10 +34,12 @@ func createIdentity(ctx context.Context, idService identity.Service) (identity.C
 }
 
 func generateKeys(config config.Configuration) {
-	p2pPub, p2pPvt := config.GetSigningKeyPair()
+	p2pPub, p2pPvt := config.GetP2PKeyPair()
```
lucasvo: Good! When I brought this up in the housekeeping meeting, I was precisely worried that in a place like this, we'd do mainAccount.GetP2PKeyPair().

mikiquantum:
@lucasvo @vimukthi-git updated pr with some comments and final fixes

build/resources/centrifuge_example.yaml
```diff
@@ -8,10 +8,10 @@ nodePort: 8082
 identityId: "0x010101010101"
 
 keys:
-  signing:
-    publicKey: ../../build/resources/signingKey.pub.pem
-    privateKey: ../../build/resources/signingKey.key.pem
   p2p:
+    publicKey: ../../build/resources/p2pKey.pub.pem
```
vimukthi-git: I believe you are fixing one of our `good first issue`s here :D 
#625

Issue #519, opened 2018-11-29
Uniqueness of documents for NFT


manuelpolzhofer:
Requirement: P2P Layer supports the [uniqueness fields as laid out in the P2P Epic](https://hackmd.io/Wy78jBRfQWSz6KLU21pk4A?both#NFT-Support).

It is only allowed once to mint an NFT for a document.

Therefore  information about the NFT will be added inside the document.

According to [privacy-enabled NFT paper](https://github.com/centrifuge/paper-privacy-enabled-nfts/blob/master/main.tex#L189).
```json
"documentIdentifier": "0x...",
    // [...]
 	"onChainRegistrations": [
		// Registration:
		{
			"type": "0x...", // uint256
			"registry": "0x...", // address (20byte)
			"tokenID": "0x..." // uint256
			"concatRegistration": "..." // string
		},
	]
	// [...]
```

## Workflow
1. User A wants to mint an NFT.
2. User A creates a new random NFT tokenID.
3. User A adds the type,registry and tokenID to the document.
4. Request Signature from all other collaborators
5. Commit updated document to anchorRepository contract.
6. Call NFT mint method with additional proof for tokenId
 
```javascript

require(
MerkleProofSha256.verifyProof(
_proofs[tokenIdIdx],
_merkleRoot,
//same tokenId can't be minted twice
_hashLeafData(supportedFields_[tokenIdIdx], tokenId, _salts[tokenIdIdx])
),
      "merkle tree needs to validate the tokenId"
    );
    
    
```

vimukthi-git:
@vedhavyas could you comment on whats complete here? may be we can close this?

PR #711, opened 2019-01-31
Fix NFT rule/role update order

vimukthi-git:
Issue Link: #519


nft/ethereum_payment_obligation.go
```diff
@@ -143,79 +129,113 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
...
-		log.Infof("Sent off ethTX to mint [tokenID: %x, anchor: %x, registry: %x] to payment obligation contract. Ethereum transaction hash [%x] and Nonce [%v] and Check [%v]",
-			requestData.TokenID, requestData.AnchorID, requestData.To, ethTX.Hash(), ethTX.Nonce(), ethTX.CheckNonce())
-		log.Infof("Transfer pending: 0x%x\n", ethTX.Hash())
+		model, err = s.docSrv.DeriveFromCoreDocument(cd)
+		if err != nil {
+			errOut <- err
+		}
 
-		res, err := ethereum.QueueEthTXStatusTask(accountID, txID, ethTX.Hash(), true, s.queue)
+		_, _, err = s.docSrv.Update(contextutil.WithTX(ctx, txID), model)
```
vimukthi-git: Doing this rule update and anchoring before minting is the main change

nft/payment_obligation.go
```diff
@@ -2,8 +2,42 @@ package nft
...
+
+// BigInt converts tokenID to big int
+func (t TokenID) BigInt() *big.Int {
+	return utils.ByteSliceToBigInt(t[:])
+}
+
+// URI gets the URI for this token
+func (t TokenID) URI() string {
+	// TODO please fix this
+	return "http:=//www.centrifuge.io/DUMMY_URI_SERVICE"
```
lucasvo: Will you fix this? I think here we should read from the ethereum contract to get the URI. This should be part of the NFT epic.

nft/ethereum_payment_obligation.go
manuelpolzhofer: As discussed in person. We need to wait for the document update because the latest current version is needed and the tokenID needs to be part of the merkle proof.

nft/ethereum_payment_obligation.go
```diff
@@ -111,111 +115,149 @@ func (s *ethereumPaymentObligation) prepareMintRequest(ctx context.Context, docu
...
 
-// sendMintTransaction sends the actual transaction to mint the NFT
-func (s *ethereumPaymentObligation) sendMintTransaction(
-	cid identity.CentID,
-	contract ethereumPaymentObligationContract,
-	opts *bind.TransactOpts,
-	requestData MintRequest,
-	registry common.Address,
-	documentID []byte) (uuid.UUID, error) {
+		cd, err := model.PackCoreDocument()
```
manuelpolzhofer: After the model refactoring, we can call the model directly here correct?

transactions/manager.go
```diff
@@ -12,38 +12,69 @@ import (
...
+type Manager interface {
 	// ExecuteWithinTX executes a transaction within a transaction
 	ExecuteWithinTX(ctx context.Context, accountID identity.CentID, existingTxID uuid.UUID, desc string, work func(accountID identity.CentID, txID uuid.UUID, txMan Manager, err chan<- error)) (txID uuid.UUID, done chan bool, err error)
-
-	// CreateTransaction
-	// TODO [TXManager] remove this once TX Manager update is complete
-	CreateTransaction(accountID identity.CentID, desc string) (*Transaction, error)
 	GetTransaction(accountID identity.CentID, id uuid.UUID) (*Transaction, error)
-	SaveTransaction(tx *Transaction) error
+	UpdateTaskStatus(accountID identity.CentID, id uuid.UUID, status Status, taskName, message string) error
```
vimukthi-git: This is a key txManager update in this pr, we use task status instead of tx status to track progress now in tasks

documents/anchor_task.go
```diff
@@ -119,31 +119,41 @@ func (d *documentAnchorTask) RunTask() (res interface{}, err error) {
...
+	tr, err := tq.EnqueueJob(documentAnchorTaskName, params)
+	if err != nil {
+		return nil, err
+	}
+
+	return tr, nil
+}
+
+// CreateAnchorTransaction creates a transaction for anchoring a document using transaction manager
+func CreateAnchorTransaction(txMan transactions.Manager, tq queue.TaskQueuer, self identity.CentID, txID uuid.UUID, documentID []byte) (uuid.UUID, chan bool, error) {
```
vimukthi-git: this would be used by all doc updates to create or update an existing tx with task status.

vimukthi-git: need to add task status as pending here

vimukthi-git:
@manuelpolzhofer yes those are different parts of NFT uniqueness, that we need to implement. This one is only about getting the flow right to provide proofs of uniqueness to the contract.

nft/ethereum_payment_obligation.go
```diff
@@ -111,111 +115,149 @@ func (s *ethereumPaymentObligation) prepareMintRequest(ctx context.Context, docu
...
+	}, done, nil
 }
 
-// OwnerOf returns the owner of the NFT token on ethereum chain
-func (s *ethereumPaymentObligation) OwnerOf(registry common.Address, tokenID []byte) (owner common.Address, err error) {
-	contract, err := s.bindContract(registry, s.ethClient)
-	if err != nil {
-		return owner, errors.New("failed to bind the registry contract: %v", err)
-	}
+func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID, documentID []byte, registryAddress, depositAddress string, proofFields []string) func(accountID identity.CentID, txID uuid.UUID, txMan transactions.Manager, errOut chan<- error) {
```
manuelpolzhofer: I would split the minter funcs maybe into smaller funcs.

documents/service.go
```diff
@@ -58,10 +58,10 @@ type Service interface {
 	ReceiveAnchoredDocument(ctx context.Context, model Model, senderID []byte) error
 
 	// Create validates and persists Model and returns a Updated model
-	Create(ctx context.Context, model Model) (Model, uuid.UUID, error)
+	Create(ctx context.Context, model Model) (Model, uuid.UUID, chan bool, error)
```
vedhavyas: not sure why we need to have the chan here since this is ignored mostly. I would prefer not having this part of the interface

nft/ethereum_payment_obligation.go
vedhavyas: cyclomatic complexity of this function is more than optimum. Why did we move this here instead of having it part of the documents package? I prefer NFT package doesn't know about the documents package. Mpve this to documents package pls

nft/ethereum_payment_obligation_test.go
```diff
@@ -226,13 +195,13 @@ func TestPaymentObligationService(t *testing.T) {
...
 			ctxh := testingconfig.CreateAccountContext(t, &mockCfg)
-			_, err := service.MintNFT(ctxh, decodeHex(test.request.Identifier), test.request.RegistryAddress, test.request.DepositAddress, test.request.ProofFields)
+			_, _, err := service.MintNFT(ctxh, decodeHex(test.request.Identifier), test.request.RegistryAddress, test.request.DepositAddress, test.request.ProofFields)
 			if test.err != nil {
 				assert.Equal(t, test.err.Error(), err.Error())
 			} else if err != nil {
 				panic(err)
 			}
-			docService.AssertExpectations(t)
+			//docService.AssertExpectations(t)
```
vedhavyas: uncomment or remove pls

nft/ethereum_payment_obligation_test.go
```diff
@@ -226,13 +195,13 @@ func TestPaymentObligationService(t *testing.T) {
...
 			paymentOb.AssertExpectations(t)
 			idService.AssertExpectations(t)
 			//ethClient.AssertExpectations(t)
```
vedhavyas: this one too, pls

documents/service.go
vimukthi-git: its needed so that any code that uses this service function can wait on the update in an async way or the proper go way, instead of polling db. We allow polling of API for outside users but no need to do that inside the node. Therefore this has to be a convention we follow for all aync background funcs.

nft/ethereum_payment_obligation.go
vimukthi-git: I thing this is refactoring we can consider when we have the entire flow ready, `premature optimisation is the root of all evil` :)

vimukthi-git:
Merging this now, Follow ups - 
- Call smart contract to get the token URL
- Refactor `mint` closure method to be smaller once we have entire flow
- More tests for txManager edge cases

cc @lucasvo @manuelpolzhofer @vedhavyas

nft/ethereum_payment_obligation.go
vedhavyas: I don't consider this to be a premature optimization @vimukthi-git 
I'm talking about breaking down the function to smaller chunks so that its easier to test and understand

vimukthi-git: Agree, just that it seems kind of arbitrary to break it down right now, as only testing we can do is to mock things out individually and see if the function uses mocks correctly. I would love to think about this separately and do a proper test when completing the flow.

PR #712, opened 2019-02-01
identity: bind identity service to did

manuelpolzhofer:
-  bind identity service to did

### DIFF
--- a/identity/did/bootstrapper.go
+++ b/identity/did/bootstrapper.go
@@ -20,9 +20,6 @@ type Bootstrapper struct{}
 // BootstrappedDIDService stores the id of the service
 const BootstrappedDIDService string = "BootstrappedDIDService"
 
-// BootstrappedDID stores the id of the identity
-const BootstrappedDID string = "BootstrappedDID"
-
 var smartContractAddresses *config.SmartContractAddresses
 
 // Bootstrap initializes the factory contract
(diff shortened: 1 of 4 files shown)


identity/did/bootstrapper.go
```diff
@@ -48,7 +48,14 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
...
 
-	identity := NewIdentity(cfg, client)
+	//TODO config needs to store correct did format
+	id, err := cfg.GetIdentityID()
+	if err != nil {
+		return err
+	}
+	did := NewDID(common.BytesToAddress(id))
+
+	identity := NewIdentity(cfg, client, &did)
```
mikiquantum: There will be multiple identities under the same host. Why do we need to bootstrap here? Is this `identity` only used by the p2p layer? In which case we should rename it.

manuelpolzhofer: I thought multi-account is abstracted in that way, that the config already includes the identity address of the current account. 

I thought the bootstrapper would be called for every different identity in the multi-account scenario. If not I will remove it.

I added it to the bootstrapper to easier use the identity later on.

mikiquantum: > I thought multi-account is abstracted in that way, that the config already includes the identity address of the current account.

Not at bootstrap level. Accounts are dynamically created, loaded per request basis.

> I thought the bootstrapper would be called for every different identity in the multi-account scenario. If not I will remove it.

Nope, bootstrapper is only called at start up time. Then dynamically load the accounts per request. I would remove this as it is confusing. Within your code you should use the `context.Context` to identify the ID to use.

manuelpolzhofer: I already looked it up. Thanks for the clarification.

manuelpolzhofer: I will add the changes for context in the next PR

PR #723, opened 2019-02-05
go-centrifuge Slides

vimukthi-git:
Issue Link: <Github link>


vedhavyas:
Let's move this to a new repo called presentations. We can have a single repo for all the talks/presentations done so far and any in the future there.

CC: @lucasvo

vimukthi-git:
@vedhavyas there are somethings we need to discuss before doing that, part of it I was going to show in the presentation it self.
I don't know if you read the presentation but part of the reason why this here is explained in the slides them selves. Also slides refer to code in the repo, so they won't go out of sync as a reference.

presentation/presentation.slide
```diff
@@ -0,0 +1,205 @@
...
+* Wait but first about Errors :)
+
+We have a few conventions around error handling,
+- All errors should be checked and propagated up the call chain (no panic or fatal)
+- Use constant error values
+- Use *github.com/centrifuge/go-centrifuge/errors* package to create errors
+
+* Bootstrappers
+
+.code ../bootstrap/bootstrapper.go /Bootstrapper/,/^}/
```
vimukthi-git: @vedhavyas this is an example of a code reference

presentation/presentation.slide
```diff
@@ -0,0 +1,254 @@
...
+- A model must not have any runtime dependencies such as database. Using a compile time dep such as a lib is fine
+- A model must wrap all coredocument operations
+- No direct access to internal fields must be allowed
+
+* Document Services
+
+- A Service must be defined to handle all business logic related to a single Model.
+- Some service functions are generic enough for them to be in the `DefaultService`.
+
+* Service Interface
```
charleenfei: Validators can be added as a slide to this presentation :)

manuelpolzhofer: my additional slide ideas:
- ethereum testnet/ centrifuge testnet
- multi-account
- command line interface

presentation/presentation.slide
vedhavyas: You can always refer this from another repo AFAIK

vimukthi-git: I checked, it wouldn't include the code blocks with another repo link. Anyway these slides are intended as documentation, Its not a presentation per se. So I think we can just keep it in the repo it self. Its also easier to keep it update with code changes this way.

presentation/presentation.slide
vimukthi-git: @manuelpolzhofer I believe what you mean by `command line interface` is how it works technically? I will make a note of that.
Regarding `multi-account` there is already a slide here `Accounts` for that, what would you like it to be covered?

PR #724, opened 2019-02-05
identity: create identity uses txManager

manuelpolzhofer:
- added new dependencies to service
  - txManager and queue

- identity byte code compare
 - after the identity contract is created the on-chain bytecode is checked


identity/did/bootstrapper.go
```diff
@@ -42,7 +46,17 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
 		return err
 	}
 
-	service := NewService(cfg, factoryContract, client)
+	txManager, ok := context[transactions.BootstrappedService].(transactions.Manager)
+	if !ok {
+		return errors.New("transactions repository not initialised")
+	}
+
+	if _, ok := context[bootstrap.BootstrappedQueueServer]; !ok {
```
mikiquantum: Can you use the same format as above?
```suggestion
      queueSrv, ok := context[bootstrap.BootstrappedQueueServer].(*queue.Server)
      if !ok {
         return errors.New("queue hasn't been initialized")
      }
```

identity/did/service.go
```diff
@@ -41,34 +51,104 @@ func CalculateCreatedAddress(address common.Address, nonce uint64) common.Addres
...
-	did := NewDID(identityAddress)
+	txID, done, err := s.txManager.ExecuteWithinTX(context.Background(), idConfig.ID, uuid.Nil, "Check TX status", s.createIdentityTX(opts))
+	if err != nil {
+		return nil, err
+	}
+
+	isDone := <-done
+	// non async task
+	if !isDone {
+		err = s.txManager.WaitForTransaction(idConfig.ID, txID)
```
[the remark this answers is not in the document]
vimukthi-git: done = false for failure
done = true for sucess

identity/did/service.go
```diff
@@ -41,34 +51,104 @@ func CalculateCreatedAddress(address common.Address, nonce uint64) common.Addres
...
+		return nil, err
+	}
+
+	identityAddress, err := s.calculateIdentityAddress(ctx)
+	if err != nil {
+		return nil, err
+	}
 
-	did := NewDID(identityAddress)
+	txID, done, err := s.txManager.ExecuteWithinTX(context.Background(), idConfig.ID, uuid.Nil, "Check TX status", s.createIdentityTX(opts))
```
mikiquantum:
```suggestion
	txID, done, err := s.txManager.ExecuteWithinTX(context.Background(), idConfig.ID, uuid.Nil, "Identity Creation TX status", s.createIdentityTX(opts))
```
Or similar, so there is context of which action has invoked the execute. Ideally we would use constants here. What do you think?

identity/did/service.go
```diff
@@ -41,34 +51,104 @@ func CalculateCreatedAddress(address common.Address, nonce uint64) common.Addres
 	return crypto.CreateAddress(address, nonce)
 }
 
-func (s *service) CreateIdentity(ctx context.Context) (id *DID, confirmations chan *id.WatchIdentity, err error) {
-	opts, err := s.client.GetTxOpts(s.config.GetEthereumDefaultAccountName())
+func (s *service) createIdentityTX(opts *bind.TransactOpts) func(accountID id.CentID, txID uuid.UUID, txMan transactions.Manager, errOut chan<- error) {
+	return func(accountID id.CentID, txID uuid.UUID, txMan transactions.Manager, errOut chan<- error) {
+		ethTX, err := s.client.SubmitTransactionWithRetries(s.factoryContract.CreateIdentity, opts)
+		fmt.Println("trans submitted")
```
mikiquantum: remove

identity/did/service.go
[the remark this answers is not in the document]
manuelpolzhofer: ahh, I misunderstood the flag.
I thought `true==transactionFinished false==pending`.

I will fix

identity/did/service.go
manuelpolzhofer: I will add constant. Good idea.

identity/did/service.go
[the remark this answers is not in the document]
vimukthi-git: Agree its misleading
can you rename the return param to something that gives the correct meaning?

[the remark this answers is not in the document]
vimukthi-git: may be `done` -> `success`

identity/did/service.go
manuelpolzhofer: @mikiquantum I addressed the mentioned issues

Issue #716, opened 2019-02-03
Make CoreDocument methods as Model methods


mikiquantum:
Details: https://hackmd.io/YaTDM7csQ0WSefALCgCKbQ?both

Currently, the coredocument interact directly with a coredocument. These interactions should happen through the [model-validator architecture](https://godoc.org/github.com/centrifuge/go-centrifuge/documents).

Interfaces for these functionalities should be written and implemented in the existing document model, with methods on the interface-implementing structs.

- [ ] Coredocument.go should be refactored to follow the above design
- [ ] Doc.go should be updated

#### Testing
fix tests that break

#### Acceptance Criteria
<What is the acceptance criteria of this ticket?>

#### Dependencies and affected external components
<Outline any dependencies and components that could be affected by this>

PR #725, opened 2019-02-05
CoreDocModel + PrepareNewVersion

charleenfei:
First of many PRs addressing issue #716.

This PR moves a copy of the ReadACL files into the documents package so the methods can be accessed by the CoreDocModel. This will be refactored as the CoreDoc refactor progresses.


documents/model.go
```diff
@@ -25,3 +31,134 @@ type Model interface {
...
+func (m *CoreDocumentModel) GetDocument() (*coredocumentpb.CoreDocument, error) {
+	cd := m.document
+	if cd == nil {
+		return nil, errors.New("error getting document in CoreDocModel")
+	}
+	return m.document, nil
+}
+
+// SetDocument set the coredocument in the CoreDocumentModel to an updated coredocument
+func (m *CoreDocumentModel) SetDocument(ncd *coredocumentpb.CoreDocument) *CoreDocumentModel {
```
vedhavyas: Let's not expose the CoreDocument through setters, all mutations on the CD should be done through CoreDocModel

documents/model.go
```diff
@@ -25,3 +31,134 @@ type Model interface {
...
+		CurrentVersion:     id,
+		NextVersion:        utils.RandomSlice(32),
+	}
+	return &CoreDocumentModel{
+		cd,
+	}
+}
+
+// GetDocument returns the coredocument from the CoreDocumentModel
+func (m *CoreDocumentModel) GetDocument() (*coredocumentpb.CoreDocument, error) {
```
vedhavyas: Lets not return a pointer to coredocumentpb.Coredocument since we can mutate it after getting it from this function. Return a value instead

documents/read_acls_test.go
```diff
@@ -0,0 +1,125 @@
+// +build unit
+
+package documents
+
+import (
+	"testing"
+
+	"github.com/centrifuge/go-centrifuge/coredocument"
```
vedhavyas: since we moved the read acls to documents package, update the tests to use those functions

charleenfei: I want to wait to update these tests because this whole package will be refactored anyway in a future PR which deals with the ReadACL methods and how they can be refactored into the CoreDocModel

vedhavyas: At the moment, there is Read ACL stuff in documents and coredocument, only one should the source. So lets either remove the read acl stuff from documents package and expose required methods in the core document.

documents/model.go
```diff
@@ -25,3 +31,132 @@ type Model interface {
...
+		CurrentVersion:     id,
+		NextVersion:        utils.RandomSlice(32),
+	}
+	return &CoreDocumentModel{
+		cd,
+	}
+}
+
+// GetDocument returns the coredocument from the CoreDocumentModel
+func (m *CoreDocumentModel) GetDocument() (coredocumentpb.CoreDocument, error) {
```
mikiquantum: We said we do not need this.

documents/model.go
```diff
@@ -25,3 +31,132 @@ type Model interface {
...
+
+// CoreDocumentModel contains methods which handle all interactions mutating or reading from a core document
+// Access to a core document should always go through this model
+type CoreDocumentModel struct {
+	Document *coredocumentpb.CoreDocument
+}
+
+// NewCoreDocModel returns a new CoreDocumentModel
+// Note: collaborators and salts are to be filled by the caller
+func newCoreDocModel() *CoreDocumentModel {
```
mikiquantum: I think this needs to be public

documents/model.go
```diff
@@ -25,3 +31,132 @@ type Model interface {
...
+	ncd.Roles = m.Document.Roles
+	ncd.ReadRules = m.Document.ReadRules
+	addCollaboratorsToReadSignRules(ncd, ucs)
+
+	err = ndm.fillSalts()
+	if err != nil {
+		return nil, err
+	}
+
+	cd, err := m.GetDocument()
```
mikiquantum:
```suggestion
	cd  := m.Document
```
Or better let's use the vars defined at the beginning of the method, as they are pointers.

PR #727, opened 2019-02-06
retry when txhash not found

vedhavyas:
Since infura load balances between multiple nodes, if the transaction is not propagated to the current node, it will return "not found". We should retry in this scenario instead of marking the transaction to be failed


vedhavyas:
@manuelpolzhofer agree! updated

PR #728, opened 2019-02-06
identity: added context to add key and execute method

manuelpolzhofer:
-  added context to add key and execute method
- removed stored did in identity service


vedhavyas:
Can we increase the test coverage for this @manuelpolzhofer ?

identity/did/identity.go
```diff
@@ -129,8 +132,28 @@ func logTxHash(tx *types.Transaction) {
...
+	tc, err := contextutil.Account(ctx)
+	if err != nil {
+		return did, err
+	}
+
+	addressByte, err := tc.GetIdentityID()
+	if err != nil {
+		return did, err
+	}
+	did = NewDID(common.BytesToAddress(addressByte))
```
vedhavyas: Do we need to expose `NewDID` ?

identity/did/identity.go
```diff
@@ -129,8 +132,28 @@ func logTxHash(tx *types.Transaction) {
...
+	return did, nil
+
+}
+
+func (i identity) AddKey(ctx context.Context, key Key) (chan *ethereum.WatchTransaction, error) {
```
vedhavyas: add comments pls

identity/did/identity.go
manuelpolzhofer: I have to think about how exactly this will replace CentID.

manuelpolzhofer:
@vedhavyas Unit tests are still missing. I will add additional tests?

identity/did/identity.go
manuelpolzhofer: I thought we only add comments as required only to the interface method and not to the non-public implementation.

identity/did/identity.go
vedhavyas: It should be multi-step
1. lets make CentID type to `common.Address` and rename it to `identity.DID`
2. Change all references of `did.DID` to `identity.DID` in did package and wherever used

identity/did/identity.go
vedhavyas: adding comments to interface explains what method should do or return but adding it to the implementation explains implementation specific details.

vedhavyas:
@manuelpolzhofer did you add the tests ?

PR #729, opened 2019-02-06
identity: execute method and add key using txManager

manuelpolzhofer:
please review #728 first


manuelpolzhofer:
renaming of the file causes the PR to look larger than it actually is.

identity/did/factory.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	identityAddress := CalculateCreatedAddress(factoryAddress, nonce)
+	log.Infof("Calculated Address of the identity contract: 0x%x\n", identityAddress)
+	return &identityAddress, nil
+}
+
+func (s *factory) isIdentityContract(identityAddress common.Address) error {
+	contractCode, err := s.client.GetEthClient().CodeAt(context.Background(), identityAddress, nil)
```
mikiquantum: Do we know in the case of upgradable smart contracts, if this returns the proxy contract or the abi implementation one? Otherwise this will not work once we support zos upgradable smart contracts

identity/did/factory_integration_test.go
```diff
@@ -0,0 +1,65 @@
...
+	assert.Nil(t, err, "create identity should be successful")
+
+	client := ctx[ethereum.BootstrappedEthereumClient].(ethereum.Client)
+
+	contractCode, err := client.GetEthClient().CodeAt(context.Background(), did.toAddress(), nil)
+	assert.Nil(t, err, "should be successful to get the contract code")
+
+	assert.Equal(t, true, len(contractCode) > 3000, "current contract code should be arround 3378 bytes")
+
+	fmt.Println(did)
```
mikiquantum: remove

identity/did/factory_service.go
```diff
@@ -2,68 +2,181 @@ package did
 
 import (
```
mikiquantum: I dont think the name of this file should be `factory_service`, instead service as we talked about?

identity/did/factory.go
manuelpolzhofer: As discussed. Identity doesn't have a proxy contract.

Anyway would be interesting.

@rdinicut do you know?

What would `eth.getCode` return in case of upgradeability?

rdinicut: it will return the bytecode without the contructor and initialisation code

rdinicut: I did not test that but I am assuming it will return the bytecode from the Proxy Zos contract,

PR #730, opened 2019-02-07
add nft uniqueness check

vedhavyas:
Add uniqueness check for NFT minting.

Proofs will be part of the other PR

### DIFF
--- a/nft/ethereum_payment_obligation.go
+++ b/nft/ethereum_payment_obligation.go
@@ -146,7 +146,9 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
 	}
 
 	registry := common.HexToAddress(registryAddress)
-	if isNFTMinted(cd.Nfts, registry.Bytes()) {
+	mt := getStoredNFT(cd.Nfts, registry.Bytes())
+	// check if the nft is successfully minted
+	if mt != nil && s.isNFTMinted(registry, mt.TokenId) {
 		return nil, nil, errors.NewTypedError(ErrNFTMinted, errors.New("registry %v", registry.String()))
 	}
 
@@ -154,7 +156,6 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
 	// We use context.Background() for now so that the transaction is only limited by ethereum timeouts
 	txID, done, err := s.txManager.ExecuteWithinTX(context.Background(), cid, uuid.Nil, "Minting NFT",
 		s.minter(ctx, tokenID, model, registry, depositAddress, proofFields))
-
 	if err != nil {
 		return nil, nil, err
 	}
@@ -163,6 +164,13 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
 		TransactionID: txID.String(),
 		TokenID:       tokenID.String(),
 	}, done, nil
+}
+
+func (s *ethereumPaymentObligation) isNFTMinted(registry common.Address, tokenID []byte) bool {
+	// since OwnerOf throws when owner is zero address,
+	// if err is not thrown, we can assume that NFT is minted
+	_, err := s.OwnerOf(registry, tokenID)
+	return err == nil
 }
 
 func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID, model documents.Model, registry common.Address, depositAddress string, proofFields []string) func(accountID identity.CentID, txID uuid.UUID, txMan transactions.Manager, errOut chan<- error) {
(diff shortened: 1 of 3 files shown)


nft/ethereum_payment_obligation.go
```diff
@@ -247,6 +257,25 @@ func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID,
 	}
 }
 
+func isNFTMinted(nfts []*coredocumentpb.NFT, registry []byte) bool {
```
manuelpolzhofer: Should we maybe check also if the NFT exists in the registry itself?

What would happen, if I want to mint an NFT  and the document update is successful but 
the Ethereum transaction fails?

It would not be possible to retry the minting because `isNFTMinted` would return true?

Maybe isNFTMinted should only return true if the registry is in the CD and the tokenID exists in the contract?

nft/ethereum_payment_obligation.go
```diff
@@ -247,6 +257,25 @@ func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID,
...
+	for _, nft := range nfts {
+		if bytes.Equal(nft.RegistryId[:20], registry) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func addNFT(cd *coredocumentpb.CoreDocument, registry, tokenID []byte) {
```
manuelpolzhofer: Maybe move the function to the coredocument package?

mikiquantum: I wouldn't move anything in to the coredoc package as it is going away soon. This code will be refactored with Charly's work to use the cd model instead.

vedhavyas: We are getting rid of coredocument package. To much dependency on a package for simple function might not be needed

nft/ethereum_payment_obligation.go
vedhavyas: That is not that simple. If we want to check the contract, we first need to connect document with NFT. Then verify if an NFT has been minted for this document through this registry.

I am a little out of sync with the smart contracts at the moment, so if we have a way to check this, then we can check that. But if we have that functionality, then we dont need to have this as a part of coredocument.

manuelpolzhofer: The method is in ethereumPaymentObligation
```golang
 func (s *ethereumPaymentObligation) OwnerOf(registry common.Address, tokenID []byte (owner common.Address, err error) 
```
The check would basically be `if address != "0x000000..."` the tokenID is minted.

We need to store it in the coredocument as well. Because in the future an additional proof is needed, that the tokenID/registry is part of the coredocument to proof uniqueness. 

But maybe we don't need the addtional check for now in the isNFTMinted method

nft/ethereum_payment_obligation.go
vedhavyas: But once we migrate to coreDocModel, we can make all the mutating functions as part of that model 👍

nft/ethereum_payment_obligation.go
vedhavyas: Nice!
Let me update.

nft/ethereum_payment_obligation.go
```diff
@@ -147,23 +166,21 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
...
-		// registry address
-		registry := common.HexToAddress(registryAddress)
-
-		model, err := s.docSrv.GetCurrentVersion(ctx, documentID)
-		if err != nil {
-			errOut <- err
-			return
-		}
-
 		cd, err := model.PackCoreDocument()
```
lucasvo: Why do we need to deal with the embedded data in this method? Shouldn't it be possible to create a new version without having to do this? Maybe a helper to do this per model would be good?

nft/ethereum_payment_obligation.go
```diff
@@ -132,11 +135,27 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
 	}
 
 	tokenID := NewTokenID()
+	model, err := s.docSrv.GetCurrentVersion(ctx, documentID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cd, err := model.PackCoreDocument()
```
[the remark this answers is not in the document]
vedhavyas: until we have the coredocmodel ready

nft/ethereum_payment_obligation.go
vedhavyas: prepare new version doesn't copy the data since we assume the data is going to change. We copy the data here to new version because the data is not changing but rather only read acl related stuff on cd.

This will change once we have the coredocmodel ready and all the mutations happens through that model itself

PR #731, opened 2019-02-07
DocModelRefactor pt. 2

charleenfei:
Issue Link: a continuation of #716 

This PR contains the rest of the methods from coredocument, the related tests, as well as the readACL methods (not including NFT related functionality) and testing.


documents/model.go
```diff
@@ -110,19 +123,353 @@ func (m *CoreDocumentModel) PrepareNewVersion(collaborators []string) (*CoreDocu
...
+		return err
+	}
+
+	doc.SigningRoot = tree.RootHash()
+	return nil
+}
+
+// AccountCanRead validate if the core document can be read by the account .
+// Returns an error if not.
+func (m *CoreDocumentModel) AccountCanRead(account identity.CentID) bool {
```
vedhavyas: Can you add `NFTOwnerCanRead` as well ?

documents/read_acls_test.go
```diff
@@ -2,124 +2,74 @@
...
-	// not the same owner
-	owner := common.BytesToAddress(utils.RandomSlice(20))
-	tr = mockRegistry{}
-	tr.On("OwnerOf", registry, tokenID).Return(owner, nil).Once()
-	validator = coredocument.NftValidator(tr)
-	err = validator.NFTOwnerCanRead(cd, registry, tokenID, account)
-	assert.Error(t, err)
-	tr.AssertExpectations(t)
-}
+//func Test_addNFTToReadRules(t *testing.T) {
```
vedhavyas: Why is the file deleted and tests commented out ?

charleenfei: because I have moved those methods into documents/model and the remaining tests cover functionalities which I have not moved yet into documents/model.

documents/model.go
charleenfei: I am not entirely clear how you imagined the NFT interactions with the CoreDocModel to be, in particular the token registry and how CoreDocuments are currently used in the ethereum payment obligation files. This is why I haven't moved these functionalities yet.

PR #733, opened 2019-02-07
identity: added isSignedWithPurpose support

manuelpolzhofer:
please review #729 first
- added support for isSignedWithPurpose
- added function to util package to convert an address to 32 bytes
  - starting with 12 zeros on the left


utils/tools.go
```diff
@@ -65,6 +65,19 @@ func CheckMultiple32BytesFilled(b []byte, bs ...[]byte) bool {
 	return true
 }
 
+// AddressTo32Bytes converts an address to 32 a byte array
+// The length of an address is 20 bytes. First 12 bytes are filled with zeros.
+func AddressTo32Bytes(address common.Address) [32]byte {
```
mikiquantum: This sounds like an interesting function to have just passed the `bytes` and pad up front will all the zeros (not just address specific). 

Just out of curiosity why padding upfront?

mikiquantum: Can we use shifting bytes instead of for loop? Sounds a bit more efficient.

manuelpolzhofer: In Solidity if you convert an address to byte32 the padding happens in the left.

vedhavyas: You can instead do something like this

```
data := append(make([]byte, 12, 12), address.Bytes()...)
return SliceToByte32(data)
```

This is much efficient than for loop IMO

manuelpolzhofer: hehe. I tried it already at did a benchmark test.
The current solution is faster

PR #736, opened 2019-02-08
NFT uniqeness proofs

vedhavyas:
Add NFT proofs
- Uniqueness proof
- NFT read access proof

More here - https://hackmd.io/ps0X0jRkRHaACP6aPtGzag


coredocument/read_acls.go
```diff
@@ -140,15 +141,15 @@ func getRole(key []byte, roles []*coredocumentpb.Role) (*coredocumentpb.Role, er
 	return nil, errors.New("role %d not found", key)
 }
 
-// isAccountInRole returns true if account is in the given role as collaborators.
-func isAccountInRole(role *coredocumentpb.Role, account identity.CentID) bool {
-	for _, id := range role.Collaborators {
+// IsAccountInRole returns true if account is in the given role as collaborators.
+func IsAccountInRole(role *coredocumentpb.Role, account identity.CentID) (int, bool) {
```
mikiquantum: Wouldn't be more succinct if the return was an `int` only? Return index of role in collabs or -1 if not present. Renaming to `FindAccountInRole`.

coredocument/read_acls.go
```diff
@@ -228,18 +230,18 @@ func findRole(
 	return false
 }
 
-// isNFTInRole checks if the given nft(registry + token) is part of the core document role.
-func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []byte) bool {
+// IsNFTInRole checks if the given nft(registry + token) is part of the core document role.
+func IsNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []byte) (int, bool) {
```
mikiquantum: Same here?

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,154 @@ func newEthereumPaymentObligation(
...
-		return MintRequest{}, err
+		return mreq, err
 	}
 
 	return requestData, nil
 
 }
 
+// getNFTProofs generates required NFT uniqueness proofs
+func getNFTProofs(cd *coredocumentpb.CoreDocument, req MintNFTRequest, tokenID TokenID, cid identity.CentID) ([]proofspb.Proof, error) {
```
mikiquantum: Maybe instead: `generateNFTProofs(...)` ?

mikiquantum: In addition, I believe coredoc manipulation will only be done in the near future from the codedoc model layer. We should think on how this would look like by then. For the purpose of this PR it is ok to leave it but let's sync with Charly on it.

coredocument/read_acls.go
vedhavyas: I was think of a pattern that would look like 

`if _, ok := FindAccountInRole; ok {...}` since the index is ignored in most usages except this one.

nft/ethereum_payment_obligation.go
vedhavyas: Yeah, when the coredoc is ready we will move these over there. Until then, it should be okay here IMO

coredocument/read_acls.go
vedhavyas: This was one pattern I saw mostly used since checking index for -1 might not be  intuitive IMO

mikiquantum: Sounds good!

mikiquantum: Oh, adapt the comment as well.

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
+
+	var pfs []*proofspb.Proof
+	pfs = append(pfs, docProofs.FieldProofs...)
+	for _, p := range nftProofs {
+		p := p
+		pfs = append(pfs, &p)
 	}
 
-	toAddress := common.HexToAddress(depositAddress)
+	toAddress := common.HexToAddress(req.DepositAddress)
```
lucasvo: Wouldn't be nice not to "rename" the address here and call it `depositAddress`?
```suggestion
	depositAddress := common.HexToAddress(req.DepositAddress)
```

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
 	if err != nil {
-		return MintRequest{}, err
+		return mreq, err
 	}
 
 	return requestData, nil
 
 }
 
+// generateNFTProofs generates required NFT uniqueness proofs
```
lucasvo: This method doesn't just do the uniqueness.

```suggestion
// generateNFTProofs generates required NFT proofs
```

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
+
+	if req.SubmitRoleProof != "" {
+		pf, err := createRoleProof(cd, cdTree, req.SubmitRoleProof, cid)
+		if err != nil {
+			return nil, err
+		}
+
+		pfs = append(pfs, pf)
+	}
+
```
lucasvo: I think a proof is missing here for the field `NextVersion`. We should probably also add an option to the MintRequest to only add it if required. I think @rdinicut needs it to verify the document is the newest version before minting the NFT.

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
+		return proof, ErrNFTRoleMissing
+	}
+
+	pk := fmt.Sprintf("roles[%s].collaborators[%d]", hexutil.Encode(role.RoleKey), idx)
+	return cdTree.CreateProof(pk)
+}
+
+// createNFTReadAccessProof creates proof that nft exists in the role
+func createNFTReadAccessProof(cd *coredocumentpb.CoreDocument, cdTree *proofs.DocumentTree, registry common.Address, tokenID []byte) (proofs []proofspb.Proof, err error) {
+
```
lucasvo: extra space

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
+		return found
+	})
+
+	if !found {
+		return nil, ErrNFTRoleMissing
+	}
+
+	fields := []string{
+		fmt.Sprintf("read_rules[%d].roles[%d]", rridx, ridx),          // proof that a read rule exists with the nft role
```
lucasvo: rridx & ridx also need to be hex encoded AFAIK

nft/ethereum_payment_obligation.go
```diff
@@ -80,45 +88,155 @@ func newEthereumPaymentObligation(
...
+		proofs = append(proofs, pf)
+	}
+
+	return proofs, nil
+}
+
+func createTokenProof(cd *coredocumentpb.CoreDocument, cdTree *proofs.DocumentTree, registry common.Address) (proof proofspb.Proof, err error) {
+	nft := getStoredNFT(cd.Nfts, registry.Bytes())
+	if nft == nil {
+		return proof, errors.New("nft is missing from the p2p document")
```
lucasvo:
```suggestion
		return proof, errors.New("nft is missing from the document")
```

nft/ethereum_payment_obligation.go
```diff
@@ -145,7 +263,21 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
 		return nil, nil, err
 	}
 
-	registry := common.HexToAddress(registryAddress)
+	// if the role_proof is given check if the role exists with account id as collaborator
+	// at the moment, since we generate role key randomly, this will not work
```
lucasvo: I would remove this comment. It will be outdated pretty soon. And the proof itself still works, it's just not the most helpful :)

nft/ethereum_payment_obligation.go
```diff
@@ -145,7 +263,21 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, documentID []by
...
+	// but once we figure out acls on the client API, this will continue working with out any chnages
+	if request.SubmitRoleProof != "" {
+		_, role := getRoleForAccount(cd, request.SubmitRoleProof, cid)
+		if role == nil {
+			return nil, nil, ErrNFTRoleMissing
+		}
+	}
+
+	if !request.GrantNFTReadAccess && request.SubmitNFTReadAccessProof {
+		return nil, nil, errors.New("enabled grant nft access to generate NFT read access proof")
```
lucasvo: This error message doesn't really say what the problem is, I think it's not clear enough.

lucasvo: Also, wouldn't it be better to do this rather basic check further up at the beginning before doing a bunch of other validation?

nft/ethereum_payment_obligation.go
```diff
@@ -275,6 +416,23 @@ func getStoredNFT(nfts []*coredocumentpb.NFT, registry []byte) *coredocumentpb.N
 	return nil
 }
 
+// getRoleForAccount returns the matching role and index of the account
+func getRoleForAccount(cd *coredocumentpb.CoreDocument, roleName string, id identity.CentID) (idx int, role *coredocumentpb.Role) {
+	sha := sha256.New()
```
lucasvo: We shouldn't default to sha256 here but require the client to specify the hash of the role, not the name. I don't think it should be a requirement that you use a specific hash function for roles.

nft/payment_obligation.go
```diff
@@ -54,10 +54,22 @@ func (t TokenID) String() string {
...
 
+// MintNFTRequest holds required fields for minting NFT
+type MintNFTRequest struct {
+	DocumentID               []byte
+	RegistryAddress          string
+	DepositAddress           string
+	ProofFields              []string
+	GrantNFTReadAccess       bool
+	SubmitTokenProof         bool
+	SubmitRoleProof          string
```
lucasvo: I would rename this to make it clear that this is the actual role. And as mentioned above, I'd make it a []byte
```suggestion
	AccountRoleID          []byte
```

coredocument/read_acls.go
```diff
@@ -125,12 +124,14 @@ type readAccessValidator struct {
 // Returns an error if not.
 func (r readAccessValidator) AccountCanRead(cd *coredocumentpb.CoreDocument, account identity.CentID) bool {
 	// loop though read rules
-	return findRole(cd, coredocumentpb.Action_ACTION_READ_SIGN, func(role *coredocumentpb.Role) bool {
-		return isAccountInRole(role, account)
+	return FindRole(cd, coredocumentpb.Action_ACTION_READ_SIGN, func(_, _ int, role *coredocumentpb.Role) bool {
```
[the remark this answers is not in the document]
vedhavyas: Yes, NFTs will have only READ action atm. Hence, we are filtering based on that

nft/ethereum_payment_obligation.go
vedhavyas: It should be part of the default proofs fields that get sent by the client in the request. My understanding is that it was the choice we made instead of adding specific proofs by default in the Mint method

nft/ethereum_payment_obligation.go
vedhavyas: Nope, else tests would have failed ;)

nft/payment_obligation.go
vedhavyas: This is client facing API request, we cannot have []byte. We will take it as a hex if this user gives the 32 byte role label

nft/ethereum_payment_obligation.go
vedhavyas: I have used sha256 because we haven't defined how the role label is filled out ATM. Would users will give 32 byte key for role? My understanding from the discussion with @pstehlik is that we might take a possible name like "supplier" as the role label. In that case, we would need to covert it to a 32 byte word

nft/ethereum_payment_obligation.go
vedhavyas: This is the second validation check in the request. This is in the top-level validation

coredocument/read_acls.go
```diff
@@ -140,15 +141,15 @@ func getRole(key []byte, roles []*coredocumentpb.Role) (*coredocumentpb.Role, er
...
+func IsAccountInRole(role *coredocumentpb.Role, account identity.CentID) (int, bool) {
+	for i, id := range role.Collaborators {
 		if bytes.Equal(id, account[:]) {
-			return true
+			return i, true
 		}
 	}
 
-	return false
+	return 0, false
```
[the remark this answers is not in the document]
vedhavyas: According to the Golang standard, we will send the default value along with false bool, The function should always check for true before using the index just like how we check err value before using the remaining return arguments. If the function caller uses the index when the role is not found, that would be clearly wrong

coredocument/read_acls.go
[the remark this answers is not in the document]
lucasvo: The method is called ACTION_READ_SIGN but the method is called AccountCanRead. That's what I'm struggling with.

coredocument/read_acls.go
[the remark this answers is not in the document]
lucasvo: Ok, I was used to the python way to do this. But if this is go standard, then let's do it this way.

nft/ethereum_payment_obligation.go
lucasvo: Ok, you are right we can just supply it as a field.

nft/ethereum_payment_obligation.go
lucasvo: hm, ok that's odd. We should definitely investigate that there are no bugs hidden somewhere there. I'll look into this.

nft/ethereum_payment_obligation.go
lucasvo: Read the error message again. I don' think it makes sense. Shouldn't it be: `disabled GrantNFTReadAccess...`

nft/ethereum_payment_obligation.go
lucasvo: The idea is that this is a convention we suggest people to follow but it isn't hard coded. We can't reverse this hash so when we would return the roles to the user we couldn't show the preimage of the hash which leads to odd behavior anyway. 

The idea is that in the future we can allow users to setup "labels" for these identifiers but this would work on a higher level in the node.

nft/payment_obligation.go
lucasvo: Ah didn't catch that this is the client model. Somewhere else I spotted string as well.

coredocument/read_acls.go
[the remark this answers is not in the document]
vedhavyas: The action is ACTION_READ_SIGN fro all accounts at the moment. So the method Account can read is actually filtering the right action.

documents/invoice/model.go
```diff
@@ -264,16 +262,16 @@ func (i *Invoice) getInvoiceSalts(invoiceData *invoicepb.InvoiceData) (*proofs.S
 // ID returns document identifier.
 // Note: this is not a unique identifier for each version of the document.
 func (i *Invoice) ID() ([]byte, error) {
-	coreDoc, err := i.PackCoreDocument()
+	coreDocModel, err := i.PackCoreDocument()
```
lucasvo: Why do we need to do this here? Can't we expose a getter on the CoreDocumentModel to get the identifier? This seems a fairly standard thing.

documents/invoice/model.go
```diff
@@ -376,19 +378,19 @@ func (i *Invoice) getDocumentDataTree() (tree *proofs.DocumentTree, err error) {
 }
 
 // CreateProofs generates proofs for given fields
-func (i *Invoice) CreateProofs(fields []string) (coreDoc *coredocumentpb.CoreDocument, proofs []*proofspb.Proof, err error) {
+func (i *Invoice) CreateProofs(fields []string) (proofs []*proofspb.Proof, err error) {
 	// There can be failure scenarios where the core doc for the particular document
 	// is still not saved with roots in db due to failures during getting signatures.
-	coreDoc, err = i.PackCoreDocument()
+	_, err = i.PackCoreDocument()
```
lucasvo: Why do we call pack here? Isn't that abusing the pack function for some sort of validation?

documents/invoice/service.go
```diff
@@ -107,7 +107,7 @@ func (s service) validateAndPersist(ctx context.Context, old, new documents.Mode
 	}
 
 	// we use CurrentVersion as the id since that will be unique across multiple versions of the same document
-	err = s.repo.Create(self.ID[:], inv.CoreDocument.CurrentVersion, inv)
+	err = s.repo.Create(self.ID[:], inv.CoreDocumentModel.Document.CurrentVersion, inv)
```
lucasvo: See my above comment, where you call pack to get the DocumentIdentifier. This doesn't seem necessary?

documents/invoice/service.go
```diff
@@ -127,13 +127,13 @@ func (s service) Create(ctx context.Context, inv documents.Model) (documents.Mod
...
 
-	cd, err := inv.PackCoreDocument()
+	dm, err := inv.PackCoreDocument()
 	if err != nil {
 		return nil, uuid.Nil, nil, err
 	}
 
 	txID := contextutil.TX(ctx)
-	txID, done, err := documents.CreateAnchorTransaction(s.txManager, s.queueSrv, self.ID, txID, cd.CurrentVersion)
+	txID, done, err := documents.CreateAnchorTransaction(s.txManager, s.queueSrv, self.ID, txID, dm.Document.CurrentVersion)
```
lucasvo: I don't get when you're using pack and when you don't :D

documents/model.go
```diff
@@ -97,6 +111,7 @@ func NewLeafProperty(literal string, compact []byte) proofs.Property {
 
 // NewCoreDocModel returns a new CoreDocumentModel
 // Note: collaborators and salts are to be filled by the caller
+// TODO: double check if registry should be initialised as nil
```
lucasvo: What will you do about this?

documents/model.go
```diff
@@ -740,17 +749,197 @@ func (m *CoreDocumentModel) ValidateDocumentAccess(docReq *p2ppb.GetDocumentRequ
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return ndm, ndm.setCoreDocumentSalts()
+}
+
+// IsNFTMinted checks if the there is an NFT that is minted against this document in the given registry.
+func (m *CoreDocumentModel) IsNFTMinted(tokenRegistry TokenRegistry, registry common.Address) bool {
```
vimukthi-git: I think this function must not exist in a model
A model should not have any runtime dependancies, it makes them hard to test. We can make this part of some nft utility and call it when needed from outside the model.

documents/model.go
```diff
@@ -740,17 +749,197 @@ func (m *CoreDocumentModel) ValidateDocumentAccess(docReq *p2ppb.GetDocumentRequ
...
+	if err != nil {
+		return nil, errors.New("failed to generate signing root proofs: %v", err)
+	}
+
+	cdtree, err := m.GetDocumentTree()
+	if err != nil {
+		return nil, errors.New("failed to generate document tree: %v", err)
+	}
+
+	// We support fields that belong to different document trees, as we do not prepend a tree prefix to the field, the approach
```
vimukthi-git: This comment is not relevant, since we only create proofs for `cdtree`

documents/model.go
```diff
@@ -740,17 +749,197 @@ func (m *CoreDocumentModel) ValidateDocumentAccess(docReq *p2ppb.GetDocumentRequ
...
+		proof.SortedHashes = append(proof.SortedHashes, dataRoot)
+		proof.SortedHashes = append(proof.SortedHashes, signingRootProofHashes...)
+		proofs = append(proofs, &proof)
+	}
+
+	return proofs, nil
+}
+
+func getReadAccessProofKeys(m *CoreDocumentModel, registry common.Address, tokenID []byte) (pks []string, err error) {
+	var rridx int  // index of the read rules which contain the role
```
vimukthi-git: Abbreviations need to be capitalised right? eg: rridx -> RRIDX

nft/payment_obligation_integration_test.go
```diff
@@ -37,72 +49,169 @@ func TestMain(m *testing.M) {
...
+	assert.True(t, errors.IsOfType(nft.ErrNFTMinted, err))
+}
+
+func failMintNFT(t *testing.T, grantNFT, nftReadAccess bool, roleProof []byte) {
+	ctx, id, registry, depositAddr, _, _ := prepareForNFTMinting(t)
+	req := nft.MintNFTRequest{
+		DocumentID:               id,
+		RegistryAddress:          registry,
+		DepositAddress:           common.HexToAddress(depositAddr),
+		ProofFields:              []string{"invoice.gross_amount", "invoice.currency", "invoice.due_date", "collaborators[0]"},
```
lucasvo: Do we still need to use `collaborators[0]` here?

documents/model.go
vedhavyas: not when they are starting with abbrevations like `id`

documents/model.go
vedhavyas: Ahh. good catch

documents/model.go
vedhavyas: I agree with you but checking the NFT is minted requires us to expose the `coredocumentpb.CoreDocument` which is what we tried avoiding with coredocmodel refactor.

documents/model.go
```diff
@@ -433,9 +441,10 @@ func (m *CoreDocumentModel) CalculateSigningRoot(dataRoot []byte) error {
 // AccountCanRead validate if the core document can be read by the account .
 // Returns an error if not.
 func (m *CoreDocumentModel) AccountCanRead(account identity.CentID) bool {
-	// loop though read rules
-	return m.findRole(coredocumentpb.Action_ACTION_READ_SIGN, func(role *coredocumentpb.Role) bool {
-		return isAccountInRole(role, account)
+	// loop though read rules, check all the rules
+	return m.findRole(nil, func(_, _ int, role *coredocumentpb.Role) bool {
+		_, found := isAccountInRole(role, account)
```
lucasvo: Wouldn't this change lead to any extra read rule to also match? Shouldn't we explicitly check that it's read or read_sign?

documents/model.go
vedhavyas: Example:
```
tokenID
id
idSomething
```

vimukthi-git: I don't get it, in this case the entire variable name is an abbreviation right? not just the start.

documents/model.go
lucasvo: Wouldn't we just mock the tokenRegistry here? I think this is fine the way you did it Ved.

documents/model.go
lucasvo: I agree with ved, internal abbreviated short variables like this are better than rRIDX or something similar.

documents/model.go
vimukthi-git: @lucasvo The function looks like it was hacked to make it testable, its a code smell for me :) But if it's hard to change its fine. Alternative is to define a method to get the NFT ID check externally.

documents/model.go
vimukthi-git: @lucasvo it wont be like that, according to go convention the variable name should be RRIDX

documents/model.go
vimukthi-git: I know, this is something we discussed already in the discussion for coredoc refactor as well, a model having

documents/model.go
vedhavyas: yeap. Agreed and changed

nft/payment_obligation_integration_test.go
vedhavyas: removed it. Part of the test we are skipping

documents/model.go
vedhavyas: @vimukthi-git it should be small abbreviation if the variable name contains only that example: `id`.
This would change if the variable name contains more than an abbreviation. Example "tokenID". 
If the variable name starts with the abbreviation it would be like this `xmlHTTPRequest`.

More here - https://github.com/golang/go/wiki/CodeReviewComments#initialisms

documents/model.go
```diff
@@ -128,6 +130,17 @@ func NewCoreDocModel() *CoreDocumentModel {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return ndm, ndm.setCoreDocumentSalts()
+}
+
+// prepareNewVersion prepares the next version of the CoreDocument
+// Note: salts needs to be filled by the caller
+func (m *CoreDocumentModel) prepareNewVersion(collaborators []string) (*CoreDocumentModel, error) {
```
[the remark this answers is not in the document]
vedhavyas: yeap, it will add the new collaborators to the coredoc

PR #739, opened 2019-02-10
Using almost latest precise proofs

mikiquantum:
Issue Link: <Github link>


coredocument/coredocument.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
```
mikiquantum: Disregard this file, is going away soon.

coredocument/coredocument_test.go
```diff
@@ -15,7 +15,6 @@ import (
 	"github.com/centrifuge/go-centrifuge/contextutil"
```
mikiquantum: Same with this file.

coredocument/coredocument.go
```diff
@@ -304,15 +306,54 @@ func GetExternalCollaborators(selfCentID identity.CentID, doc *coredocumentpb.Co
 	return collabs, nil
 }
 
+// GenerateNewSalts generates salts for new document
+func GenerateNewSalts(document proto.Message, prefix string) (*proofs.Salts, error) {
```
lucasvo: I know you're saying that we don't need to look at this file but maybe we can actually skip the GenerateNewSalts method altogether. The salts are modified when generating the tree so this is not necessary.

mikiquantum: I believe then you refer to the other `GenerateNewSalts` in the below model file. The main reason to keep a method like this, is to have explicitness on when the salts are generated and stored in the models. Otherwise I am not sure where to put that call.

mikiquantum:
@lucasvo Following our conversation offline, I will merge up now and discuss cleaning up salt generation later.

PR #741, opened 2019-02-11
identity: createConfig

manuelpolzhofer:
- create config with new identity


cmd/common.go
```diff
@@ -58,6 +68,110 @@ func addKeys(config config.Configuration, idService identity.Service) error {
 	return nil
 }
 
+func getKeyPairsFromConfig(config config.Configuration) (map[int]did.Key, error) {
```
mikiquantum: I believe this function should be more in the identity package as it will be reused from the API as well?

cmd/common.go
```diff
@@ -58,6 +68,110 @@ func addKeys(config config.Configuration, idService identity.Service) error {
...
+		return nil, err
+	}
+
+	address32Bytes := utils.AddressTo32Bytes(common.HexToAddress(secp256k1.GetAddress(pk)))
+	keys[identity.KeyPurposeEthMsgAuth] = did.NewKey(address32Bytes, big.NewInt(identity.KeyPurposeEthMsgAuth), big.NewInt(did.KeyTypeECDSA))
+
+	return keys, nil
+}
+
+func addKeysFromConfig(ctx map[string]interface{}, cfg config.Configuration) error {
```
mikiquantum: Same here

cmd/common.go
```diff
@@ -58,6 +68,110 @@ func addKeys(config config.Configuration, idService identity.Service) error {
...
+	}
+
+	err = idSrv.AddKey(tctx, keys[identity.KeyPurposeEthMsgAuth])
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func createIdentity(ctx map[string]interface{}, cfg config.Configuration, configFile *viper.Viper) error {
```
mikiquantum: Same here

identity/did/bootstrapper.go
```diff
@@ -80,7 +80,7 @@ func getAnchorAddress() common.Address {
 // ---------------------------------------------------------------------------------------------------------------------
 func migrateNewIdentityContracts() {
```
mikiquantum: This logic in this file is only for testing, we should think about removing it to the test bootstapper instead.

manuelpolzhofer:
ah okay. for the codecov the test needs to be in the same package

PR #743, opened 2019-02-12
Compact Precise Proofs

mikiquantum:
Issue Link: <Github link>


coredocument/coredocument.go
```diff
@@ -16,6 +16,41 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+const (
```
mikiquantum: Disregard this file.

coredocument/coredocument_test.go
```diff
@@ -215,10 +215,11 @@ func TestNewWithCollaborators(t *testing.T) {
```
mikiquantum: Disregard this file.

documents/model.go
```diff
@@ -18,6 +18,22 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+const (
```
mikiquantum: As sometimes we create the tree manually, we have to provide a deterministic compact name to each of the fields. Some of them might have a correspondent field in the coredocument and some dont. This keeps a map of the literal name and the compact name.

documents/model.go
```diff
@@ -53,6 +69,27 @@ const (
 	ErrZeroCollaborators = errors.Error("require at least one collaborator")
 )
 
+// NewDefaultTree returns a DocumentTree with default opts
+func NewDefaultTree(salts *proofs.Salts) *proofs.DocumentTree {
```
mikiquantum: I wanted to enforce a single way of creating a tree with the supported defaults. Here is the methods we should be using.

documents/model.go
```diff
@@ -53,6 +69,27 @@ const (
...
+	if prefix != "" {
+		prop = proofs.NewProperty(prefix)
+	}
+
+	t := proofs.NewDocumentTree(proofs.TreeOptions{CompactProperties: true, EnableHashSorting: true, Hash: sha256.New(), ParentPrefix: prop, Salts: salts})
+	return &t
+}
+
+// NewLeafProperty returns a proof property with the literal and the compact
+func NewLeafProperty(literal string, compact []byte) proofs.Property {
```
mikiquantum: Same for creating custom properties, I wanted to enforce that we always pass the compact name as well

documents/model_test.go
```diff
@@ -287,18 +287,20 @@ func TestGetDocumentRootTree(t *testing.T) {
 	tree, err := dm.GetDocumentRootTree()
 
 	// Manually constructing the two node tree:
-	signaturesLengthLeaf := sha256.Sum256(append([]byte("signatures.length0"), make([]byte, 32)...))
-	expectedRootHash := sha256.Sum256(append(signaturesLengthLeaf[:], dm.Document.SigningRoot...))
+	signaturesLengthLeaf := sha256.Sum256(append(append(compactProperties[SignaturesField], []byte{48}...), make([]byte, 32)...))
```
mikiquantum: Looks like with the new precise proof update the order of hashing leaves has changed.

documents/model.go
lucasvo: Not an immediate action item, but we have to make sure that we don't accidentally create two fields with the same property (one in invoice and one in CoreDoc for example). 

We could achieve this using reserved statements and putting field numbers into there for the fields of the other documents.

```
message Foo {
  reserved 2, 15, 9 to 11;
  reserved "foo", "bar";
}
```

coredocument/coredocument.go
```diff
@@ -16,6 +16,41 @@ import (
...
+	SignaturesField:   {0, 0, 0, 6},
+	SigningRootField:  {0, 0, 0, 10},
+}
+
+func NewDefaultTree(salts *proofs.Salts) *proofs.DocumentTree {
+	return NewDefaultTreeWithPrefix(salts, "")
+}
+
+func NewDefaultTreeWithPrefix(salts *proofs.Salts, prefix string) *proofs.DocumentTree {
+	var prop proofs.Property
```
lucasvo: Technically we also have to prefix the compact version of the properties so we'll need to have a byte and string prefix. I think it's fine to implement this in a later PR but it's something we should think about. Perhaps we actually can pass in a Property as a prefix instead.

documents/model_test.go
lucasvo: But why did you say the order has changed?

lucasvo: It's property, value, salt. Was it different before?

Actually []byte{48} -> seems to be the wrong length value here, that to me looks like an ASCII encoded "SPACE" character and not a 0. Though this will be fixed when changing the way we generate the signatures. So not worth fixing I think.

documents/model.go
mikiquantum: Yes, agree the initial definition of the protocol documents should have that defined upfront. In this particular case, for low level general fields, I would argue to have those fields defined in the coredoc proto message, even though they are calculated every time.

documents/model_test.go
mikiquantum: within leaves, not within a leaf. Not sure why but if you look at the next line 291 I had to change the order. 
Regarding the ascii code, int(48) is 0 in ASCII, as the node logic for length fields does: `[]byte("0")`. I could transform that to encode an int in bytes today.

coredocument/coredocument.go
mikiquantum: True, I will add another entry in the map for our prefixes used as a "reserved" prefix.

mikiquantum: The lengthProp method needs to add as well the compact property for `length`: https://github.com/centrifuge/precise-proofs/blob/13d3af957299c614237c42cdc331f7acd7c7d201/proofs/property.go#L127-L133

documents/model_test.go
mikiquantum: Actually this should be addressed once we create the signatures as a tree soon.

PR #746, opened 2019-02-13
identity: additional service methods

manuelpolzhofer:
added new methods to service
- GetClientP2PURL(did DID) (string, error)
- Exists(ctx context.Context, did DID) error
- ValidateKey(ctx context.Context, did DID, key []byte, purpose int64) error
- GetClientsP2PURLs(did []*DID) ([]string, error)
- GetKeysByPurpose(did DID, purpose *big.Int) ([][32]byte, error)

refactored CALL methods
- DID passed instead of context


cmd/common_test.go
```diff
@@ -69,20 +68,21 @@ func TestCreateConfig(t *testing.T) {
 
 	// contract exists
 	id, err := cfg.GetIdentityID()
+	centID := did.NewDID(common.BytesToAddress(id))
```
vedhavyas: call this `account` maybe ?

PR #748, opened 2019-02-14
identity migration: update to latest contracts

manuelpolzhofer:
- removed second contract repo
- removed registry from stored config

migrated packages
- did
- cmd

### DIFF
--- a/anchors/bootstrapper.go
+++ b/anchors/bootstrapper.go
@@ -1,6 +1,7 @@
 package anchors
 
 import (
+	"fmt"
 	"github.com/centrifuge/go-centrifuge/config"
 	"github.com/centrifuge/go-centrifuge/config/configstore"
 	"github.com/centrifuge/go-centrifuge/errors"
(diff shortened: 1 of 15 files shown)


anchors/bootstrapper.go
```diff
@@ -29,6 +30,9 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	}
 	client := ctx[ethereum.BootstrappedEthereumClient].(ethereum.Client)
 
+	fmt.Println("anchors")
```
mikiquantum: I know this is testing material, let's try to keep as "production" as possible, bc otherwise when we merge to develop it will be harder to catch.

build/scripts/tests/run_integration_tests.sh
```diff
@@ -20,6 +20,7 @@ for d in $(go list -tags=integration ./... | grep -v vendor); do
         cat profile.out >> coverage.txt
         rm profile.out
     fi
+    exit $status
```
mikiquantum: Do we need this here?

identity/did/bootstrapper.go
```diff
@@ -66,99 +71,32 @@ func bindFactory(factoryAddress common.Address, client ethereum.Client) (*Factor
...
-	smartContractAddresses = GetSmartContractAddresses()
-
-}
-
-// RunNewSmartContractMigrations migrates smart contracts to localgeth
-// TODO: func will be removed after migration
-func runNewSmartContractMigrations() {
-
+// TODO: store identity bytecode in config and remove func
+func getIdentityByteCode() string {
```
mikiquantum: Where are we using this?

manuelpolzhofer: We are using it for CreateIdentity to check if it was successful

anchors/bootstrapper.go
manuelpolzhofer: fixed in next PR

build/scripts/tests/run_integration_tests.sh
manuelpolzhofer: fixed in next PR

PR #752, opened 2019-02-15
Feat/acls access token read acls

charleenfei:
Issue Link: #678 

Covers Access Token aspects of: 

- issuance
- adding AT to the Read Rules 
- validation of Document Access with AT


documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+	tm = append(tm, grantee...)
+	tm = append(tm, roleID)
+	tm = append(tm, docID...)
+
+	// fetch key pair from identity
+	privateKey := id.Keys[identity.KeyPurposeEthMsgAuth].PrivateKey
+	pubKey := id.Keys[identity.KeyPurposeEthMsgAuth].PublicKey
+
+	// sign the assembled access token message, return signature
+	sig, err := crypto.SignMessage(privateKey, tm, "CurveSecp256K1", true)
```
lucasvo: it would be nice to expose the curves as constants in the crypto package and then refer to it instead of adding a string here, @vimukthi-git you initially refactored the crypto package. What do you think?

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+	if err != nil {
+		return nil, err
+	}
+	// assemble the access token, appending the signature and public keys
+	at := new(coredocumentpb.AccessToken)
```
lucasvo: Why not use this construction:
```
at := coredocumentpb.AccessToken{
    Identifier: tokenIdentifier,
    Granter: granterId,
    Grantee: grantee,
    ...
}
```

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
 
 	return false
 }
+
+// assembleAccessToken assembles a Read Access Token from the payload received
+func assembleAccessToken(dm *CoreDocumentModel, id identity.IDConfig, payload documentpb.AccessTokenParams) (*coredocumentpb.AccessToken, error) {
+	tokenIdentifier := utils.RandomSlice(32)
+	granterID := id.ID[:]
+	roleID := byte(len(dm.Document.Roles))
+	grantee, err := hexutil.Decode(payload.GetGrantee())
```
lucasvo:
```suggestion
	granteeID, err := hexutil.Decode(payload.GetGrantee())
```

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+		return errors.New("invalid access token granter")
+	}
+	// TODO:  check that the granter is the owner of the public key on the access token
+	
+	// check that the account requesting access is the grantee of the access token
+	if !bytes.Equal(account[:], token.Grantee) {
+		return errors.New("access token grantee is not the same as the account requesting access")
+	}
+	// validate the access token
+	err = validateAT(token.Key, token)
```
lucasvo: I think this should be done first, as this is the "cheapest" check and invalidates everything else.

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+	}
+
+	return nil
+}
+
+// validateAT validates that given access token against its signature
+// ethereum specific variation is currently false
+
+func validateAT (publicKey []byte, token *coredocumentpb.AccessToken) error {
+	tm := append(token.Identifier, token.Granter...)
```
lucasvo: I think it would be a good idea to create a method that generates the signaturePayload and use that message both when creating the token and when validating the token instead of duplicating it here.

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+
+		for _, rk := range rule.Roles {
+			role, err := getRole(rk, cd.Roles)
+			if err != nil {
+				// seems like roles and rules are not in sync
+				// skip to next one
+				continue
+			}
+
+			at, err := onRole(role, tokenID)
```
lucasvo: Instead of doing this call here, I think it's better to cache the roles that the tokenID is in before iterating over all the rules and then simply checking if it's in the list. 

```go
// Line 885
var tokenRoles := findRolesForAT(roles *coredocumentpb.Roles, tokenID []byte) 

// Line 899
for _, r := range(tokenRoles) {
    if r == role {
        // match
    }
}
```

documents/model.go
vimukthi-git: This was on my mind as well, it makes sense to add constants.

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+}
+
+// AddAccessTokenToReadRules adds the AccessToken(s) to the read rules of the document
+func (m *CoreDocumentModel) AddAccessTokenToReadRules(id identity.IDConfig, payload documentpb.AccessTokenParams) error {
+	at, err := assembleAccessToken(m, id, payload)
+	if err != nil {
+		return errors.New("failed to construct AT: %v", err)
+	}
+
+	role := new(coredocumentpb.Role)
```
vimukthi-git: Please create a constructor with parameters for the role

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+	}
+	// check that the token granter has the correct access type for the document
+	validGranter := m.AccountCanRead(granterID)
+	if !validGranter {
+		return errors.New("invalid access token granter")
+	}
+	// TODO:  check that the granter is the owner of the public key on the access token
+	
+	// check that the account requesting access is the grantee of the access token
+	if !bytes.Equal(account[:], token.Grantee) {
```
vimukthi-git: CentID has an equals method I think, if not you can create

documents/model.go
```diff
@@ -754,3 +755,154 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+		return err
+	}
+
+	granterID, err := identity.ToCentID(token.Granter)
+	if err != nil {
+		return err
+	}
+	// check that the token granter has the correct access type for the document
+	validGranter := m.AccountCanRead(granterID)
+	if !validGranter {
```
vimukthi-git: Remove `validGranter := m.AccountCanRead(granterID)` and do
```suggestion
	if m.AccountCanRead(granterID) {
```

documents/model.go
lucasvo: Actually they are already public, you can use: crypto.CurveSecp256K1
```suggestion
	sig, err := crypto.SignMessage(privateKey, tm, crypto.CurveSecp256K1, true)
```

documents/model.go
charleenfei: why is this preferable than !validGranter?

documents/model.go
```diff
@@ -729,10 +731,12 @@ func (m *CoreDocumentModel) ValidateDocumentAccess(docReq *p2ppb.GetDocumentRequ
...
 		}
-		//// case AccessTokenValidation
-		// case p2ppb.AccessType_ACCESS_TYPE_ACCESS_TOKEN_VERIFICATION:
-		//
-		// case p2ppb.AccessType_ACCESS_TYPE_INVALID:
+	case p2ppb.AccessType_ACCESS_TYPE_ACCESS_TOKEN_VERIFICATION:
+		if m.ATOwnerCanRead(docReq.AccessTokenRequest, requesterCentID) != nil {
+			return errors.New("requester does not have access")
+		}
+	case p2ppb.AccessType_ACCESS_TYPE_INVALID:
```
vedhavyas: I think we can skip this case and use default here ?

documents/model.go
```diff
@@ -729,10 +731,12 @@ func (m *CoreDocumentModel) ValidateDocumentAccess(docReq *p2ppb.GetDocumentRequ
 		if m.NFTOwnerCanRead(registry, docReq.NftTokenId, requesterCentID) != nil {
 			return errors.New("requester does not have access")
 		}
-		//// case AccessTokenValidation
-		// case p2ppb.AccessType_ACCESS_TYPE_ACCESS_TOKEN_VERIFICATION:
-		//
-		// case p2ppb.AccessType_ACCESS_TYPE_INVALID:
+	case p2ppb.AccessType_ACCESS_TYPE_ACCESS_TOKEN_VERIFICATION:
+		if m.ATOwnerCanRead(docReq.AccessTokenRequest, requesterCentID) != nil {
```
vedhavyas: rename `requesterCentID` to `peer`?

documents/model.go
vimukthi-git: just idiomatic go way of not creating unnecessary variables.

documents/model.go
vimukthi-git: lets define constructor functions, I'm sure they would be useful to have

documents/model.go
```diff
@@ -754,3 +758,153 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+	})
+	if err != nil {
+		return err
+	}
+
+	granterID, err := identity.ToCentID(token.Granter)
+	if err != nil {
+		return err
+	}
+	// check that the token granter has the correct access type for the document
```
vedhavyas:
```suggestion
	// check that the token granter has the access to the document
```

documents/model.go
```diff
@@ -754,3 +758,153 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
 
 	return false
 }
+
+// assembleAccessToken assembles a Read Access Token from the payload received
+func assembleAccessToken(dm *CoreDocumentModel, id identity.IDConfig, payload documentpb.AccessTokenParams) (*coredocumentpb.AccessToken, error) {
+	tokenIdentifier := utils.RandomSlice(32)
+	granterID := id.ID[:]
+	roleID := byte(len(dm.Document.Roles))
```
vedhavyas: lets use the 32 byte role label here. We will remove the increment role id soon

documents/model.go
```diff
@@ -754,3 +758,153 @@ func isNFTInRole(role *coredocumentpb.Role, registry common.Address, tokenID []b
...
+}
+
+// AddAccessTokenToReadRules adds the AccessToken(s) to the read rules of the document
+func (m *CoreDocumentModel) AddAccessTokenToReadRules(id identity.IDConfig, payload documentpb.AccessTokenParams) error {
+	at, err := assembleAccessToken(m, id, payload)
+	if err != nil {
+		return errors.New("failed to construct AT: %v", err)
+	}
+
+	role := new(coredocumentpb.Role)
```
vedhavyas: Role should be created first and then use the created role label in the access token

documents/model.go
lucasvo: I would just use utils.RandomSlice(32) for this as well.

charleenfei:
> I would rename methods that has `AT` suffix, to just use `Token` instead since there is nothing referred to as a token here other than the access token.
> eg: ValidateAT => ValidateToken

we do have another token type (NFTs)

vimukthi-git:
@charleenfei `AT` is confusing as an abbreviation to me, may be subjective. Thats said lets see how it's being used, we can always rename stuff later.

documents/model.go
```diff
@@ -851,20 +862,34 @@ func (m *CoreDocumentModel) IsAccountInRole(roleKey []byte, account identity.DID
...
 
 // assembleTokenMessage assembles a token message
-func assembleTokenMessage(tokenIdentifier []byte, granterID []byte, granteeID []byte, roleID []byte, docID []byte) []byte {
+func assembleTokenMessage(tokenIdentifier []byte, granterID []byte, granteeID []byte, roleID []byte, docID []byte) ([]byte, error) {
+	tokenIdentifiers := [][]byte{tokenIdentifier, roleID, docID}
+	for _, id := range tokenIdentifiers {
+		if len(id) != IDByteCount {
+			return nil, errors.New("invalid identifier length")
+		}
+	}
```
lucasvo: You should also check granter and grantee for the right length :)

lucasvo: or maybe even better, use `identity.DID` type for it?

vimukthi-git: that definitely makes sense

charleenfei: yep, now that identity is merged in the bytelength is correct (20)

documents/model.go
```diff
@@ -32,6 +38,21 @@ const (
 	SignaturesField = "signatures"
 	// SigningRootField represents the signature root property of a tree
 	SigningRootField = "signing_root"
+
+	// ErrZeroCollaborators error when no collaborators are passed
+	ErrZeroCollaborators = errors.Error("require at least one collaborator")
+	// nftByteCount is the length of combined bytes of registry and tokenID
+	nftByteCount = 52
+	// ErrNFTRoleMissing errors when role to generate proof doesn't exist
+	ErrNFTRoleMissing = errors.Error("NFT Role doesn't exist")
```
vimukthi-git: please move error constants to `error.go`

documents/model.go
```diff
@@ -614,24 +625,25 @@ func getRole(key []byte, roles []*coredocumentpb.Role) (*coredocumentpb.Role, er
 	return nil, errors.New("role %d not found", key)
 }
 
+func newRole() *coredocumentpb.Role {
```
vimukthi-git: 👍

documents/model.go
vimukthi-git: 👍

Issue #692, opened 2019-01-24
Convert transaction IDs returned from API into hex and accept hex encoded txIDs when received


vimukthi-git:
Suggested Impl:

We can define a type like the following

```
type TxID *uuid.UUID

func (tid TxID) String() {
   // convert tid to hex string
}

func FromString(tidHex string) TxID {
// convert tid hex string to TxID
}

```

Then use the above type everywhere we use the current UUID based tx ID.

#### Testing
- Tests for above type
- Testworld should pass

#### Acceptance Criteria
- All Client APIs return hex converted TX IDs
- All Client APIs accept hex encoded TX IDs if they require.

cc @lucasvo @vedhavyas

lucasvo:
Does it make sense to actually derive it from UUID? Why not derive it from the type ethereum uses internally for IDs?

lucasvo:
Also I'd like to use completely random UUIDs and nothing that has a timestamp by default.

vedhavyas:
We are using uuid v4 which is completely random. Timestamp will not be part of it

But I don't understand the previous comment about deriving

PR #754, opened 2019-02-18
Converting TX API to use Hex strings

vimukthi-git:
Issue Link: #692


transactions/transaction.go
```diff
@@ -36,9 +51,53 @@ func NewLog(action, message string) Log {
 	}
 }
 
+// TxID is a centrifuge transaction ID. Internally represented by a UUID. Externally visible as a byte slice or a hex encoded string.
+type TxID uuid.UUID
```
vimukthi-git: main change

vimukthi-git:
@vedhavyas its needed to avoid cyclic dependancies, if you know of a way for it to work without, I would welcome that.

vimukthi-git:
@vedhavyas Thanks for the discussion, What I would take from it is that,
- We need to refactor the `configstore.RetriveConfig`(A bad design by me :P ) call so that it doesn't create dependancies everywhere. (Follow up pr)
- Since transaction package is needed everywhere(including `contextutil`) and transaction package may end up needing accounts and identity lets keep the new separate package structure inside transactions package. I find that the model(interface + sub package imple) that I found in my learnings to work seems to have been found by other people such as here https://go-talks.appspot.com/github.com/benbjohnson/structuring-applications-for-growth/main.slide#6 https://codeandtalk.com/v/gophercon-denver-2016/gophercon-2016-lightning-talk-ben-johnson

documents/purchaseorder/model_test.go
```diff
@@ -8,38 +8,32 @@ import (
...
-	"github.com/stretchr/testify/mock"
-
-	"github.com/centrifuge/go-centrifuge/storage/leveldb"
-
-	"github.com/centrifuge/go-centrifuge/testingutils/config"
-
-	"github.com/centrifuge/go-centrifuge/identity/ethid"
-
-	"github.com/centrifuge/go-centrifuge/config/configstore"
-
```
vedhavyas: Thank you for arranging the imports

vimukthi-git: you made me have OCD about these now :D

PR #758, opened 2019-02-19
Identity

mikiquantum:
Issue Link: <Github link>


transactions/manager.go
```diff
@@ -19,12 +19,12 @@ type Config interface {
...
 // Manager is a manager for centrifuge transactions.
 type Manager interface {
-	// ExecuteWithinTX executes a transaction within a transaction
-	ExecuteWithinTX(ctx context.Context, accountID identity.CentID, existingTxID uuid.UUID, desc string, work func(accountID identity.CentID, txID uuid.UUID, txMan Manager, err chan<- error)) (txID uuid.UUID, done chan bool, err error)
-	GetTransaction(accountID identity.CentID, id uuid.UUID) (*Transaction, error)
-	UpdateTaskStatus(accountID identity.CentID, id uuid.UUID, status Status, taskName, message string) error
-	GetTransactionStatus(accountID identity.CentID, id uuid.UUID) (*transactionspb.TransactionStatusResponse, error)
-	WaitForTransaction(accountID identity.CentID, txID uuid.UUID) error
+	// ExecuteWithinTX executes the given unit of work within a transaction
+	ExecuteWithinTX(ctx context.Context, accountID identity.DID, existingTxID uuid.UUID, desc string, work func(accountID identity.DID, txID uuid.UUID, txMan Manager, err chan<- error)) (txID uuid.UUID, done chan bool, err error)
```
manuelpolzhofer: 👍

anchors/anchor.go
```diff
@@ -133,7 +133,7 @@ func NewCommitData(blockHeight uint64, anchorID AnchorID, documentRoot DocumentR
 }
 
 // GenerateCommitHash generates Keccak256 message from AnchorID, CentID, DocumentRoot
-func GenerateCommitHash(anchorID AnchorID, centrifugeID identity.CentID, documentRoot DocumentRoot) []byte {
+func GenerateCommitHash(anchorID AnchorID, centrifugeID identity.DID, documentRoot DocumentRoot) []byte {
```
manuelpolzhofer: After the migration, we should think about if need the identity address as part of the anchor commit.
In case we using the execute method for anchoring later.

build/configs/default_config.yaml
```diff
@@ -31,6 +37,12 @@ networks:
       identityRegistry: "0x9660c039d311453af0d58c5666723d9c2fa7d6ec"
       anchorRepository: "0x7f854dfa98012d7fa55c803bba2260bcdee4b5ed"
       paymentObligation: "0xdb0581A9328664855328AdDb0E251184640f9e5D"
+    # Latest deployed Smart Contracts Byte code
+    contractByteCode:
```
manuelpolzhofer: 👍

manuelpolzhofer: maybe later we can replace it with a path to an extra file

build/configs/testing_config.yaml
```diff
@@ -6,7 +6,7 @@ ethereum:
       password: ''
       address: 89b0a86583c4444acfd71b463e0d3c55ae1412a5
 
-identityId: "0x010101010101"
+identityId: "0x89b0a86583c4444acfd71b463e0d3c55ae1412a5"
```
manuelpolzhofer: Is it a dummy address or do we need a real identity contract for the testing_config?

config/configstore/service.go
```diff
@@ -81,63 +82,51 @@ func (s service) GenerateAccount() (config.Account, error) {
 		return nil, err
 	}
 
-	id, confirmations, err := s.idService.CreateIdentity(ctx, identity.RandomCentID())
+	DID, err := s.idFactory.CreateIdentity(ctx)
```
manuelpolzhofer: should we call it `DID` or `did`?

identity/did.go
```diff
@@ -17,14 +51,63 @@ func (d DID) ToAddress() common.Address {
...
+	return d.ToAddress().String()
+}
+
+// BigInt returns DID in bigInt
+func (c DID) BigInt() *big.Int {
+	return utils.ByteSliceToBigInt(c[:])
+}
+
+// Equal checks if d == other
+func (d DID) Equal(other DID) bool {
```
manuelpolzhofer: 👍

identity/did.go
```diff
@@ -64,6 +150,12 @@ type ServiceDID interface {
 	// ValidateKey checks if a given key is valid for the given centrifugeID.
 	ValidateKey(ctx context.Context, did DID, key []byte, purpose int64) error
 
+	// ValidateSignature checks if signature is valid for given identity
+	ValidateSignature(signature *coredocumentpb.Signature, message []byte) error
```
manuelpolzhofer: I think this method is not needed. I replaced it with `IsSignedWithPurpose`.

identity/ideth/service.go
```diff
@@ -137,7 +138,7 @@ func (i service) AddKey(ctx context.Context, key id.KeyDID) error {
 
 	// TODO: did can be passed instead of randomCentID after CentID is DID
```
manuelpolzhofer:
```suggestion

```

identity/ideth/service.go
```diff
@@ -166,7 +167,7 @@ func (i service) AddMultiPurposeKey(ctx context.Context, key [32]byte, purposes
 	}
 
 	// TODO: did can be passed instead of randomCentID after CentID is DID
```
manuelpolzhofer:
```suggestion

```

identity/ideth/service.go
```diff
@@ -317,6 +318,21 @@ func (i service) GetKeysByPurpose(did id.DID, purpose *big.Int) ([][32]byte, err
...
 }
 
+// CurrentP2PKey returns the latest P2P key
+func (i service) CurrentP2PKey(did id.DID) (ret string, err error) {
+	keys, err := i.GetKeysByPurpose(did, big.NewInt(id.KeyPurposeP2P))
+	if err != nil {
+		return ret, err
+	}
+	lastKey := keys[len(keys)-1]
+	p2pID, err := ed25519.PublicKeyToP2PKey(lastKey)
```
manuelpolzhofer: The method is similar to `GetClientP2PURL` I think we could combine some parts after the migration.

Maybe we could add the following check. 
```golang
if key.RevokedAt.Cmp(big.NewInt(0)) != 0 {
       return "", errors.New("p2p key has been revoked")
}
```

config/configstore/model.go
```diff
@@ -574,15 +611,15 @@ func (acc *Account) GetKeys() (idKeys identity.IDKeys, err error) {
 	}
 	address32Bytes := utils.AddressTo32Bytes(common.HexToAddress(secp256k1.GetAddress(pk)))
 
-	keys[identity.KeyPurposeEthMsgAuth] = identity.IDKey{
+	keys[identity.KeyPurposeEthMsgAuth] = config.IDKey{
```
lucasvo: @manuelpolzhofer Should we still use this purpose here? I thought we got rid of it?

build/configs/testing_config.yaml
mikiquantum: It happens to be a real address, as I took it from the eth_account, but I believe that it doesnt get used.

config/configstore/service.go
mikiquantum: I like DID for method arguments and I extended that to internal variables within the method too. But I am open to suggestions.

identity/did.go
mikiquantum: Maybe a dummy question, but that smart contract function, extracts the public key out of the message and signature provided, and checks that the identity has that key. Is that equivalent to verifying that a message has been signed with the key that is claimed to? Is there any security concern in there?

manuelpolzhofer: Yes, it is the same.   The only difference I see is that `ValidateSignature` implicit assume using the signing key for the check and the `IsSignedWithPurpose` needs explicit the purpose.

config/configstore/model.go
manuelpolzhofer: We still need to rename the purpose constants. It would be a purpose "ACTION_KEY" after the refactoring. If that was the question?

PR #763, opened 2019-02-20
Id migration: new anchoring

manuelpolzhofer:
- new anchor contract used
- txManager instead of events
- CommitAnchor
  - async task


anchors/bootstrapper.go
```diff
@@ -29,26 +31,26 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	}
 	client := ctx[ethereum.BootstrappedEthereumClient].(ethereum.Client)
 
-	repositoryContract, err := NewEthereumAnchorRepositoryContract(cfg.GetContractAddress(config.AnchorRepo), client.GetEthClient())
+	anchorContractAddr := cfg.GetContractAddress(config.AnchorRepo)
+	fmt.Println(anchorContractAddr.String())
```
mikiquantum: remove plz

anchors/anchor_repository_integration_test.go
```diff
@@ -3,125 +3,84 @@
...
 
-	watchCommittedAnchor := <-confirmations
-	assert.Nil(t, watchCommittedAnchor.Error, "No error should be thrown by context")
-	assert.Equal(t, watchCommittedAnchor.CommitData.AnchorID, anchorIDTyped, "Resulting anchor should have the same ID as the input")
-	assert.Equal(t, watchCommittedAnchor.CommitData.DocumentRoot, docRootTyped, "Resulting anchor should have the same document hash as the input")
 }
 
+/*
+
+//TODO can be removed because CommitAnchor is sync
```
mikiquantum: even though we it might not be relevant for our use case of anchoring, I would like to keep it here, as it tests out our tx pool mechanism. Unless we cannot trigger it anymore?

testingutils/setup.go
```diff
@@ -139,7 +139,7 @@ func SetupSmartContractAddresses(cfg config.Configuration, sca *config.SmartCont
 func BuildIntegrationTestingContext() map[string]interface{} {
 	projDir := GetProjectDir()
 	StartPOAGeth()
-	RunSmartContractMigrations()
+	//RunSmartContractMigrations()
```
mikiquantum: uncomment

PR #768, opened 2019-02-22
Feat/cleanup Core document

vedhavyas:


documents/doc.go
```diff
@@ -1,15 +1,15 @@
 /*
-Package documents implements Centrifuge document models.
```
vedhavyas: I will revert these

crypto/sign.go
```diff
@@ -42,9 +42,9 @@ func VerifySignature(pubKey, message, signature []byte) error {
 
 // Sign the document with the private key and return the signature along with the public key for the verification
 // assumes that signing root for the document is generated
-func Sign(centIDBytes []byte, privateKey []byte, pubKey []byte, payload []byte) *coredocumentpb.Signature {
+func Sign(id []byte, privateKey []byte, pubKey []byte, payload []byte) *coredocumentpb.Signature {
```
lucasvo: Keytools shouldn't deal with creating protobuf messages IMO. I would prefer to move this out of the keytools package. I think this would be a good issue to add to house-keeping and address later.

documents/invoice/model.go
```diff
@@ -259,79 +264,53 @@ func (i *Invoice) getInvoiceSalts(invoiceData *invoicepb.InvoiceData) (*proofs.S
...
-// Note: this is not a unique identifier for each version of the document.
-func (i *Invoice) ID() ([]byte, error) {
-	coreDocModel, err := i.PackCoreDocument()
-	if err != nil {
-		return []byte{}, err
-	}
-	return coreDocModel.Document.DocumentIdentifier, nil
-}
-
 // PackCoreDocument packs the Invoice into a Core Document
```
lucasvo:
```suggestion
// PackCoreDocument packs the Invoice into a CoreDocument
```

documents/invoice/model.go
```diff
@@ -259,79 +264,53 @@ func (i *Invoice) getInvoiceSalts(invoiceData *invoicepb.InvoiceData) (*proofs.S
...
 // If the, Invoice is new, it creates a valid identifiers
```
lucasvo:
```suggestion
// If the, Invoice is new, it creates a valid identifier
```

lucasvo: Actually looking at the method, this comment is incorrect. Is that true?

documents/invoice/model.go
```diff
@@ -349,13 +328,20 @@ func (i *Invoice) Type() reflect.Type {
 	return reflect.TypeOf(i)
 }
 
-// CalculateDataRoot calculates the data root and sets the root to core document
-func (i *Invoice) CalculateDataRoot() ([]byte, error) {
+// DataRoot calculates the data root and sets the root to core document
+func (i *Invoice) DataRoot() ([]byte, error) {
```
lucasvo: I would like to leave the method name here, calculate implies it modifies the document. It is not a simple getter. Don't you think that might be missed here?

documents/invoice/model.go
```diff
@@ -259,79 +264,53 @@ func (i *Invoice) getInvoiceSalts(invoiceData *invoicepb.InvoiceData) (*proofs.S
...
+	invData := i.createP2PProtobuf()
+	data, err := proto.Marshal(invData)
 	if err != nil {
-		return nil, errors.NewTypedError(err, errors.New("couldn't serialise InvoiceData"))
+		return cd, errors.New("couldn't serialise InvoiceData: %v", err)
 	}
 
-	invoiceAny := any.Any{
+	embedData := &any.Any{
 		TypeUrl: documenttypes.InvoiceDataTypeUrl,
```
lucasvo: You could use the method you added below here:
```suggestion
		TypeUrl: i.DocumentType()
```

documents/invoice/model.go
```diff
@@ -377,20 +363,61 @@ func (i *Invoice) getDocumentDataTree() (tree *proofs.DocumentTree, err error) {
...
-	// There can be failure scenarios where the core doc for the particular document
-	// is still not saved with roots in db due to failures during getting signatures.
-	_, err = i.PackCoreDocument()
+	tree, err := i.getDocumentDataTree()
 	if err != nil {
 		return nil, errors.New("createProofs error %v", err)
 	}
 
-	tree, err := i.getDocumentDataTree()
+	return i.CoreDocument.CreateProofs(documenttypes.InvoiceDataTypeUrl, tree, fields)
```
lucasvo:
```suggestion
	return i.CoreDocument.CreateProofs(i.DocumentType(), tree, fields)
```

documents/invoice/service.go
```diff
@@ -54,17 +55,15 @@ func DefaultService(
...
-// DeriveFromCoreDocumentModel takes a core document model and returns an invoice
-func (s service) DeriveFromCoreDocumentModel(dm *documents.CoreDocumentModel) (documents.Model, error) {
-	var model documents.Model = &Invoice{
-		CoreDocumentModel: dm,
-	}
-	err := model.UnpackCoreDocument(dm)
+// DeriveFromCoreDocument takes a core document model and returns an invoice
+func (s service) DeriveFromCoreDocument(cd coredocumentpb.CoreDocument) (documents.Model, error) {
+	inv := new(Invoice)
+	err := inv.UnpackCoreDocument(cd)
```
lucasvo: 👍

documents/processor.go
```diff
@@ -129,65 +106,45 @@ func (dp defaultProcessor) RequestSignatures(ctx context.Context, model Model) e
 		return errors.New("failed to validate model for signature request: %v", err)
 	}
 
-	err = dp.p2pClient.GetSignaturesForDocument(ctx, dm)
+	signs, err := dp.p2pClient.GetSignaturesForDocument(ctx, model)
```
lucasvo:
```suggestion
	sigs, err := dp.p2pClient.GetSignaturesForDocument(ctx, model)
```

crypto/sign.go
vimukthi-git: agree

mikiquantum: We are going to use the sign method that is part of the Account (all keys live in that context), there is a function already in the package. It is debatable to return the Signature proto there or just a byte array but I like the idea of having the signing as part of the account struct.

documents/invoice/model.go
vedhavyas: It doesn't calculate always. only when the data root is missing, else returns the already calculated one. Hence renamed

documents/read_acls.go
```diff
@@ -0,0 +1,336 @@
...
+	// check if the account can read the doc
+	if cd.AccountCanRead(account) {
+		return nil
+	}
+
+	// check if the nft is present in read rules
+	found := findRole(cd.Document, func(_, _ int, role *coredocumentpb.Role) bool {
+		_, found := isNFTInRole(role, registry, tokenID)
+		return found
+	}, coredocumentpb.Action_ACTION_READ)
```
charleenfei: you can use the Action constant here (ACLRead)

documents/read_acls.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// AccountCanRead validate if the core Document can be read by the account .
+// Returns an error if not.
+func (cd *CoreDocument) AccountCanRead(account identity.CentID) bool {
+	// loop though read rules, check all the rules
+	return findRole(cd.Document, func(_, _ int, role *coredocumentpb.Role) bool {
+		_, found := isAccountInRole(role, account)
+		return found
+	}, coredocumentpb.Action_ACTION_READ, coredocumentpb.Action_ACTION_READ_SIGN)
```
charleenfei: same as above, ACLReadSign

documents/read_acls.go
```diff
@@ -0,0 +1,336 @@
...
+// addNFTToReadRules adds NFT token to the read rules of core Document.
+func (cd *CoreDocument) addNFTToReadRules(registry common.Address, tokenID []byte) error {
+	nft, err := ConstructNFT(registry, tokenID)
+	if err != nil {
+		return errors.New("failed to construct NFT: %v", err)
+	}
+
+	role := &coredocumentpb.Role{RoleKey: utils.RandomSlice(32)}
+	role.Nfts = append(role.Nfts, nft)
+	cd.addNewRule(role, coredocumentpb.Action_ACTION_READ)
```
charleenfei: same as above

documents/read_acls.go
```diff
@@ -0,0 +1,336 @@
...
+		z, found := isNFTInRole(role, registry, tokenID)
+		if found {
+			rridx = i
+			ridx = j
+			rk = role.RoleKey
+			nftIdx = z
+		}
+
+		return found
+	}, coredocumentpb.Action_ACTION_READ)
```
charleenfei: ^

documents/read_acls.go
vedhavyas: Lets be explict

documents/read_acls.go
vedhavyas: Dont need to create already existing constants

documents/read_acls.go
vedhavyas: These are mostly internal. We dont need to create short forms and relying on protobuf generated constants is better IMO

documents/invoice/model.go
```diff
@@ -377,20 +355,60 @@ func (i *Invoice) getDocumentDataTree() (tree *proofs.DocumentTree, err error) {
...
+// AddNFT adds NFT to the Invoice.
+func (i *Invoice) AddNFT(grantReadAccess bool, registry common.Address, tokenID []byte) error {
+	cd, err := i.CoreDocument.AddNFT(grantReadAccess, registry, tokenID)
+	if err != nil {
+		return err
 	}
 
-	proofs, err = i.CoreDocumentModel.CreateProofs(tree, fields)
-	return proofs, err
+	i.CoreDocument = cd
```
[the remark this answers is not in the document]
vedhavyas: Because AddNFT will not mutate the existing NFT but rather prepare a new version, add NFT to that and return that CD

documents/coredocument.go
```diff
@@ -0,0 +1,522 @@
...
+// NewCoreDocumentWithCollaborators generates new core Document, adds collaborators, adds read rules and fills salts
+func NewCoreDocumentWithCollaborators(collaborators []string) (*CoreDocument, error) {
+	cd := newCoreDocument()
+	ids, err := identity.NewDIDsFromStrings(collaborators)
+	if err != nil {
+		return nil, errors.New("failed to decode collaborators: %v", err)
+	}
+
+	for i := range ids {
+		cd.Document.Collaborators = append(cd.Document.Collaborators, ids[i][:])
```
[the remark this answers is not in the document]
vedhavyas: it will be in the following PR. Too many changes in this one already

documents/coredocument.go
```diff
@@ -0,0 +1,522 @@
...
+}
+
+// GetCollaborators returns the collaborators from the role with READ_SIGN ability.
+func (cd *CoreDocument) GetCollaborators(filterIDs ...identity.DID) (ids []identity.DID, err error) {
+	exclude := make(map[string]struct{})
+	for _, id := range filterIDs {
+		exclude[id.String()] = struct{}{}
+	}
+
+	for _, c := range cd.Document.Collaborators {
```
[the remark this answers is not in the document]
vedhavyas: yeap.

documents/coredocument.go
```diff
@@ -0,0 +1,522 @@
...
+			continue
+		}
+
+		ids = append(ids, id)
+	}
+
+	return ids, nil
+}
+
+func fetchUniqueCollaborators(oldCollabs [][]byte, newCollabs []string) (ids []identity.DID, err error) {
```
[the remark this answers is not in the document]
vedhavyas: okay. Will change it to `fetchNewUniqueCollaborators`

Issue #755, opened 2019-02-18
Add prefix to subtrees


mikiquantum:
We need to add a deterministic prefix to the coredocument tree and signing tree.
Both for human readable and compact property.

Open questions:
- Should we adapt the proof value to take the prefix? or abstract that logic to the node, to always try all the trees for a property match. What if compact property has a match in 2 trees (bad)?

Keep in sync with NFT hardcoded proofs.

#### Testing
<Describe required tests: Unit, Integration, Testworld or Functional>

#### Acceptance Criteria
<What is the acceptance criteria of this ticket?>

#### Dependencies and affected external components
<Outline any dependencies and components that could be affected by this>

lucasvo:
For now, let's not abstract it away from the client, it's a source of errors (and security issues).

mikiquantum:
I am leaning towards that as well. Let's make sure then we give Razvan the new compact properties so they can be hardcoded in the NFT SCs.

vedhavyas:
@vimukthi-git can we close this?

vimukthi-git:
@vedhavyas nope, the signing tree prefixes needs to be handled with this

mikiquantum:
Closing this as the signature tree will add the prefix as part of its own spec.

PR #770, opened 2019-02-22
Add prefixes to Signing and CoreDoc trees

vimukthi-git:
Issue Link: #755


documents/model.go
```diff
@@ -31,14 +32,23 @@ const (
...
-	SigningRootField:  {0, 0, 0, 10},
+func compactProperties(key string) []byte {
+	m := map[string][]byte{
+		CDRootField:       {0, 0, 0, 7},
+		DataRootField:     {0, 0, 0, 5},
+		DocumentTypeField: {0, 0, 0, 100},
+		SignaturesField:   {0, 0, 0, 6},
+		SigningRootField:  {0, 0, 0, 10},
+		CDTreePrefix:      {0, 0, 0, 11},
+		SigningTreePrefix: {0, 0, 0, 12},
```
vimukthi-git: @rdinicut the prefixes and their keys CDTreePrefix and SigningTreePrefix

documents/invoice/model.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const prefix string = "invoice"
 
-var compactPrefix = []byte{1, 0, 0, 0}
+func compactPrefix() []byte { return []byte{1, 0, 0, 0} }
```
lucasvo: I think it would be nice to move this to centrifuge-protobufs: https://github.com/centrifuge/centrifuge-protobufs/blob/master/documenttypes/prototypes.go

documents/model.go
lucasvo: This could also be a candidate to move to the other repo?

documents/model.go
```diff
@@ -290,21 +301,23 @@ func (m *CoreDocumentModel) GetDocumentTree() (tree *proofs.DocumentTree, err er
...
-	coreDocTree, err := m.GetDocumentTree()
+	coreDocTree, err := m.GetCoreDocumentTree()
 	if err != nil {
 		return nil, err
 	}
 
 	// create the signing tree with data root and coredoc root as siblings
-	tree = NewDefaultTree(ConvertToProofSalts(m.Document.CoredocumentSalts))
+	tree = NewDefaultTreeWithPrefix(ConvertToProofSalts(m.Document.CoredocumentSalts), SigningTreePrefix, compactProperties(SigningTreePrefix))
+	parentProp := NewLeafProperty(SigningTreePrefix, compactProperties(SigningTreePrefix))
```
lucasvo: Why not call it propPrefix?

```suggestion
	propPrefix := NewLeafProperty(SigningTreePrefix, compactProperties(SigningTreePrefix))
```

documents/model.go
```diff
@@ -328,13 +341,13 @@ func (m *CoreDocumentModel) GetDocumentRootTree() (tree *proofs.DocumentTree, er
...
 	// The first leave added is the signing_root
-	err = tree.AddLeaf(proofs.LeafNode{Hash: document.SigningRoot, Hashed: true, Property: NewLeafProperty(SigningRootField, compactProperties[SigningRootField])})
+	err = tree.AddLeaf(proofs.LeafNode{Hash: document.SigningRoot, Hashed: true, Property: NewLeafProperty(SigningRootField, compactProperties(SigningRootField))})
 	if err != nil {
 		return nil, err
 	}
 
 	// For every signature we create a LeafNode
-	sigProperty := NewLeafProperty(SignaturesField, compactProperties[SignaturesField])
+	sigProperty := NewLeafProperty(SignaturesField, compactProperties(SignaturesField))
```
lucasvo: Wouldn't it be possible to call flatten with `document.Signatures` here instead of iterating through them in line 364?

lucasvo: Or will you implement this in a separate PR?

vimukthi-git: Im not sure how this would work, let me check. I think it makes sense to do the changes in another pr together with moving stuff out of the repo. Trying to keep changes to a minimum so that we can move ahead with NFTs.

lucasvo: OK, I'm fine with that. It should work out of the box, you can just create a tree from the signatures message.

vimukthi-git:
@lucasvo Will wait until identity updates are merged to merge this
cc @mikiquantum

documents/model.go
mikiquantum: A couple of questions come up for me:
1. Do we need 4 bytes here? Or we can get away with it with let's say 2.
2. Should we follow a different pattern between prefixes and fields? invoice/po are `{x,0,0,0}` vs CDTree and SigningTreePrefix `{0,0,0,x}`.
I am fine renaming invoice/po or these ones, but would be better to keep it consistent.

@vimukthi-git @lucasvo thoughts?

lucasvo: We could switch two two bytes, though I wouldn't do less. I think @vimukthi-git picked 32 because that's the standard field length. Though we are not required to oblige to it at all.

vimukthi-git: @lucasvo yes, I have no opinion on this, just using uint32.
@mikiquantum I can make that change.

PR #777, opened 2019-02-25
remove bytecode check

mikiquantum:
Issue Link: <Github link>


identity/ideth/factory.go
```diff
@@ -89,12 +89,10 @@ func isIdentityContract(identityAddress common.Address, client ethereum.Client)
 		return err
 	}
 
-	deployedContractByte := common.Bytes2Hex(contractCode)
-	idbc := client.GetContractBytecode(config.Identity)
-	identityContractByte := idbc[2:] // remove 0x prefix
-	if deployedContractByte != identityContractByte {
-		return errors.New("deployed identity contract bytecode not correct")
+	if len(contractCode) == 0 {
```
mikiquantum: Simplified check.

manuelpolzhofer:
```golang
contractCode = "0x"
len(contractCode) = ?
```
I think it would be 2

mikiquantum: contract code is a byte array it will never have `0x`

PR #779, opened 2019-02-26
Correcting convention for document trees

vimukthi-git:
Issue Link: <Github link>


vimukthi-git:
those are {1,0,0,0}, {2,0,0,0} etc.

PR #780, opened 2019-02-26
Sign with secp

mikiquantum:
Changes:
- Using secp as signing curve
- Small cleanup on IDConfig that eventually should die
- Centralized the usage of sign and verify in crypto package and in account


cmd/common_test.go
```diff
@@ -78,7 +78,7 @@ func TestCreateConfig(t *testing.T) {
 	// Keys exists
 	// type KeyPurposeEthMsgAuth
 	idSrv := ctx[identity.BootstrappedDIDService].(identity.ServiceDID)
-	pk, _, err := secp256k1.GetEthAuthKey(cfg.GetEthAuthKeyPair())
+	pk, _, err := secp256k1.GetSigningKeyPair(cfg.GetEthAuthKeyPair())
```
lucasvo: Aren't we getting rid of ethauthkey?

config/configstore/model.go
```diff
@@ -546,12 +546,11 @@ func (acc *Account) GetEthereumContextWaitTimeout() time.Duration {
 
 // SignMsg signs a message with the signing key
 func (acc *Account) SignMsg(msg []byte) (*coredocumentpb.Signature, error) {
-	//TODO change signing keys to curve ed25519
 	keys, err := acc.GetKeys()
 	if err != nil {
 		return nil, err
 	}
-	signature, err := crypto.SignMessage(keys[identity.KeyPurposeSigning].PrivateKey, msg, crypto.CurveEd25519, true)
+	signature, err := crypto.SignMessage(keys[identity.KeyPurposeSigning].PrivateKey, msg, crypto.CurveSecp256K1, true)
```
lucasvo: I think we should get rid of `crypto.SignMessage` long term and directly call the secp package or at least have a distinct way to do ethereum signatures that hash the message first.

documents/validator.go
```diff
@@ -250,17 +250,20 @@ func readyForSignaturesValidator(didBytes, priv, pub []byte) Validator {
 			return errors.New("expecting only one signature")
 		}
 
-		s := crypto.Sign(didBytes, priv, pub, cd.SigningRoot)
+		s, err := crypto.SignMessage(priv, cd.SigningRoot, crypto.CurveSecp256K1, true)
```
lucasvo: I don't understand why we actually need to sign here to verify...  What does this check do?

p2p/receiver/handler_integration_test.go
```diff
@@ -146,7 +153,7 @@ func TestHandler_HandleInterceptorReqSignature(t *testing.T) {
 	assert.NotNil(t, resp.Signature.Signature, "must be non nil")
 	sig := resp.Signature
 	doc := dm.Document
-	assert.True(t, ed25519.Verify(sig.PublicKey, doc.SigningRoot, sig.Signature), "signature must be valid")
+	assert.True(t, secp256k1.VerifySignatureWithAddress(common.BytesToAddress(sig.PublicKey).String(), hexutil.Encode(sig.Signature), doc.SigningRoot), "signature must be valid")
```
lucasvo: I think it's much better to call secp directly like it is done here.

vimukthi-git: I would like to keep some wrapping for signing and verification in `crypto` package so that we don't endup having to refactor many places when changing the signature schemes

cmd/common_test.go
mikiquantum: Yes I believe so, I rather do that as part of a new PR. We need to coordinate with the UI as well.

documents/validator.go
mikiquantum: This check is triggered when an update/create in a document and before the request signatures from peers is triggered. It checks that the signature[0] "creator" is valid. I am not sure if this is completely needed. In the case that we want a precheck at that stage, we should check all the signatures in the document not only the first one.

mikiquantum: A verify on each of them instead of sign and validate.

p2p/receiver/handler_integration_test.go
mikiquantum: I agree on the wrapper, I havent added yet but was on my mind as soon as I got confirmation that we didnt need any other type of signing for secp curves.

crypto/sign_test.go
```diff
@@ -7,44 +7,35 @@ import (
 	"os"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
+
```
vedhavyas: fix the spacing pls

crypto/sign_test.go
```diff
@@ -92,7 +84,8 @@ func TestSignAndVerifyMessageEthereum(t *testing.T) {
 	fmt.Println("signature:", hexutil.Encode(signature))
 	fmt.Println("Generated Signature can also be verified at https://etherscan.io/verifySig")
 
-	correct := VerifyMessage(publicKey, testMsg, signature, CurveSecp256K1, true)
+	pk32 := utils.AddressTo32Bytes(common.HexToAddress(address))
+	correct := VerifyMessage(pk32[:], testMsg, signature, CurveSecp256K1)
```
vedhavyas: I think there are a bunch of redundant tests since we moved to default eth prefix for secp256k. Can you remove those as well?

crypto/verify.go
```diff
@@ -3,23 +3,19 @@ package crypto
...
 	switch curveType {
 	case CurveSecp256K1:
-		if ethereumVerify {
-			address := secp256k1.GetAddress(publicKey)
-			return secp256k1.VerifySignatureWithAddress(address, hexutil.Encode(signatureBytes), message)
-		}
-
-		return secp256k1.VerifySignature(publicKey, message, signatureBytes)
+		signatureBytes := make([]byte, len(signature))
+		copy(signatureBytes, signature)
```
vedhavyas: this is not needed since you are passing a hex to the Verify, the memory space is different.

documents/service.go
```diff
@@ -218,11 +219,16 @@ func (s service) RequestDocumentSignature(ctx context.Context, model Model) (*co
...
+	acc, err := contextutil.Account(ctx)
 	if err != nil {
 		return ErrDocumentConfigAccountID
 	}
 
+	idBytes, err := acc.GetIdentityID()
+	if err != nil {
+		return err
+	}
+	DID := identity.NewDIDFromBytes(idBytes)
```
vedhavyas:
```suggestion
	did := identity.NewDIDFromBytes(idBytes)
```

PR #783, opened 2019-02-26
Use getter method instead of public variable in anchor contract

charleenfei:
Post-Identity fix:

The node gets the anchor by directly accessing the public variable Commits. We should use the getter method for it. Commits will be private in the next version


manuelpolzhofer:
AnchorRespositry was the only place where we used public variables in the contracts directly?

anchors/service.go
```diff
@@ -20,7 +20,10 @@ import (
 type anchorRepositoryContract interface {
 	PreCommit(opts *bind.TransactOpts, _anchorID *big.Int, signingRoot [32]byte, expirationBlock *big.Int) (*types.Transaction, error)
 	Commit(opts *bind.TransactOpts, anchorID *big.Int, documentRoot [32]byte, documentProofs [][32]byte) (*types.Transaction, error)
-	Commits(opts *bind.CallOpts, anchorID *big.Int) (docRoot [32]byte, err error)
+	GetAnchorById(opts *bind.CallOpts, anchorID *big.Int) (struct {
```
vedhavyas: lets not use anonymous structs. create a type for this and use that

anchors/service_test.go
```diff
@@ -21,10 +21,20 @@ type mockAnchorRepo struct {
...
 
-func (m *mockAnchorRepo) Commits(opts *bind.CallOpts, anchorID *big.Int) (docRoot [32]byte, err error) {
+func (m *mockAnchorRepo) GetAnchorById(opts *bind.CallOpts, anchorID *big.Int) (struct {
+	AnchorID     *big.Int
+	DocumentRoot [32]byte
+}, error) {
 	args := m.Called(opts, anchorID)
-	docRoot, _ = args.Get(0).([32]byte)
-	return docRoot, args.Error(1)
+	type Response struct {
```
vedhavyas: once the struct is created, you can pass the struct as is instead of creating one here

PR #786, opened 2019-02-28
Enabling all nft tests

vimukthi-git:
Issue Link: <Github link>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: go
 go:
   - 1.11.4
 
+node_js:
+  - "10.15.1"
+
 # Maybe this helps with building branches in a fork
 go_import_path: github.com/centrifuge/go-centrifuge
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: go
 go:
   - 1.11.4
 
-node_js:
-  - "10.15.1"
-
 # Maybe this helps with building branches in a fork
 go_import_path: github.com/centrifuge/go-centrifuge
 
@@ -26,6 +23,7 @@ cache:
     - /tmp/depcache
 
 before_install:
+  - nvm install 10.15.1
   - npm install -g truffle@4.1.13
 
 install:
--- a/bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
+++ b/bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
@@ -38,9 +38,9 @@ var bootstappers = []bootstrap.TestBootstrapper{
 	&invoice.Bootstrapper{},
 	&purchaseorder.Bootstrapper{},
 	&nft.Bootstrapper{},
-	&queue.Starter{},
 	p2p.Bootstrapper{},
 	documents.PostBootstrapper{},
+	&queue.Starter{},
 }
 
 func TestFunctionalEthereumBootstrap() map[string]interface{} {
--- a/build/scripts/migrate.sh
+++ b/build/scripts/migrate.sh
@@ -32,7 +32,7 @@ fi
 if [[ "X${FORCE_MIGRATE}" == "Xtrue" ]];
 then
     echo "Running the Solidity contracts migrations for local geth"
-    sleep 5 # allow geth block gas limit to increase to more than 7000000
+    sleep 30 # allow geth block gas limit to increase to more than 7000000
     ${CENT_ETHEREUM_CONTRACTS_DIR}/scripts/migrate.sh localgeth
 else
     echo "Not migrating the Solidity contracts"
--- a/build/scripts/migrate.sh
+++ b/build/scripts/migrate.sh
@@ -32,6 +32,7 @@ fi
 if [[ "X${FORCE_MIGRATE}" == "Xtrue" ]];
 then
     echo "Running the Solidity contracts migrations for local geth"
+    sleep 5 # allow geth block gas limit to increase to more than 7000000
     ${CENT_ETHEREUM_CONTRACTS_DIR}/scripts/migrate.sh localgeth
 else
     echo "Not migrating the Solidity contracts"
--- a/identity/ideth/execute_integration_test.go
+++ b/identity/ideth/execute_integration_test.go
@@ -19,15 +19,13 @@ import (
 	"context"
 	"testing"
 
-	"github.com/centrifuge/go-centrifuge/testingutils/identity"
-
-	"github.com/centrifuge/go-centrifuge/config"
-
 	"github.com/centrifuge/go-centrifuge/anchors"
 	"github.com/centrifuge/go-centrifuge/bootstrap"
+	"github.com/centrifuge/go-centrifuge/config"
 	"github.com/centrifuge/go-centrifuge/ethereum"
 	id "github.com/centrifuge/go-centrifuge/identity"
 	"github.com/centrifuge/go-centrifuge/queue"
+	"github.com/centrifuge/go-centrifuge/testingutils/identity"
 	"github.com/centrifuge/go-centrifuge/transactions"
 	"github.com/centrifuge/go-centrifuge/utils"
 	"github.com/ethereum/go-ethereum/common"
--- a/identity/ideth/execute_integration_test.go
+++ b/identity/ideth/execute_integration_test.go
@@ -42,6 +42,7 @@ func getAnchorAddress(cfg config.Configuration) common.Address {
 }
 
 func TestExecute_successful(t *testing.T) {
+	t.SkipNow()
 	did := deployIdentityContract(t)
 	aCtx := getTestDIDContext(t, *did)
 	idSrv := initIdentity()
--- a/identity/ideth/factory_contract.go
+++ b/identity/ideth/factory_contract.go
@@ -4,6 +4,7 @@
 package ideth
 
 import (
+	"math/big"
 	"strings"
 
 	ethereum "github.com/ethereum/go-ethereum"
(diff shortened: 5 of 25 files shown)


nft/ethereum_payment_obligation.go
```diff
@@ -267,39 +276,49 @@ type MintRequest struct {
...
 
 type proofData struct {
+	Props  [][]byte
 	Values [][]byte
 	Salts  [][32]byte
 	Proofs [][][32]byte
 }
 
 func createProofData(proofspb []*proofspb.Proof) (*proofData, error) {
+	readRoleIndex := 5
```
vimukthi-git: these should be handled in a better way

bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
```diff
@@ -38,9 +38,9 @@ var bootstappers = []bootstrap.TestBootstrapper{
 	&invoice.Bootstrapper{},
 	&purchaseorder.Bootstrapper{},
 	&nft.Bootstrapper{},
-	&queue.Starter{},
 	p2p.Bootstrapper{},
 	documents.PostBootstrapper{},
+	&queue.Starter{},
```
vimukthi-git: strange that previous code worked for integration tests.

vedhavyas: Why do you think so?
The starter bootstrapper only needs the queue server which is put into context by queueBoostrapper. I think it should be working before this change

nft/ethereum_payment_obligation.go
```diff
@@ -314,7 +333,7 @@ func createProofData(proofspb []*proofspb.Proof) (*proofData, error) {
 		proofs[i] = property
 	}
 
-	return &proofData{Values: values, Salts: salts, Proofs: proofs}, nil
+	return &proofData{Props: props, Values: values, Salts: salts, Proofs: proofs}, nil
```
vedhavyas: Why are these properties required?

bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
vimukthi-git: Queue must be started after all task types are registered, we were skipping the anchorTask before because start being before `documents.PostBootstrapper{}`

nft/payment_obligation_integration_test.go
```diff
@@ -191,27 +195,9 @@ func mintNFTWithProofs(t *testing.T, grantAccess, tokenProof, readAccessProof bo
 }
 
 func TestEthereumPaymentObligation_MintNFT(t *testing.T) {
-	t.SkipNow()
 	tests := []struct {
 		grantAccess, tokenProof, readAccessProof bool
 	}{
-		{
```
vimukthi-git: removing these as they don't serve any purpose

testworld/nft_test.go
```diff
@@ -6,38 +6,39 @@ import (
...
 )
 
 func TestPaymentObligationMint_invoice_successful(t *testing.T) {
-	t.SkipNow()
 	t.Parallel()
 	tests := []struct {
 		name                                     string
 		grantAccess, tokenProof, readAccessProof bool
 	}{
-		{
```
vimukthi-git: these are also not very useful

vimukthi-git:
@mikiquantum now the only test that fails is `execute_integration_test.TestExecute_successful` do you have an idea why?

bootstrap/bootstrappers/testingbootstrap/testing_bootstrap.go
vedhavyas: I think that is still fine since the task registration can still happen after the queue starts

identity/ideth/execute_integration_test.go
```diff
@@ -44,6 +42,7 @@ func getAnchorAddress(cfg config.Configuration) common.Address {
 }
 
 func TestExecute_successful(t *testing.T) {
+	t.SkipNow()
```
vimukthi-git: @mikiquantum @manuelpolzhofer I decided to skip this for now, so that we can integrate the nfts and since the feature is not used. We can debug later.

vimukthi-git: May be @manuelpolzhofer can look at it if this is something important for audit?

manuelpolzhofer: I will fix the test. asap the PR is merged

vimukthi-git:
@mikiquantum 
@vedhavyas You are welcome to review and make comments about any remaining stuff. I will merge this now to unblock other stuff. 
I will also add more non-happy path nft tests to testworld later.

build/scripts/migrate.sh
```diff
@@ -32,6 +32,7 @@ fi
 if [[ "X${FORCE_MIGRATE}" == "Xtrue" ]];
 then
     echo "Running the Solidity contracts migrations for local geth"
+    sleep 30 # allow geth block gas limit to increase to more than 7000000
```
mikiquantum: Is this the fix to wait until at least one block mined so the contracts can be created? I saw some code with a loop too. one or the other might be redundant?

testingutils/setup.go
```diff
@@ -39,33 +39,49 @@ func RunSmartContractMigrations() {
...
+
+	payobdat, err := findContractDeployJSON("PaymentObligation.json")
 	if err != nil {
 		panic(err)
 	}
-	idFactoryAddrOp := getOpForContract(".contracts.IdentityFactory.address")
-	anchorRepoAddrOp := getOpForContract(".contracts.AnchorRepository.address")
-	payObAddrOp := getOpForContract(".contracts.PaymentObligation.address")
+
+	addrOp := getOpForContract(".networks.8383.address")
```
mikiquantum: with this hardcoded network ID we will not be able to run integration/testworld tests against rinkeby/ropsten...

.travis.yml
```diff
@@ -23,6 +23,7 @@ cache:
     - /tmp/depcache
 
 before_install:
+  - nvm install 10.15.1
```
mikiquantum: why do we need to install nvm?

vedhavyas:
@vimukthi-git just noticed this. I think most of us ignore the github emails :-D. I do ignore them since its just too much noise.
I think we should start reaching out to the reviewers on slack instead IMO

vimukthi-git:
@vedhavyas in gmail you can create nice filters, so that you can find relevant ones :)

build/scripts/migrate.sh
vimukthi-git: the other one is for local, above is for travis. For local there may be a docker geth running already so the retry works instead of just waiting unnecessarily. Ideally we should query geth for the block gas limit, but I found that creates other issues, above seems simple enought

testingutils/setup.go
vimukthi-git: We don't migrate contracts for other networks, its read from config, so it should be fine.

.travis.yml
vimukthi-git: its needed for latest smart contracts

identity/ideth/factory_contract.go
```diff
@@ -14,8 +15,20 @@ import (
...
+	_ = big.NewInt
+	_ = strings.NewReader
+	_ = ethereum.NotFound
+	_ = abi.U256
+	_ = bind.Bind
+	_ = common.Big1
+	_ = types.BloomLookup
+	_ = event.NewSubscription
+)
+
```
vedhavyas: What was the issue here ?

nft/ethereum_payment_obligation.go
```diff
@@ -267,40 +287,59 @@ type MintRequest struct {
...
+		}
+		if i == tokenRoleIndex {
+			props[1] = p.GetCompactName()
+		}
+
+		if i < 3 {
+			values[i] = p.Value
+		}
+
+		if i == readRoleIndex {
```
vedhavyas: you are already doing the same check above, move the logic there instead

nft/ethereum_payment_obligation.go
vedhavyas: @vimukthi-git why are these required?

testworld/nft_test.go
```diff
@@ -6,38 +6,39 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/centrifuge/go-centrifuge/documents"
+
```
vedhavyas: fix the spacing here pls

testworld/nft_test.go
vedhavyas: can remove them instead of commenting pls

nft/ethereum_payment_obligation.go
vimukthi-git: Its required by the smart contract to construct certain properties to verify
you can ask @rdinicut

testworld/nft_test.go
vimukthi-git: Im keeping them to enable failing tests later

nft/ethereum_payment_obligation.go
vimukthi-git: thanks, missed that. But this is something Im going to refactor soon as commented

identity/ideth/factory_contract.go
vimukthi-git: I don't know, thats how the bindings are generated

.travis.yml
vimukthi-git: Im not installing NVM btw, its already there on travis. Im just installing the latest version of nodejs.

PR #787, opened 2019-02-28
Adapt purposes + remove ethauth

mikiquantum:
New purposes: https://hackmd.io/twV_x2O1SvKlklWMZT7IBg?both

Added ACTION key when creating an identity, NODE eth address by default.

Removed usage of ethauth key

Removed usage of IDConfig


identity/did.go
```diff
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/centrifuge/go-centrifuge/crypto/secp256k1"
-	"github.com/ethereum/go-ethereum/common/hexutil"
```
vedhavyas: can we fix the spacing here ?

identity/did.go
```diff
@@ -29,22 +28,31 @@ const (
...
-	// KeyPurposeEthMsgAuth represents a key used for ethereum txns
-	KeyPurposeEthMsgAuth = 3
-
 	// KeyTypeECDSA has the value one in the ERC725 identity contract
 	KeyTypeECDSA = 1
 )
 
+// KeyPurposeP2PDiscovery is calculated out of Hex(sha256("CENTRIFUGE@P2P_DISCOVERY"))
+func KeyPurposeP2PDiscovery() *Purpose {
+	hashed := "88dbd1f0b244e515ab5aee93b5dee6a2d8e326576a583822635a27e52e5b591e"
```
vedhavyas: lets create a constant and use it. I'd actually prefer calculating hash instead of having a hash. What do you think ?

vedhavyas: better yet, we can have these purposes as global variables that are calculated through init function instead of creating every time through the function call. Better for memory allocation

config/configstore/model.go
```diff
@@ -550,7 +526,7 @@ func (acc *Account) SignMsg(msg []byte) (*coredocumentpb.Signature, error) {
 	if err != nil {
 		return nil, err
 	}
-	signature, err := crypto.SignMessage(keys[identity.KeyPurposeSigning].PrivateKey, msg, crypto.CurveSecp256K1)
+	signature, err := crypto.SignMessage(keys[identity.KeyPurposeSigning().Name].PrivateKey, msg, crypto.CurveSecp256K1)
```
vedhavyas: The key lookup can be simplfied if we have the constant string, right ?

config/configstore/model.go
```diff
@@ -562,51 +538,66 @@ func (acc *Account) SignMsg(msg []byte) (*coredocumentpb.Signature, error) {
...
 
+func (acc *Account) getEthereumAccountAddress() ([]byte, error) {
+	var ethAddr struct {
+		Address string `json:"address"`
+	}
+	err := json.Unmarshal([]byte(acc.GetEthereumAccount().Key), &ethAddr)
+	if err != nil {
+		return nil, err
+	}
+	return hexutil.Decode("0x" + ethAddr.Address)
```
vedhavyas: You can use the std lib `hex` for this instead of prepending "0x" IMO

contextutil/context.go
```diff
@@ -52,6 +43,19 @@ func TX(ctx context.Context) transactions.TxID {
 	return tid
 }
 
+// AccountDID extracts the AccountConfig DID from the given context value
+func AccountDID(ctx context.Context) (identity.DID, error) {
+	acc, err := Account(ctx)
+	if err != nil {
+		return identity.DID{}, err
```
vedhavyas: lets name the return values so that you dont have to recreate the DID for every return

documents/invoice/service.go
```diff
@@ -71,13 +71,13 @@ func (s service) DeriveFromCreatePayload(ctx context.Context, payload *clientinv
 		return nil, documents.ErrDocumentNil
 	}
 
-	id, err := contextutil.Self(ctx)
+	selfDID, err := contextutil.AccountDID(ctx)
```
vedhavyas:
```suggestion
	did, err := contextutil.AccountDID(ctx)
```

identity/did.go
```diff
@@ -29,21 +25,83 @@ const (
...
+	hashed := "774a43710604e3ce8db630136980a6ba5a65b5e6686ee51009ed5f3fded6ea7e"
+	v, _ := new(big.Int).SetString(hashed, 16)
+	return Purpose{Name: keyPurposeSigning, HexValue: hashed, Value: v}
+}
+
+// Purpose contains the different representation of purpose along the code
+type Purpose struct {
+	Name     string
+	HexValue string
+	Value    *big.Int
```
vedhavyas: I think we should not use the pointer here for Value. Since we are returning the Purpose as is, I can potentially change it even through a copy

utils/tools.go
```diff
@@ -76,13 +76,28 @@ func AddressTo32Bytes(address common.Address) [32]byte {
...
+func ByteArrayTo32BytesLeftPadded(in []byte) ([32]byte, error) {
+	byte32 := [32]byte{}
+	if len(in) > 32 {
+		return byte32, errors.New("incorrect input length %d should be 32", len(in))
+	}
+	if len(in) == 0 {
+		return byte32, nil
+	}
+	for i := 1; i <= len(in); i++ {
+		byte32[32-i] = in[len(in)-i]
```
vedhavyas: I feel this can be simplified.

pseudo code
```
toBepadded := 32 - len(in)
out := append(make([]byte, toBePadded, toBePadded), in...)
return utils.SliceToByte32(out)
```

identity/did.go
mikiquantum: yikes! Good catch!

contextutil/context.go
mikiquantum: Actually in this scenario err would be shadowed from the named return. So I think it is cleaner to keep the explicit return.

PR #790, opened 2019-02-28
Updated ReadACL tests to validate signing keys

charleenfei:

### DIFF
--- a/documents/read_acls.go
+++ b/documents/read_acls.go
@@ -77,7 +77,7 @@ func (cd *CoreDocument) NFTOwnerCanRead(tokenRegistry TokenRegistry, registry co
 	}, coredocumentpb.Action_ACTION_READ)
 
 	if !found {
-		return errors.New("nft not found in the Document")
+		return ErrNftNotFound
 	}
 
 	// get the owner of the NFT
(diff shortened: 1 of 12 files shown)


documents/read_acls.go
```diff
@@ -361,16 +358,29 @@ func (cd *CoreDocument) ATOwnerCanRead(tokenID, docID []byte, account identity.D
 		return true
 	}, coredocumentpb.Action_ACTION_READ)
 
+	if !found {
+		return at, errors.New("access token not found")
```
vimukthi-git: Lets move the errors in this file in to error.go as constants.

vedhavyas: It is not a standard practice in go @vim. We can create a new error constant but new fileis not needed since the definition should be in a place where the error is used

vimukthi-git: @vedhavyas It's a standard in our project now, we discussed this and we agreed. Besides go projects does it by defining an errors package that has all the errors in it. We don't want to define all errors in our errors package. Thats why we collect errors to an error.go. Having properly defined error apis has only upsides. Then there is only one place to look for, for errors that can be check for.

Even if someone doesn't agree we don't break our convention just for one file so that it looks like 3 different people wrote the code. So let's please move the error constants.

PR #791, opened 2019-02-28
New Signature structure

mikiquantum:
Issue Link: <Github link>


documents/coredocument.go
```diff
@@ -304,38 +343,22 @@ func (cd *CoreDocument) DocumentRootTree() (tree *proofs.DocumentTree, err error
...
-			Property: sigProperty.SliceElemProp(proofs.FieldNumForSliceLength(i)),
-		}
-
-		err = leaf.HashNode(h, true)
-		if err != nil {
-			return nil, err
-		}
-
-		sigLeafList[i+1] = leaf
+	properties := signatureTree.PropertyOrder()
```
mikiquantum: We do not have a way to get all the leaves from the tree, @vedhavyas is adding that function I will merge after that is done.

documents/coredocument.go
```diff
@@ -46,6 +45,9 @@ const (
 
 	// SigningTreePrefix is the human readable prefix for signing tree props
 	SigningTreePrefix = "signing_tree"
+
+	// SignaturesTreePrefix is the human readable prefix for signing tree props
```
vedhavyas:
```suggestion
	// SignaturesTreePrefix is the human readable prefix for signature props
```

documents/coredocument.go
```diff
@@ -75,6 +78,7 @@ func newCoreDocument() *CoreDocument {
 		DocumentIdentifier: id,
 		CurrentVersion:     id,
 		NextVersion:        utils.RandomSlice(idSize),
+		SignatureData:      new(coredocumentpb.SignatureData),
```
vedhavyas: we don't need this if we checking for nil in `AppendSignatures` methods, right ?

documents/coredocument.go
```diff
@@ -287,6 +295,37 @@ func (cd *CoreDocument) getSigningRootProofHashes() (hashes [][]byte, err error)
 	return rootProof.SortedHashes, err
 }
 
+func (cd *CoreDocument) getSignatureDataSalts() ([]*coredocumentpb.DocumentSalt, error) {
```
vedhavyas: can we rename this as `setSignatureDataSalts` and not return the salts. Instead, we get the salts from the coredoc ? 
This will avoid any manipulation of salts once set to the document

documents/coredocument_test.go
```diff
@@ -4,9 +4,12 @@ package documents
 
 import (
 	"crypto/sha256"
+	"fmt"
 	"os"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common/hexutil"
+
```
vedhavyas: can you fix the import spacing pls

documents/coredocument.go
```diff
@@ -177,6 +184,7 @@ func (cd *CoreDocument) PrepareNewVersion(collaborators []string, initSalts bool
 		Roles:              cd.Document.Roles,
 		ReadRules:          cd.Document.ReadRules,
 		Nfts:               cd.Document.Nfts,
+		SignatureData:      new(coredocumentpb.SignatureData),
```
vedhavyas: We don't need this if we have nil check for `AppendSignatures`, right?

identity/ideth/service.go
```diff
@@ -448,7 +448,7 @@ func (i service) AddKeysForAccount(acc config.Account) error {
 
 // ValidateSignature validates a signature on a message based on identity data
 func (i service) ValidateSignature(signature *coredocumentpb.Signature, message []byte) error {
-	centID := id.NewDIDFromBytes(signature.EntityId)
+	centID := id.NewDIDFromBytes(signature.SignerId)
```
vedhavyas:
```suggestion
	did := id.NewDIDFromBytes(signature.SignerId)
```

documents/coredocument.go
```diff
@@ -304,38 +343,22 @@ func (cd *CoreDocument) DocumentRootTree() (tree *proofs.DocumentTree, err error
...
+	signatureTreeLeaves := make([]proofs.LeafNode, len(properties))
+	for _, prop := range properties {
+		i, leaf := signatureTree.GetLeafByProperty(prop.ReadableName())
+		// Need to set leaf hashed otherwise it will be rehashed by tree.Generate() below
+		leaf.Hashed = true
+		signatureTreeLeaves[i] = *leaf
 	}
 
-	err = tree.AddLeaves(sigLeafList)
+	err = tree.AddLeaves(signatureTreeLeaves)
```
vimukthi-git: I thought we would only have to add the roothash of the tree to docRootTree just like signingRoot,
why are we adding all the leaves to the tree?
@lucasvo is it not what you meant in our last discussion?

mikiquantum: So we can create a proof of the fields directly using the documentroot tree and not having to load a different tree for it.

lucasvo: Yes, you're definitely right @vimukthi-git it should just be the root hash of `getSignatureDataTree`

vimukthi-git: https://images.zenhubusercontent.com/56946663638e718c995ffc5e/5c974b8b-d7b7-4111-8713-4adf9edd8cbc

lucasvo: @mikiquantum we definitely wanted to achieve the opposite. The goal was not add extra levels to the root tree because performance wise it doesn't make sense.

mikiquantum: So that means we need to iterate over at max 3 trees to find fields to create proofs.

Gopkg.toml
```diff
@@ -28,7 +28,8 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 
 [[constraint]]
   name = "github.com/centrifuge/centrifuge-protobufs"
-  revision = "20c7fd5e4210e68d2a5662a8714d51ceeae031b8"
+  revision = "54e86196998895a9d7e5b34d63d9d7717e61fe75"
+  source = "github.com/mikiquantum/centrifuge-protobufs"
```
mikiquantum: will change as soon as it is merged.

documents/coredocument.go
```diff
@@ -231,53 +240,81 @@ func (cd *CoreDocument) addNewRule(role *coredocumentpb.Role, action coredocumen
...
+	treeProofs[dataPrefix] = newTreeProof(dataTree, append([][]byte{cdRoot}, srpHashes...))
+	treeProofs[SignaturesTreePrefix] = newTreeProof(signatureTree, srpHashes)
+	treeProofs[CDTreePrefix] = newTreeProof(cdTree, append([][]byte{dataRoot}, srpHashes...))
+
+	return generateProofs(fields, treeProofs)
+}
 
-	proofs = append(proofs, pfs...)
-	return proofs, nil
+func getDataTreePrefix(dataTree *proofs.DocumentTree) (string, error) {
```
mikiquantum: This can be removed as soon as we have merged: https://github.com/centrifuge/precise-proofs/pull/72.
And to merge that we have to integrate latest precise proofs changes as currently they break go-centrifuge.

documents/coredocument.go
```diff
@@ -296,6 +333,39 @@ func (cd *CoreDocument) GetSigningRootProof() (hashes [][]byte, err error) {
...
+		proofSalts, err := GenerateNewSalts(cd.Document.SignatureData, SignaturesTreePrefix, compactProperties(SignaturesTreePrefix))
+		if err != nil {
+			return nil, err
+		}
+		cd.Document.SignatureDataSalts = ConvertToProtoSalts(proofSalts)
+	}
+	return cd.Document.SignatureDataSalts, nil
+}
+
+// GetSignatureDataTree returns the merkle tree for the Signature Data root.
```
mikiquantum: core change

documents/coredocument.go
```diff
@@ -313,38 +383,15 @@ func (cd *CoreDocument) DocumentRootTree() (tree *proofs.DocumentTree, err error
...
-	sigProperty := NewLeafProperty(SignaturesField, compactProperties(SignaturesField))
-	sigLeafList := make([]proofs.LeafNode, len(cd.Document.Signatures)+1)
-	sigLengthNode := proofs.LeafNode{
-		Property: sigProperty.LengthProp(proofs.DefaultSaltsLengthSuffix),
-		Salt:     make([]byte, idSize),
-		Value:    []byte(fmt.Sprintf("%d", len(cd.Document.Signatures))),
-	}
-	h := sha256.New()
-	err = sigLengthNode.HashNode(h, true)
+	// Second leaf from the signature data tree
```
mikiquantum: core change

p2p/client.go
```diff
@@ -282,7 +282,7 @@ func validateSignatureResp(
 		return version.IncompatibleVersionError(header.NodeVersion)
 	}
 
-	err := ideth.ValidateCentrifugeIDBytes(resp.Signature.EntityId, receiver)
+	err := ideth.ValidateCentrifugeIDBytes(resp.Signature.SignerId, receiver)
```
vimukthi-git: this function should be moved to `identity` package.

documents/coredocument.go
```diff
@@ -231,53 +240,81 @@ func (cd *CoreDocument) addNewRule(role *coredocumentpb.Role, action coredocumen
 	cd.Document.ReadRules = append(cd.Document.ReadRules, rule)
 }
 
+// TreeProof is a helper structure to pass to create proofs
+type TreeProof struct {
```
vimukthi-git: 👍

p2p/client.go
vedhavyas: `ValidateCentrifugeIDBytes` sounds mouthful. Can we come up with a better name for this?

vimukthi-git: ValidateDIDBytes

documents/coredocument.go
mikiquantum: there are couple other places where we access signature data before it has been initialized.

PR #792, opened 2019-03-01
fixed identity execute tests by adding an action key

manuelpolzhofer:


build/scripts/tests/run_testworld.sh
```diff
@@ -8,7 +8,7 @@ eval "$cleanup"
 
 status=$?
 
-output="go test -coverprofile=profile.out -covermode=atomic -tags=testworld github.com/centrifuge/go-centrifuge/testworld 2>&1"
+output="go test -coverprofile=profile.out -covermode=atomic -tags=testworld github.com/centrifuge/go-centrifuge/testworld 2>&1 -v"
```
vimukthi-git: are you sure this works? I dont think this flag would be taken as valid

vedhavyas: Yeah, this should work. But its less confusing we move the `-v` before output discard.

```suggestion
output="go test -coverprofile=profile.out -covermode=atomic -tags=testworld ./testworld/... -v 2>&1"
```

PR #794, opened 2019-03-01
Enable precommit and hiding anchors

vimukthi-git:
Issue Link: #762


anchors/anchor_repository_integration_test.go
```diff
@@ -31,13 +34,68 @@ func TestMain(m *testing.M) {
...
 	anchorID := utils.RandomSlice(32)
+	signingRoot := utils.RandomSlice(32)
+
+	anchorIDTyped, _ := anchors.ToAnchorID(anchorID)
+	preCommitAnchor(t, anchorID, signingRoot)
+	valid := anchorRepo.HasValidPreCommit(anchorIDTyped)
+	assert.True(t, valid)
+}
+
+func TestPreCommit_CommitAnchor_Integration(t *testing.T) {
```
vimukthi-git: this verifies that pre-commit -> commit flow works with contract, its just a matter of verifying coredoc processing flow. 
So I think smart contracts are ready for pre-audit now. @lucasvo @pstehlik @rdinicut

testworld/park.go
```diff
@@ -270,7 +266,7 @@ func newHost(
 
 func (h *host) init() error {
 	if h.createConfig {
-		err := cmd.CreateConfig(h.dir, h.ethNodeUrl, h.accountKeyPath, h.accountPassword, h.network, h.apiPort, h.p2pPort, h.bootstrapNodes, h.txPoolAccess, h.p2pTimeout, h.smartContractAddrs)
+		err := cmd.CreateConfig(h.dir, h.ethNodeUrl, h.accountKeyPath, h.accountPassword, h.network, h.apiPort, h.p2pPort, h.bootstrapNodes, h.txPoolAccess, false, h.p2pTimeout, h.smartContractAddrs)
```
vimukthi-git: for now testworld hosts won't have pre-commit enabled, I will enable it later for some of the hosts.

documents/anchor.go
```diff
@@ -24,7 +25,7 @@ type updaterFunc func(id []byte, model Model) error
 
 // AnchorDocument add signature, requests signatures, anchors document, and sends the anchored document
 // to collaborators
-func AnchorDocument(ctx context.Context, model Model, proc AnchorProcessor, updater updaterFunc) (Model, error) {
+func AnchorDocument(ctx context.Context, model Model, proc AnchorProcessor, updater updaterFunc, preAnchor bool) (Model, error) {
```
mikiquantum: Is the argument `preAnchor` not used?

vimukthi-git: shit forgot to add the if condition, that means everything is getting pre-anchored

anchors/anchor_repository_integration_test.go
```diff
@@ -31,13 +33,68 @@ func TestMain(m *testing.M) {
...
+	assert.Equal(t, docRootTyped, gotDocRoot)
+}
+
+func TestCommitAnchor_Integration(t *testing.T) {
+	anchorIDPreImage := utils.RandomSlice(32)
+	h := sha256.New()
+	_, err := h.Write(anchorIDPreImage)
+	assert.NoError(t, err)
+	var anchorID []byte
+	anchorID = h.Sum(anchorID)
```
vedhavyas:
```suggestion
	anchorID = h.Sum(nil)
```

anchors/anchor_repository_integration_test.go
```diff
@@ -31,13 +33,68 @@ func TestMain(m *testing.M) {
 	os.Exit(result)
 }
 
-func TestCommitAnchor_Integration(t *testing.T) {
+func TestPreCommitAnchor_Integration(t *testing.T) {
 	anchorID := utils.RandomSlice(32)
+	signingRoot := utils.RandomSlice(32)
+
+	anchorIDTyped, _ := anchors.ToAnchorID(anchorID)
```
vedhavyas: lets not ignore the errors here. If at some point ToAnhorID logic changes, this will fail silently. And that is a debugging nightmare for that person who is making the change

anchors/service.go
```diff
@@ -45,32 +56,38 @@ func (s *service) GetDocumentRootOf(anchorID AnchorID) (docRoot DocumentRoot, er
...
+
 	tc, err := contextutil.Account(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	ethRepositoryContract := s.anchorRepositoryContract
-	opts, err := ethereum.GetClient().GetTxOpts(tc.GetEthereumDefaultAccountName())
+	txID := contextutil.TX(ctx)
+
```
vedhavyas: fix the spacing here

build/configs/testing_config.yaml
```diff
@@ -26,3 +26,6 @@ keys:
   ethauth:
     publicKey: ../../build/resources/ethauth.pub.pem
     privateKey: ../../build/resources/ethauth.key.pem
+
+anchoring:
+  precommit: true
```
vedhavyas: add new line maybe?

documents/coredocument.go
```diff
@@ -528,3 +537,22 @@ func (cd *CoreDocument) Signatures() (signatures []coredocumentpb.Signature) {
 
 	return signatures
 }
+
+func populateVersions(cd *coredocumentpb.CoreDocument, prevCD *coredocumentpb.CoreDocument) (err error) {
+	if prevCD != nil {
+		cd.PreviousVersion = prevCD.CurrentVersion
+		cd.CurrentVersion = prevCD.NextVersion
+		cd.CurrentPreimage = prevCD.NextPreimage
+	} else {
```
vedhavyas: you can avoid the else here if you move the nextImage and Nextverison to the top. That makes code easier and reduce cyclomatic complexity

documents/processor.go
```diff
@@ -4,13 +4,14 @@ import (
 	"context"
 	"time"
 
+	"github.com/centrifuge/go-centrifuge/utils"
+
```
vedhavyas: fix the spacing pls

utils/tools.go
```diff
@@ -189,3 +189,29 @@ func ConvertIntToByte32(n int) ([32]byte, error) {
...
+			return nil, err
+		}
+		property = append(property, hash32)
+	}
+
+	return property, nil
+}
+
+// GenerateHashPair generates a preimage and hash pair
+func GenerateHashPair(preimageSize int) (preimage, hash []byte, err error) {
```
vedhavyas: This function is only used in the `documents` package. Lets move it there instead

vimukthi-git: its used elsewhere as well,

documents/coredocument.go
vimukthi-git: Please confirm, I don't think what you are saying can be done. Since its dealing with current version no next version

anchors/anchor_repository_integration_test.go
vimukthi-git: I think its better the way that it is right now, as the value is in callers control

anchors/service.go
vimukthi-git: this is just my preference, as its logical separation from the code below

documents/processor.go
```diff
@@ -139,17 +169,23 @@ func (dp defaultProcessor) AnchorDocument(ctx context.Context, model Model) erro
 		return errors.New("failed to get document root: %v", err)
 	}
 
-	anchorID, err := anchors.ToAnchorID(model.CurrentVersion())
+	anchorID, err := anchors.ToAnchorID(model.CurrentVersionPreimage())
```
manuelpolzhofer: Shouldn't it the pre-image of the previous document version?

documents/coredocument.go
```diff
@@ -528,3 +537,22 @@ func (cd *CoreDocument) Signatures() (signatures []coredocumentpb.Signature) {
 
 	return signatures
 }
+
+func populateVersions(cd *coredocumentpb.CoreDocument, prevCD *coredocumentpb.CoreDocument) (err error) {
+	if prevCD != nil {
+		cd.PreviousVersion = prevCD.CurrentVersion
+		cd.CurrentVersion = prevCD.NextVersion
+		cd.CurrentPreimage = prevCD.NextPreimage
```
manuelpolzhofer: okay so we are storing the current preImage and the nextPreimage inside the CD.

Need to update that in the protocol paper

documents/coredocument.go
vedhavyas: You can rewrite it as

```
    cd.NextPreimage, cd.NextVersion, err = utils.GenerateHashPair(idSize)
	if err != nil {
		return err
	}

    if prevCD != nil {
		cd.PreviousVersion = prevCD.CurrentVersion
		cd.CurrentVersion = prevCD.NextVersion
		cd.CurrentPreimage = prevCD.NextPreimage
        return nil
	}

    cd.CurrentPreimage, cd.CurrentVersion, err = utils.GenerateHashPair(idSize)
    if err != nil {
		return err
	}
	cd.DocumentIdentifier = cd.CurrentVersion
	return nil
}
```

vedhavyas: The cyclomatci complexity in this case would 3 instead of 4

documents/coredocument.go
```diff
@@ -528,3 +537,22 @@ func (cd *CoreDocument) Signatures() (signatures []coredocumentpb.Signature) {
...
+	} else {
+		cd.CurrentPreimage, cd.CurrentVersion, err = utils.GenerateHashPair(idSize)
+		cd.DocumentIdentifier = cd.CurrentVersion
+		if err != nil {
+			return err
+		}
+	}
+	cd.NextPreimage, cd.NextVersion, err = utils.GenerateHashPair(idSize)
```
manuelpolzhofer: I don't understand the method name. Why it is a hash pair?
Why not two methods GeneratePreImage() and GenerateNextVersion(preImage)?

documents/coredocument.go
vimukthi-git: I think I would go with cyclomatic complexity of 4 than creating extra hashes :)

utils/tools.go
```diff
@@ -200,3 +200,29 @@ func ConvertIntToByte32(n int) ([32]byte, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		property = append(property, hash32)
+	}
+
+	return property, nil
+}
+
+// GenerateHashPair generates a preimage and hash pair
```
manuelpolzhofer: maybe we should add some extra comment here, explaining why are we generating the next version based on the hash of pre-image.

In the literature, it is a bit related to a concept called commit-reveal scheme.

utils/tools.go
```diff
@@ -200,3 +200,29 @@ func ConvertIntToByte32(n int) ([32]byte, error) {
...
+func GenerateHashPair(preimageSize int) (preimage, hash []byte, err error) {
+	preimage = RandomSlice(preimageSize)
+	h := sha256.New()
```
manuelpolzhofer: hash function should be only called inside the crypto package.

Let's generate a sha256 function in that package.

documents/processor.go
manuelpolzhofer: Please rename:
```suggestion
	anchorIDPreImage, err := anchors.ToAnchorID(model.CurrentVersionPreimage())
```

The preImage is not the anchorID. 

The anchorID == cd.CurrentVersion

Otherwise it is confusing for the reader :)

documents/coredocument.go
```diff
@@ -528,3 +537,22 @@ func (cd *CoreDocument) Signatures() (signatures []coredocumentpb.Signature) {
...
 }
+
+func populateVersions(cd *coredocumentpb.CoreDocument, prevCD *coredocumentpb.CoreDocument) (err error) {
+	if prevCD != nil {
+		cd.PreviousVersion = prevCD.CurrentVersion
+		cd.CurrentVersion = prevCD.NextVersion
+		cd.CurrentPreimage = prevCD.NextPreimage
+	} else {
+		cd.CurrentPreimage, cd.CurrentVersion, err = utils.GenerateHashPair(idSize)
+		cd.DocumentIdentifier = cd.CurrentVersion
```
manuelpolzhofer: What about cd.PreviousVersion and cd.previousHash. Can we assume the are nil or 0x here?

documents/processor.go
vimukthi-git: no we generate the new document with correct versions so this is the way it is now

documents/coredocument.go
vimukthi-git: we have to,

documents/coredocument.go
vimukthi-git: its just easier for our purpose

documents/coredocument.go
vimukthi-git: its nil

utils/tools.go
vimukthi-git: I wouldn't explain version stuff here as this function is general, but I have added a brief explanation on coredoc

PR #797, opened 2019-03-03
Integrate precise-proofs updates

lucasvo:
This PR implements the newest version of precise proofs that supports salts in coredocuments natively. I added a `mutable bool` to all of the methods that initialize a Merkle tree that if not set to true, it will fail if a salt is missing. This should be used whenever we verify a document or create a proof. Whenever we modify the tree (and the doc) we can set the flag to true knowing that after that the coredocument needs to be persisted.


documents/coredocument_test.go
```diff
@@ -186,18 +173,19 @@ func TestGetDocumentSigningTree(t *testing.T) {
...
 	cd.Document.DataRoot = utils.RandomSlice(32)
-	assert.NoError(t, cd.setSalts())
-	tree, err := cd.signingRootTree(documenttypes.InvoiceDataTypeUrl)
+	tree, err := cd.signingRootTree(documenttypes.InvoiceDataTypeUrl, true)
 	assert.Nil(t, err)
 	assert.NotNil(t, tree)
 
 	_, leaf := tree.GetLeafByProperty(SigningTreePrefix + ".data_root")
+	for _, l := range tree.GetLeaves() {
+		fmt.Printf("P: %s V: %v", l.Property.ReadableName(), l.Value)
```
vedhavyas: missed removing this?

documents/coredocument_test.go
```diff
@@ -223,7 +211,7 @@ func TestGetDocumentRootTree(t *testing.T) {
 
 func TestCoreDocument_GenerateProofs(t *testing.T) {
 	h := sha256.New()
-	testTree := NewDefaultTree(nil)
+	testTree := NewDefaultTree(&coredocumentpb.CoreDocument{}, false)
```
vedhavyas:
```suggestion
	testTree := NewDefaultTree(new(coredocumentpb.CoreDocument), false)
```

documents/invoice/model.go
```diff
@@ -347,13 +317,15 @@ func (i *Invoice) CalculateDataRoot() ([]byte, error) {
...
-func (i *Invoice) getDocumentDataTree() (tree *proofs.DocumentTree, err error) {
+func (i *Invoice) getDocumentDataTree(mutable bool) (tree *proofs.DocumentTree, err error) {
 	invProto := i.createP2PProtobuf()
-	salts, err := i.getInvoiceSalts(invProto)
 	if err != nil {
 		return nil, err
 	}
-	t := documents.NewDefaultTreeWithPrefix(salts, prefix, compactPrefix())
+	if i.CoreDocument == nil {
+		return nil, errors.New("getDocumentDataTree error CoreDocument not set")
```
vedhavyas:
```suggestion
		return nil, errors.New("CoreDocument is nil")
```

documents/read_acls.go
```diff
@@ -113,13 +113,13 @@ func (cd *CoreDocument) addNFTToReadRules(registry common.Address, tokenID []byt
...
 	cd.addNewRule(role, coredocumentpb.Action_ACTION_READ)
-	return cd.setSalts()
+	return nil
 }
 
 // AddNFT returns a new CoreDocument model with nft added to the Core Document. If grantReadAccess is true, the nft is added
 // to the read rules.
 func (cd *CoreDocument) AddNFT(grantReadAccess bool, registry common.Address, tokenID []byte) (*CoreDocument, error) {
-	ncd, err := cd.PrepareNewVersion(nil, false)
+	ncd, err := cd.PrepareNewVersion(nil)
```
vedhavyas: lets change the function definition of PrepareNewVersion to take variardic input instead of slice. Then the call would be 

```
cd.PrepareNewVersion() // with no collaborators

cd.PrepareNewVersion(collaborator) // with one collaborator

cd.PrepareNewVerison(collaborators...) // with multiple collaborators
```

documents/tree.go
```diff
@@ -29,13 +43,32 @@ func NewLeafProperty(literal string, compact []byte) proofs.Property {
...
-func GenerateNewSalts(document proto.Message, prefix string, compactPrefix []byte) (*proofs.Salts, error) {
-	docSalts := new(proofs.Salts)
-	t := NewDefaultTreeWithPrefix(docSalts, prefix, compactPrefix)
-	err := t.AddLeavesFromDocument(document)
-	if err != nil {
-		return nil, err
-	}
-	return docSalts, nil
+// DocumentSaltsFunc returns a function that fetches and sets salts on the CoreDoc. The boolean mutable can be used to define if the salts function should error if a new field is encountered or not.
+func DocumentSaltsFunc(coredoc *coredocumentpb.CoreDocument, mutable bool) (func(compact []byte) ([]byte, error), error) {
```
vedhavyas: Let's pass only the pointer to the salts instead of passing the full coredocument. That will make the functions more generic and single source to generate trees of various documents

PR #810, opened 2019-03-04
Expose p2p signature collection errors and ignore errors at correct level

vimukthi-git:
Issue Link: <Github link>


vimukthi-git:
@eswarasai this should help with your tests.

vimukthi-git:
@lucasvo This would be covered by testworld, right now there is no way to test this other than mocking, which doesn't makes sense for a p2p request.

PR #815, opened 2019-03-05
Feat(documents): add getChangedField for write_acls

### LINKED ISSUE #801: GetChangedFields
Add a function that takes two document trees and return the compact properties of the the fields that are changed in the new tree

Acceptence criteria:
1. Test cases covering all the edge cases

vedhavyas:
Closes #801


documents/coredocument.go
```diff
@@ -174,6 +174,7 @@ func (cd *CoreDocument) PrepareNewVersion(collaborators []string, initSalts bool
 		Roles:              cd.Document.Roles,
 		ReadRules:          cd.Document.ReadRules,
 		Nfts:               cd.Document.Nfts,
+		AccessTokens:       cd.Document.AccessTokens,
```
charleenfei: good catch.

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+// if the old is nil, then it is a set operation
+// if new is nil, then it is an unset operation
+// if both old and new are set, then it is an edit operation
+type changedField struct {
+	property, old, new []byte
+}
+
+// getChangedFields takes two document trees and returns the compact value of the fields that are changed in new tree.
+// properties may have been added to the new tree or removed from the new tree. Either case, since the new tree is different from old,
+// that is considered a change
```
charleenfei: nitpick but maybe this should be on the line above. also, "In either case,".

vedhavyas: this is not a new line, just a wrap since the comment is long enough.

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+	newProps := newTree.PropertyOrder()
+
+	props := make(map[string]proofs.Property)
+	for _, p := range append(oldProps, newProps...) {
+		// we can ignore the length property since any change in slice or map will return in addition or deletion of properties in the new tree
+		if p.Text == lengthSuffix {
+			continue
+		}
+
+		if _, ok := props[p.ReadableName()]; ok {
```
charleenfei: will there be line item instances where we shouldn't ignore repeated props?

documents/write_alcs_test.go
```diff
@@ -0,0 +1,250 @@
+// +build unit
```
vimukthi-git: filename ;)

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+		props[p.ReadableName()] = p
+	}
+
+	// check each property and append it changed fields if the value is different.
+	for _, p := range props {
+		_, ol := oldTree.GetLeafByProperty(p.ReadableName())
+		_, nl := newTree.GetLeafByProperty(p.ReadableName())
+
+		if ol == nil {
+			changedFields = append(changedFields, newChangedField(p, nl, false))
```
vimukthi-git: shouldn't this append be done only if the other field new or old is not nil?

documents/error.go
```diff
@@ -35,50 +35,21 @@ const (
...
-
 	// ErrDocumentProof must be used when document proof creation fails
 	ErrDocumentProof = errors.Error("document proof error")
 
 	// ErrDataRootInvalid must be used when the data root is invalid
 	ErrDataRootInvalid = errors.Error("data root is invalid")
 
-	// Document repository errors
-
-	// ErrDocumentRepositoryModelNotRegistered must be used when the model hasn't been registered in the database repository
```
vimukthi-git: 👍

documents/write_acls.go
charleenfei: in my understanding, old == nil in this case means that the field has been added in the new version of the doc, so the fields needs to be appended to the list of changed fields.

vedhavyas: So, if the property doesn't exist in old one but exists in new one, that means there is an addition operation in the new document. If the property is not found in the new document but found in the old document, then there was a deletion operation in the new document.

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/centrifuge/precise-proofs/proofs"
+)
+
+// changedField holds the compact property, old and new value of the field that is changed
+// if the old is nil, then it is a set operation
+// if new is nil, then it is an unset operation
+// if both old and new are set, then it is an edit operation
+type changedField struct {
+	property, old, new []byte
```
mikiquantum: surprised this `new` is not a reserved word here 😄

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+
+		ov := ol.Value
+		nv := nl.Value
+		if ol.Hashed {
+			ov = ol.Hash
+			nv = nl.Hash
+		}
+
+		if !bytes.Equal(ov, nv) {
+			changedFields = append(changedFields, changedField{
```
mikiquantum: why not using here the `newChangeField` method?

documents/write_acls.go
mikiquantum: Do you think it might be interesting to store the readable name as well to be returned?

vedhavyas: TBH, i'm not sure at the moment, if we need it somewhere, we can add it here then 👍

documents/write_acls.go
vedhavyas: newChangedField will only take once one document tree, either the old or new. Here we need to get values from both the trees

documents/write_acls.go
```diff
@@ -0,0 +1,86 @@
...
+		if _, ok := props[p.ReadableName()]; ok {
+			continue
+		}
+
+		props[p.ReadableName()] = p
+	}
+
+	// check each property and append it changed fields if the value is different.
+	for _, p := range props {
+		_, ol := oldTree.GetLeafByProperty(p.ReadableName())
```
vimukthi-git: p.ReadableName() call is not needed here as you can use the keys of the map in the range call.

documents/write_acls.go
vimukthi-git: The second for-loop is off putting to me, thats why I got confused, why do we need it? can't we just iterate over the entire list of fields and do skip same ones using the map while processing the changes?

documents/write_acls.go
vedhavyas: I would need the property later again to the compact name. I'm leaning towards saving the map call.

vimukthi-git:
@vedhavyas if build keeps failing without an error from your side, try removing the cache :) May be docker geth is getting in the way.

documents/write_acls.go
vedhavyas: lets talk offline, not sure I understand your question,

vimukthi-git: I meant removing the second for loop, I hope its obvious from what I said, anyway we can always discuss. I can also do the refactoring.

PR #818, opened 2019-03-06
Add smart contract addresses for v0.5.5

vimukthi-git:
Issue Link: <Github link>


vimukthi-git:
we need some kind of commit hook or check for this on travis

lucasvo:
@vimukthi-git isn't there a make command for it already? We could run it in travis and check for modified changes like we do with protobuf.

vimukthi-git:
@lucasvo there is a command, i can check how to add that check once im done with the current stuff.

PR #819, opened 2019-03-06
Feat(documents): validate transitions

### LINKED ISSUE #802: TransitionRulesForTheAccount
Add a function that returns all the fields a given account can change

### LINKED ISSUE #803: CompareFields
ComapreFields should take list of allowedFields and changedFields and should return an error if any of the changed fields are not part of allowedFields

Criteria:
Test with all possible scenarios with different document types ex: CoreDocument, invoice, PO etc...

vedhavyas:
Closes #802 and Closes #803


documents/write_acls.go
```diff
@@ -12,6 +15,7 @@ import (
 // if both old and new are set, then it is an edit operation
 type changedField struct {
 	property, old, new []byte
+	name               string
```
vedhavyas: @mikiquantum added the name here. We will return the names for invalid transitions

documents/write_acls.go
```diff
@@ -84,3 +89,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
...
+				Field:     copyBytes(rule.Field),
+				Action:    rule.Action,
+			})
+		}
+	}
+
+	return rules
+}
+
+func copyBytes(data []byte) []byte {
```
charleenfei: maybe these could be moved to utils?

documents/write_acls.go
```diff
@@ -84,3 +89,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
...
+	if data == nil {
+		return nil
+	}
+
+	nb := make([]byte, len(data), len(data))
+	copy(nb, data)
+	return nb
+}
+
+func copyByteSlice(data [][]byte) [][]byte {
```
charleenfei: same as above

documents/write_acls.go
vedhavyas: thought so but not used elsewhere. lets keep them here until the time we need to move them ther

vedhavyas: We should aim to not have utils package. it just against Go standards

documents/write_acls.go
```diff
@@ -59,6 +63,7 @@ func getChangedFields(oldTree, newTree *proofs.DocumentTree, lengthSuffix string
 
 		if !bytes.Equal(ov, nv) {
 			changedFields = append(changedFields, changedField{
+				name:     k,
```
vimukthi-git: please modify the first for loop above to be
```
props := make(map[string]proofs.Property)
	for _, p := range append(oldProps, newProps...) {
		// we can ignore the length property since any change in slice or map will return in addition or deletion of properties in the new tree
		if p.Text == lengthSuffix {
			continue
		}

		if _, ok := props[p.ReadableName()]; ok {
			continue
		}

		props[p.ReadableName()] = p

		_, ol := oldTree.GetLeafByProperty(p.ReadableName())
		_, nl := newTree.GetLeafByProperty(p.ReadableName())

		if ol == nil {
			changedFields = append(changedFields, newChangedField(p, nl, false))
			continue
		}

		if nl == nil {
			changedFields = append(changedFields, newChangedField(p, ol, true))
			continue
		}

		ov := ol.Value
		nv := nl.Value
		if ol.Hashed {
			ov = ol.Hash
			nv = nl.Hash
		}

		if !bytes.Equal(ov, nv) {
			changedFields = append(changedFields, changedField{
				property: p.CompactName(),
				old:      ov,
				new:      nv,
			})
		}
	}
```
and remove second.

documents/write_acls.go
vimukthi-git: Please move it there or use something already there. I don't think we are ready to move out utils to its own repo yet,
as per https://twitter.com/davecheney/status/697551191752855552?lang=en

documents/write_acls.go
vimukthi-git: same here

documents/write_acls.go
```diff
@@ -84,3 +89,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
...
+func validateTransitions(rules []coredocumentpb.TransitionRule, changedFields []changedField) error {
+	cfMap := make(map[string]struct{})
+	for _, cf := range changedFields {
+		cfMap[cf.name] = struct{}{}
+	}
+
+	for _, rule := range rules {
+		for _, cf := range changedFields {
+			if isValidTransition(rule, cf) {
+				delete(cfMap, cf.name)
```
vimukthi-git: Why don't you just append the error here, if its not a valid transition?

documents/write_acls.go
vedhavyas: Unless there is a place to move thse functions there, I would like to keep them here. We may not remove utils package but we wont put there unneeded stuff that is used by the single package. Lets keep the standards community defined rather than defining our own 👍

vedhavyas: Like the tweet says, we dont want to create a new utils project :)

vimukthi-git: could you refer here to the community convention you follow? I would like to read :)

documents/write_acls.go
vedhavyas: It may be an invalid transition for this rule doesn't mean its invalid altogether. We can only be sure about the transition is invalid once we iterate all the changes to all the rules

documents/write_acls.go
```diff
@@ -84,3 +86,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
 	cf.new = v
 	return cf
 }
+
+// transitionRulesFor returns a copy all the transition rules for the account.
+func (cd *CoreDocument) transitionRulesFor(account identity.DID) (rules []coredocumentpb.TransitionRule) {
```
mikiquantum: This comment is about naming. I think in cases like this `account` is only relevant from the context of hosting identities. I would use did/id similar here instead of account, as we have account struct as well can be confusing.

documents/write_acls.go
vimukthi-git: From my experience, there are problems with both approaches,
With utils
- You can announce a reusable function so that it can be reused by everyone else
- utils makes things well tested for all implementors 
Otherwise you idea `No one else needs` becomes a self full-filling prophecy, I like to keep track of reusable functions in utils and test them properly. Also building software projects properly is multi dimensional optimisation problem, what makes sense for a library might not make sense for a daemon like ours.

documents/write_acls.go
```diff
@@ -84,3 +86,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
...
+	for _, rule := range cd.Document.TransitionRules {
+		for _, rk := range rule.Roles {
+			role, err := getRole(rk, cd.Document.Roles)
+			if err != nil {
+				continue
+			}
+
+			if _, ok := isAccountInRole(role, account); !ok {
```
mikiquantum: Following comment above: `is[DID|ID]InRole` fits better.

documents/write_acls.go
```diff
@@ -84,3 +86,121 @@ func newChangedField(p proofs.Property, leaf *proofs.LeafNode, old bool) changed
...
+	// 1. update: editing a value
+	// 2. set: setting a new value ex: adding to slice or map
+	// 3. delete: deleting the new value ex: removing from slice or map
+	// Once we have more actions, like set, increment etc.. we can do those checks here
+	return true
+}
+
+// AccountCanUpdate validates the changes made by the account in the new document.
+// returns error if the transitions are not allowed for the account
+func (cd *CoreDocument) AccountCanUpdate(ncd *CoreDocument, account identity.DID, docType string) error {
```
mikiquantum: Same here with `IDCanUpdate`

vedhavyas: I'm leaning towards `CollaboratorCanUpdate` since ID seems very generic. What do you think ?

mikiquantum: Yep, collaborator fits well too.

vimukthi-git: I think `AccountCanUpdate` makes sense as ID is just a single part of an account. Also i don't mind `CollaboratorCanUpdate`.

documents/write_acls.go
vimukthi-git: @mikiquantum This is something we need to discuss as putting `is[DID|ID]InRole` just fixes the name to a DID, which is an implementation of centrifuge account IMO.

documents/write_acls.go
vimukthi-git: according to what I described in the tech talk, the `identities` package should be renamed to `accounts` in the future and all accounts stuff moved there. Then above param would look like `account identity.DID`

vedhavyas: Not sure if accounts as generic term floating around makes sense. In the Multi-tenency perspective sure. But DID take multiple forms like collaborator(when exchanging documents), account(multi-tenency), id(node self id) etc...

I wouldn't rename it to accounts.

vimukthi-git: It can take multiple/generic forms from a `name` perspective, but not from a perspective of a `type`. Account is part of the protocol, same as what an ethereum account means for us. Thats atleast the discussion that lead to naming multi-tenancy to multi account. We can discuss, when the time comes for reorg of code anyways.

vedhavyas: @vimukthi-git @mikiquantum lets take this offline. This may be not the right place to dicuss this :)

Issue #339, opened 2018-10-21
Validate key block revocation


vedhavyas:
`Check if revokation block happened before the timeframe of the document signing, for historical validations` for `ValidateKey` in identity.go

vimukthi-git:
I will use the signature timestamp in coredocumentpb.Signature to validate the key revocation.

PR #820, opened 2019-03-06
Key revoked validation

vimukthi-git:
Issue Link: #339


vimukthi-git:
@eswarasai After this is merged all you bounty can be merged and test cases should pass.

eswarasai:
@lucasvo @vedhavyas -- Is this good to be merged? Please do let me know once done, so that I can rebase and fix the other 2 test cases in the `testworld` for invalid signatures :)

identity/ideth/service.go
```diff
@@ -430,7 +456,8 @@ func (i service) AddKeysForAccount(acc config.Account) error {
 func (i service) ValidateSignature(signature *coredocumentpb.Signature, message []byte) error {
 	centID := id.NewDIDFromBytes(signature.EntityId)
 
-	err := i.ValidateKey(context.Background(), centID, signature.PublicKey, &(id.KeyPurposeSigning.Value))
+	sigTime := time.Unix(signature.Timestamp.Seconds, int64(signature.Timestamp.Nanos))
```
vedhavyas: lets not use the signature timestamp for this. As a bad actor, I can still change the time stamp in the signature since timestamp is not included in the signature. I would pass this as nil until we have the timestamp in the coredocument

identity/ideth/service.go
```diff
@@ -336,7 +337,7 @@ func (i service) Exists(ctx context.Context, did id.DID) error {
 }
 
 // ValidateKey checks if a given key is valid for the given centrifugeID.
-func (i service) ValidateKey(ctx context.Context, did id.DID, key []byte, purpose *big.Int) error {
+func (i service) ValidateKey(ctx context.Context, did id.DID, key []byte, purpose *big.Int, validateAt *time.Time) error {
```
vedhavyas: `validateAt` -> `validAt` seems less confusing to me

identity/ideth/service.go
```diff
@@ -347,12 +348,37 @@ func (i service) ValidateKey(ctx context.Context, did id.DID, key []byte, purpos
...
+			if err != nil {
+				return err
+			}
+
+			if big.NewInt(validateAt.Unix()).Cmp(revokedAtBlock.Time()) > 0 {
+				return errors.New("the given key [%x] for purpose [%s] has been revoked before provided time %s", key, purpose.String(), validateAt.String())
+			}
+		} else {
+			// else check if the revocation block number is before the latest block number from Ethereum
+			latest, err := i.client.GetEthClient().BlockByNumber(ctx, nil)
```
vedhavyas: I'm not sure if this is the correct check. If Alice signs a document at block 3 and key was revoked at 2 and latest block is at 6, this would still go through right ?

Shouldn't we be checking the blcok at which the document was anchored ?

identity/ideth/service.go
vimukthi-git: its `validateAt`, since we are saying validate at a particular time

identity/ideth/service.go
vimukthi-git: but what does bad actor gain out of that? Since signature is not part of any root, anything can be faked here

identity/ideth/service.go
vimukthi-git: its a fall back, the actual check we should use is the one with the time.

identity/ideth/service.go
vimukthi-git: we can also make it just `at`

identity/ideth/service.go
vedhavyas: he can fake the validation of the signature since we are relying on the timestamp to validate when key is revoked, right?

vimukthi-git: yes, but any validation here is done by the bad actor here, so he can just say that signature is not valid without even trying to validate. 
@lucasvo Is this something we should consider?

vimukthi-git: @lucasvo I think we confused two different validations here, I had a discussion with @manuelpolzhofer @vedhavyas on this and its clear to me that I was just talking about p2p signature validation vs you guys talking about NFT level key revoked validations.

p2p/receiver/validator.go
```diff
@@ -89,7 +89,8 @@ func peerValidator(idService identity.ServiceDID) Validator {
 		if err != nil {
 			return err
 		}
-		return idService.ValidateKey(context.Background(), *centID, idKey, &(identity.KeyPurposeP2PDiscovery.Value))
+		// TODO provide the time for validation here by adding the sent timestamp to the p2p header
```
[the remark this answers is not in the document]
vimukthi-git: This was just a confusion, I will change this pr to do all p2p signature validations against `now`

identity/ideth/service.go
vimukthi-git: I will remove this else block

vimukthi-git:
@vedhavyas I had a call with @lucasvo , Conclusion is that
we still need to validate against signature timestamp whenever we validate a signature, for example when receiving all ready anchored documents. So the timestamp validation will remain. I will remove the latest block based fallback anyway because it serves no purpose.

documents/read_acls.go
```diff
@@ -370,7 +370,8 @@ func (cd *CoreDocument) ATGranteeCanRead(ctx context.Context, idService identity
 		return ErrReqDocNotMatch
 	}
 	// validate that the public key of the granter is the public key that has been used to sign the access token
-	err = idService.ValidateKey(ctx, granterID, at.Key, &(identity.KeyPurposeSigning.Value))
+	// TODO provide the time for validation here using the signature timestamp
```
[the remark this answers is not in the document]
vimukthi-git: https://app.zenhub.com/workspaces/centrifuge-5ba350114b5806bc2be90978/issues/centrifuge/go-centrifuge/827

p2p/receiver/validator.go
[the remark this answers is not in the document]
vimukthi-git: Will do this in another pr

PR #823, opened 2019-03-07
Latest nft

mikiquantum:
Issue Link: <Github link>


Makefile
```diff
@@ -69,6 +69,24 @@ vendorinstall: ## Installs all protobuf dependencies with go-vendorinstall
...
 
+abigen-install: ## Installs ABIGEN from vendor
+abigen-install: vendorinstall
+	go-vendorinstall github.com/ethereum/go-ethereum/cmd/abigen
+
+gen-abi-bindings: ## Generates GO ABI Bindings
+gen-abi-bindings: install-deps abigen-install
+	$(eval CONTRACTS_VERSION := $(shell cat vendor/github.com/centrifuge/centrifuge-ethereum-contracts/package.json | jq -r '.version'))
+	npm install --prefix tmp/contracts @centrifuge/ethereum-contracts@${CONTRACTS_VERSION}
+	@cat tmp/contracts/node_modules/\@centrifuge/ethereum-contracts/build/contracts/Identity.json | jq '.abi' > tmp/contracts/id.abi
```
manuelpolzhofer: can we assume the contracts are compiled and the json always exists?

mikiquantum: the contracts are stored in npm and they contain the build files. And we get the version out of the vendor folder so unless there is a corrupt version uploaded to npm then we should be good.

Issue #809, opened 2019-03-04
Timestamps and authors in CoreDocument


lucasvo:
We need to add the timestamp a change was made to the CoreDocument as well as who made the change so that when signing a state update that is clearly tied to a particular time and coming from a particular node.

In addition to adding two fields, we should add some extra validations:
1) Sender of the request needs to be the author of the document
2) When receiving a signature request, one should only sign a document that has a timestamp < 10min in the past

vimukthi-git:
@lucasvo I just realise that we record timestamp in the coredocumentpb.Signature data structure. Since we also have the sender field to record who made the change, we already have this record. Just need to add the 2nd validation you mention. 
Therefore the issue could be simplified to just,
1. Add the sender.
2. Validation the senders signature and key in the signature list.
3. Validate timestamp.

Does this make sense?

lucasvo:
The signature timestamp is not part of the signature payload, right? So that alone won't be enough. I think we should add a change date to the document and add a set of validations:

General validation:
1) Ensure that the document change date - anchor timestamp < MAX_DOC_CHANGE_ANCHOR_DELTA
2) Ensure that the signature date is > doc change date 
3) Ensure that signature date is < anchor timestamp

When requesting a pre commit (so in the sign method):
* Ensure that document change date - now < MAX_DOC_CHANGE_ANCHOR_DELTA
* 2 & 3 Apply as well.

lucasvo:
As discussed earlier, we should remove the date on the signature. We will instead always assume that signature date == document timestamp. I would name the constant `MaximumDocumentTimestampAnchorTimeDifference` and let's set it to 3600 seconds (1hr).

vimukthi-git:
@lucasvo the implementation now is the following (as requested)
### PostAnchor
- Validate all signatures are validated against doc timestamp (no sig timestamp anymore)
- Validate author has signed 
- Validate Signers are signer collaborators
- Validate document was anchored within `MaxAuthoredToCommitDuration` which is 2 hours.

### PreAnchor
- Collaborators only sign a document within `MaxAuthoredToCommitDuration` which is 2 hours.
- Signature req Sender is author

PR #842, opened 2019-03-12
Update log for coredoc

vimukthi-git:
Issue Link: #809


Gopkg.toml
```diff
@@ -28,7 +28,8 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 
 [[constraint]]
   name = "github.com/centrifuge/centrifuge-protobufs"
-  revision = "aa8450c3ed93371fb76cf9cb1783e210177c88ae"
+  revision = "29ac1d3efe3bc5a10bd56dc0ab10b891cf64e2c4"
+  source = "github.com/vimukthi-git/centrifuge-protobufs"
```
vimukthi-git: to be updated when ready to merge

documents/processor.go
```diff
@@ -75,6 +75,16 @@ func (dp defaultProcessor) PrepareForSignatureRequests(ctx context.Context, mode
 		return err
 	}
 
+	id, err := self.GetIdentityID()
+	if err != nil {
+		return err
+	}
+
+	err = model.AddUpdateLog(identity.NewDIDFromBytes(id))
```
vimukthi-git: main change

documents/validator.go
```diff
@@ -178,13 +178,35 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
 			return errors.New("atleast one signature expected")
 		}
 
+		authorFound := false
 		for _, sig := range signatures {
-			if erri := idService.ValidateSignature(&sig, sr); erri != nil {
+			tm, terr := utils.FromTimestamp(sig.Timestamp)
+			sigDID := identity.NewDIDFromBytes(sig.SignerId)
+			if model.Author().Equal(sigDID) {
```
vimukthi-git: main change,

utils/time.go
```diff
@@ -13,3 +14,13 @@ func ToTimestamp(time time.Time) *timestamp.Timestamp {
 		Nanos:   int32(time.Nanosecond()),
 	}
 }
+
+// ToTimestamp converts time.Time to timestamp.TimeStamp.
+func ToTimestampProper(time time.Time) (*timestamp.Timestamp, error) {
```
vimukthi-git: this func is temporary till we merge the bounty pr

crypto/keytools.go
```diff
@@ -11,3 +14,23 @@ const (
 	CurveEd25519   string = "ed25519"
 	CurveSecp256K1 string = "secp256k1"
 )
+
+func ObtainP2PKeypair(pubKey, privKey string) (priv crypto.PrivKey, pub crypto.PubKey, err error) {
```
lucasvo: Obtain because Get is not allowed in Go? ;-) Why not call it `Load`? A bit more common.

documents/coredocument.go
```diff
@@ -586,6 +587,26 @@ func (cd *CoreDocument) Signatures() (signatures []coredocumentpb.Signature) {
 	return signatures
 }
 
+// AddUpdateLog adds a log to the model to persist an update related meta data such as author
```
lucasvo: This description seems outdated since we removed the log message?

documents/validator.go
```diff
@@ -178,13 +178,35 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
...
+			if erri := idService.ValidateSignature(sigDID, sig.PublicKey, sig.Signature, sr, tm); erri != nil {
 				err = errors.AppendError(
 					err,
 					errors.New("signature_%s verification failed: %v", hexutil.Encode(sig.SignerId), erri))
 			}
 		}
+		if !authorFound {
+			err = errors.AppendError(
+				err,
+				errors.New("signature verification failed: author not found"))
```
lucasvo:
```suggestion
				errors.New("signature verification failed: author's signature missing on document"))
```

documents/validator.go
```diff
@@ -178,13 +178,35 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
 			return errors.New("atleast one signature expected")
 		}
 
+		authorFound := false
 		for _, sig := range signatures {
-			if erri := idService.ValidateSignature(&sig, sr); erri != nil {
+			tm, terr := utils.FromTimestamp(sig.Timestamp)
```
lucasvo: Don't you want to remove the signature timestamp in this change as discussed? We don't need it and it's just misleading.

documents/validator.go
```diff
@@ -178,13 +178,35 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
...
+			sigDID := identity.NewDIDFromBytes(sig.SignerId)
+			if model.Author().Equal(sigDID) {
+				authorFound = true
+				// if author is found, check his signature relative to signed time on the document
+				tm, terr = model.Timestamp()
```
lucasvo: We should validate all signatures with the model timestamp not just the author.

crypto/keytools.go
vimukthi-git: there is a minor difference between a `getter`(take a value that is already a field, no external calls to db etc), `obtainer`(load a value and execute some logic to derive a value) and a `loader`(load from some where) ;) , So I will keep it this way.

documents/coredocument.go
vimukthi-git: but its still a log.. IMO

documents/validator.go
vimukthi-git: I will remove, just that we didn't discuss what to do for collaborator signatures

vimukthi-git:
@lucasvo I will complete all the validations you mention and the ones in the ticker in a separate pr, this pr is about adding the logging fields and using them so that I don't have to care about protobuf repo updates anymore.

p2p/receiver/validator.go
```diff
@@ -89,8 +89,13 @@ func peerValidator(idService identity.ServiceDID) Validator {
 		if err != nil {
 			return err
 		}
-		// TODO provide the time for validation here by adding the sent timestamp to the p2p header
-		return idService.ValidateKey(context.Background(), *centID, idKey, &(identity.KeyPurposeP2PDiscovery.Value), nil)
+
+		tm, err := utils.FromTimestamp(header.Timestamp)
```
vimukthi-git: shit, missed removing this

documents/validator.go
```diff
@@ -178,13 +178,33 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
 			return errors.New("atleast one signature expected")
 		}
 
+		authorFound := false
 		for _, sig := range signatures {
-			if erri := idService.ValidateSignature(&sig, sr); erri != nil {
+			tm, terr := model.Timestamp()
```
vimukthi-git: move down

documents/validator.go
```diff
@@ -178,13 +178,33 @@ func signaturesValidator(idService identity.ServiceDID) Validator {
...
+			sigDID := identity.NewDIDFromBytes(sig.SignerId)
+			if model.Author().Equal(sigDID) {
+				authorFound = true
+			}
+
+			// terr is updated twice in previous lines, we wait until final value is determined to check for error
```
vimukthi-git: remove

PR #846, opened 2019-03-14
Document consensus validations

vimukthi-git:
Issue Link: #809 

### PostAnchor
- Validate all signatures against doc timestamp (no sig timestamp anymore)
- Validate author has signed 
- Validate Signers are signer collaborators
- Validate document was anchored within `MaxAuthoredToCommitDuration` which is 2 hours.

### PreAnchor
- Collaborators only sign a document within `MaxAuthoredToCommitDuration` which is 2 hours.
- Signature req Sender is author


documents/error.go
```diff
@@ -23,6 +23,9 @@ const (
 	// ErrDocumentNil must be used when the provided document through a function is nil
 	ErrDocumentNil = errors.Error("no(nil) document provided")
 
+	// ErrDocumentNotification must be used when a notification about a document could not be delivered
+	ErrDocumentNotification = errors.Error("could not notify of the document")
```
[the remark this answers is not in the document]
vimukthi-git: its a general error not just for webhook :)

anchors/service.go
```diff
@@ -47,12 +49,13 @@ func (s *service) HasValidPreCommit(anchorID AnchorID) bool {
 	return r
 }
 
-// GetDocumentRootOf takes an anchorID and returns the corresponding documentRoot from the chain.
-func (s *service) GetDocumentRootOf(anchorID AnchorID) (docRoot DocumentRoot, err error) {
+// getAnchor takes an anchorID and returns the corresponding documentRoot from the chain.
```
vedhavyas:
```suggestion
// GetAnchor takes an anchorID and returns the corresponding documentRoot and anchored time from the chain.
```

documents/service.go
```diff
@@ -49,8 +48,8 @@ type Service interface {
 	// CreateProofsForVersion creates proofs for a particular version of the document given the fields
 	CreateProofsForVersion(ctx context.Context, documentID, version []byte, fields []string) (*DocumentProof, error)
 
-	// RequestDocumentSignature Validates and Signs document received over the p2p layer
-	RequestDocumentSignature(ctx context.Context, model Model) (*coredocumentpb.Signature, error)
+	// ReceiveDocumentSignatureRequest Validates and Signs document received over the p2p layer
+	ReceiveDocumentSignatureRequest(ctx context.Context, model Model, sender identity.DID) (*coredocumentpb.Signature, error)
```
vedhavyas: `RequestDocumentSignature` makes sense here since the intention is to get signature. `ReceiveDocumentSignatureRequest` implies SignatureRequest is being received. This is not what implementation is doing. Lets revet this pls

p2p/receiver/handler_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/centrifuge/go-centrifuge/testingutils/identity"
+
```
vedhavyas: import spacing pls

testingutils/documents/documents.go
```diff
@@ -5,6 +5,8 @@ package testingdocuments
 import (
 	"context"
 
+	identity2 "github.com/centrifuge/go-centrifuge/identity"
+
```
vedhavyas: same as above

documents/validator.go
```diff
@@ -236,6 +289,15 @@ func anchoredValidator(repo anchors.AnchorRepository) Validator {
 			return errors.New("mismatched document roots")
 		}
 
+		tm, err := model.Timestamp()
```
vedhavyas: `Timestamp` IMO doesn't imply anything on a model. Lets change this to `CreatedAt` ?

anchors/anchor_repository.go
```diff
@@ -13,6 +14,6 @@ var log = logging.Logger("anchorRepository")
 type AnchorRepository interface {
 	PreCommitAnchor(ctx context.Context, anchorID AnchorID, signingRoot DocumentRoot) (confirmations chan bool, err error)
 	CommitAnchor(ctx context.Context, anchorID AnchorID, documentRoot DocumentRoot, documentProofs [][32]byte) (chan bool, error)
-	GetDocumentRootOf(anchorID AnchorID) (DocumentRoot, error)
+	GetAnchor(anchorID AnchorID) (docRoot DocumentRoot, anchoredTime time.Time, err error)
```
vedhavyas: `GetAnchor` -> `GetAnchorData` would be apt

documents/service.go
vimukthi-git: I renamed it exactly because of the confusion in your comment this is actually receiving it

documents/validator.go
vimukthi-git: makes sense

vimukthi-git: unfortunately I would have to defer this as right now coredoc protobuf also has the same name.

documents/service.go
vedhavyas: `ReceiveDocumentSignatureRequest(ctx context.Context, model Model, sender identity.DID) (*coredocumentpb.Signature, error)`

The function is actually returning the signature. It is not receiving, no ?

PR #847, opened 2019-03-14
Integrate precise proofs update and removing salts

lucasvo:


documents/coredocument.go
```diff
@@ -469,6 +463,8 @@ func (cd *CoreDocument) documentTree(docType string, mutable bool) (tree *proofs
 	if err != nil {
 		return nil, err
 	}
+
+	cd.CoreDocModified = false
```
lucasvo: Why do you set it to false here? Wouldn't that lead to the flag getting unset without potentially storing it?

documents/coredocument.go
```diff
@@ -619,3 +615,12 @@ func populateVersions(cd *coredocumentpb.CoreDocument, prevCD *coredocumentpb.Co
 	}
 	return nil
 }
+
+// GetTestCoreDocWithReset must only be used by tests for manipulations. It gets the embedded coredoc protobuf.
+// All calls to this function will cause a regeneration of salts next time for precise-proof trees.
+func (cd *CoreDocument) GetTestCoreDocWithReset() *coredocumentpb.CoreDocument {
```
lucasvo: does this have to be in this file? it doesn't access any private fields, right?

documents/coredocument.go
```diff
@@ -70,6 +70,15 @@ func compactProperties(key string) []byte {
 
 // CoreDocument is a wrapper for CoreDocument Protobuf.
 type CoreDocument struct {
+	// DataModified indicates that the data of the model that embeds the CoreDocument has been modified and salts needs to be generated for new fields in data precise-proof tree.
+	DataModified bool
```
lucasvo: I think we don't need three different flags here. There is only one salts field anyway.

documents/tree.go
```diff
@@ -44,31 +33,33 @@ func NewLeafProperty(literal string, compact []byte) proofs.Property {
 }
 
 // DocumentSaltsFunc returns a function that fetches and sets salts on the CoreDoc. The boolean mutable can be used to define if the salts function should error if a new field is encountered or not.
-func DocumentSaltsFunc(coredoc *coredocumentpb.CoreDocument, mutable bool) (func(compact []byte) ([]byte, error), error) {
-	salts := coredoc.Salts
+func (coredoc *CoreDocument) DocumentSaltsFunc() func(compact []byte) ([]byte, error) {
+	salts := coredoc.Document.Salts
```
lucasvo: Change comment of the method

documents/coredocument.go
lucasvo: Actually never mind, I think on second thought it makes sense.

documents/coredocument.go
```diff
@@ -83,97 +86,79 @@ func newCoreDocument() (*CoreDocument, error) {
...
-	if err := cd.setSalts(); err != nil {
-		return nil, err
-	}
-
 	return cd, nil
 }
 
-// ID returns the Document identifier
+// SetDataModified sets the DataModified flag to the given value
+func (cd *CoreDocument) SetDataModified(modified bool) {
```
vimukthi-git: remove

documents/coredocument.go
vimukthi-git: yes, but this prevents tests from breaking all over the place, when coredoc changes. We can modify this function to help with the tests as well see all the test use cases.

documents/coredocument.go
vedhavyas: What does this do?
Is this supposed to be removed ?

codecov.yml
```diff
@@ -0,0 +1,5 @@
+ignore:
```
vedhavyas: 👍

documents/coredocument.go
```diff
@@ -70,6 +70,9 @@ func compactProperties(key string) []byte {
 
 // CoreDocument is a wrapper for CoreDocument Protobuf.
 type CoreDocument struct {
+	// CoreDocModified indicates that the CoreDocument has been modified and salts needs to be generated for new fields in coredoc precise-proof tree.
+	CoreDocModified bool
```
vedhavyas: `CoreDocModified` -> `Modified` would be less redundant :)

documents/coredocument.go
vedhavyas: I don't agree. If tests are broken, we should fix them accordingly, let's not add test related code into production. This function returns the pointer of the Core document. This means I can mutate the document. Move this function into test file or to file with build tags

documents/entity/model_test.go
```diff
@@ -279,14 +278,14 @@ func TestEntityModel_GetDocumentID(t *testing.T) {
...
 
-func TestEntityModel_getDocumentDataTree(t *testing.T) {
-	e := Entity{LegalName: "test company"}
-	tree, err := e.getDocumentDataTree()
-	assert.Nil(t, err, "tree should be generated without error")
-	_, leaf := tree.GetLeafByProperty("entity.legal_name")
-	assert.NotNil(t, leaf)
-	assert.Equal(t, "entity.legal_name", leaf.Property.ReadableName())
-}
+//func TestEntityModel_getDocumentDataTree(t *testing.T) {
```
vedhavyas: Why commented out ?

documents/invoice/model.go
```diff
@@ -704,6 +673,12 @@ func (i *Invoice) CalculateSigningRoot() ([]byte, error) {
 	return i.CoreDocument.CalculateSigningRoot(i.DocumentType())
 }
 
+// CalculateDocumentRoot calculate the document root
+// TODO: Should we add this
+func (i *Invoice) CalculateDocumentRoot() ([]byte, error) {
```
vedhavyas: we dont need this

nft/ethereum_payment_obligation.go
```diff
@@ -135,8 +132,14 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, req MintNFTRequ
 		return nil, nil, errors.NewTypedError(ErrNFTMinted, errors.New("registry %v", req.RegistryAddress.String()))
 	}
 
+	cidBytes, err := tc.GetIdentityID()
```
vedhavyas: `cidBytes` -> `idBytes`

nft/ethereum_payment_obligation.go
```diff
@@ -135,8 +132,14 @@ func (s *ethereumPaymentObligation) MintNFT(ctx context.Context, req MintNFTRequ
...
+	if err != nil {
+		return nil, nil, err
+	}
+
 	// Mint NFT within transaction
 	// We use context.Background() for now so that the transaction is only limited by ethereum timeouts
+	cid := identity.NewDIDFromBytes(cidBytes)
```
vedhavyas: `cid` -> `id`

documents/tree.go
```diff
@@ -29,41 +32,35 @@ func NewLeafProperty(literal string, compact []byte) proofs.Property {
...
 
-	return protoSalts
-}
-
-// ConvertToProofSalts converts protocolSalts into proofSalts
-func ConvertToProofSalts(protoSalts []*coredocumentpb.DocumentSalt) *proofs.Salts {
-	if protoSalts == nil {
-		return nil
+		salts = append(salts, &salt)
+		cd.Document.Salts = salts
```
vedhavyas: shouldn't you also update `cd.Document.CoreDocModified` to `true`?

documents/write_acls.go
```diff
@@ -22,15 +22,15 @@ type ChangedField struct {
 // GetChangedFields takes two document trees and returns the compact property, old and new value of the fields that are changed in new tree.
 // Properties may have been added to the new tree or removed from the new tree.
 // In Either case, since the new tree is different from old, that is considered a change.
-func GetChangedFields(oldTree, newTree *proofs.DocumentTree, lengthSuffix string) (changedFields []ChangedField) {
+func GetChangedFields(oldTree, newTree *proofs.DocumentTree) (changedFields []ChangedField) {
```
vedhavyas: this will make the function unusable as a generic call. Lets bring back the suffix, let caller pass the suffix

documents/entity/model.go
```diff
@@ -267,7 +241,7 @@ func (e *Entity) PrepareNewVersion(old documents.Model, data *cliententitypb.Ent
 	}
 
 	oldCD := old.(*Entity).CoreDocument
-	e.CoreDocument, err = oldCD.PrepareNewVersion(collaborators, true, compactPrefix())
+	e.CoreDocument, err = oldCD.PrepareNewVersion(collaborators, compactPrefix())
```
vedhavyas: let's change the `PrepareNewVersion` function definition to 
```
PrepareNewVersion(prefix, collaborators ...)
```

This way its more clear IMO

nft/ethereum_payment_obligation.go
vimukthi-git: existing stuff moved around

nft/ethereum_payment_obligation.go
vimukthi-git: existing stuff moved around

vimukthi-git:
@vedhavyas comments are welcome on my prs even after merging, I will address critical stuff ASAP always.

PR #848, opened 2019-03-14
Fix identity concurrent creation

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/ethereum/transaction_status_task.go
+++ b/ethereum/transaction_status_task.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+
 	"github.com/centrifuge/go-centrifuge/transactions/txv1"
 
 	"github.com/centrifuge/go-centrifuge/errors"
@@ -25,6 +28,13 @@ const (
 
 	// TransactionAccountParam contains the name  of the account
 	TransactionAccountParam string = "Account ID"
+
+	// TransactionEventName contains the name of the event filtered
+	TransactionEventName string = "TxEventName"
+
+	// TransactionEventValueIdx contains the index of the position of the event value
+	TransactionEventValueIdx string = "TxEventValueIdx"
+
 	// TransactionStatusSuccess contains the flag for a successful receipt.status
 	TransactionStatusSuccess uint64 = 1
 
@@ -52,6 +62,10 @@ type TransactionStatusTask struct {
 	//txHash is the id of an Ethereum transaction
 	txHash    string
 	accountID identity.DID
+
+	//event filter
+	eventName     string
+	eventValueIdx int
 }
 
 // NewTransactionStatusTask returns a the struct for the task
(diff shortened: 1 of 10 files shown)


transactions/transaction.go
```diff
@@ -149,6 +160,7 @@ type Manager interface {
 	// ExecuteWithinTX executes the given unit of work within a transaction
 	ExecuteWithinTX(ctx context.Context, accountID identity.DID, existingTxID TxID, desc string, work func(accountID identity.DID, txID TxID, txMan Manager, err chan<- error)) (txID TxID, done chan bool, err error)
 	GetTransaction(accountID identity.DID, id TxID) (*Transaction, error)
+	UpdateTransactionWithValue(accountID identity.DID, id TxID, key string, value []byte) error
```
vimukthi-git: do we need to have this as part of the interface?

mikiquantum: I would like to, as this is the base for allowing the storage of values that are calculated in an ansynchronous task that can be consumed later on.

identity/ideth/factory.go
```diff
@@ -107,12 +109,12 @@ func (s *factory) CreateIdentity(ctx context.Context) (did *id.DID, err error) {
 		return nil, err
 	}
 
-	identityAddress, err := s.CalculateIdentityAddress(ctx)
+	calcIdentityAddress, err := s.CalculateIdentityAddress(ctx)
```
manuelpolzhofer: If we get the `identityAddress` why calculating the address at all?

We could just use  createdAddr?

identity/ideth/factory.go
```diff
@@ -123,13 +125,29 @@ func (s *factory) CreateIdentity(ctx context.Context) (did *id.DID, err error) {
...
+	}
+	createdAddr := common.BytesToAddress(idCreated.Value)
+	log.Infof("ID Created with address: %s", createdAddr.Hex())
 
+	if calcIdentityAddress.Hex() != createdAddr.Hex() {
+		log.Infof("[Recovered] Found race condition creating identity, calculatedDID[%s] vs createdDID[%s]", calcIdentityAddress.Hex(), createdAddr.Hex())
 	}
 
-	err = isIdentityContract(*identityAddress, s.client)
+	err = isIdentityContract(createdAddr, s.client)
```
manuelpolzhofer: We need to update the isIdentityContract method.

Instead of comparing the bytes we should ask the identityFactory if it created the identity.

ethereum/transaction_status_task.go
```diff
@@ -167,13 +224,21 @@ func (tst *TransactionStatusTask) RunTask() (resp interface{}, err error) {
...
+		if err != ErrTransactionFailed {
+			err = gocelery.ErrTaskRetryable
+		}
+		return nil, err
 	}
 
-	if err != ErrTransactionFailed {
-		return nil, gocelery.ErrTaskRetryable
+	if tst.eventName != "" {
+		v, err := tst.getEventValueFromTransactionReceipt(ctx, tst.txHash, tst.eventName, tst.eventValueIdx)
```
manuelpolzhofer: okay, so we always add the events to the centrifuge transactions.

mikiquantum: Not always, only when there an event filter passed to the task.

identity/ideth/factory.go
mikiquantum: We need to pass some identity to the tx manager. And it gives us good visibility when a collision happened.

identity/ideth/factory.go
```diff
@@ -123,13 +125,29 @@ func (s *factory) CreateIdentity(ctx context.Context) (did *id.DID, err error) {
...
+		return nil, err
+	}
+	idCreated, ok := tx.Values[identityCreatedEventName]
+	if !ok {
+		return nil, errors.New("Couldn't find value for %s", identityCreatedEventName)
+	}
+	createdAddr := common.BytesToAddress(idCreated.Value)
+	log.Infof("ID Created with address: %s", createdAddr.Hex())
 
+	if calcIdentityAddress.Hex() != createdAddr.Hex() {
```
manuelpolzhofer: ah okay. We just logging it

mikiquantum:
@manuelpolzhofer changes applied. Any other comments?

transactions/transaction.go
vimukthi-git: I understand, but the tasks are only allowed to update transactions using base task, transactions should not be updated unless within transactions package, because it could cause race conditions and other weird side effects. So we should not expose any update methods on the transaction manager.

Issue #835, opened 2019-03-11
Missing Node Test Cases


lucasvo:
This ticket is to implement two testcases with our [`testworld` p2p simulation frame work](https://github.com/centrifuge/go-centrifuge/tree/develop/testworld). 

Refer `testworld/README.md` for env setup and a small guide on how to run. `testworld/park_test.go` is a good starting point for any other tests you define. Copy it and define a file called `missing_node_test.go` for your tests.

#### Alice creates a document with Frank as a collaborator who doesn't have an identity
Try to share a document with a collaborator that doesn't exist on centrifuge. The node should reject creating the document because the identity can't be found.

#### Alice creates a document with Frank as a collaborator who has an identity but no P2P key
Try to share a document with a collaborator that doesn't exist on centrifuge. The node should anchor a document that lacks the signature of Frank.
 
This issue has a gitcoin bounty attached to it. If you'd like to start working on it, feel free to reach out to me (lucas@centrifuge.io) directly, join our slack channel #bounties on our [slack](
https://centrifuge.io/slack/).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it as part of the Centrifuge fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/centrifuge/go-centrifuge/835/2593).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/centrifuge/go-centrifuge/835/2593).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $47,288.53 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 4 days from now.
Please review their action plans below:


[...]

PR #849, opened 2019-03-14
Missing Node test cases

eswarasai:
### Description
Added test cases for missing node

Fixes: #835

### DIFF
--- a/identity/ideth/service.go
+++ b/identity/ideth/service.go
@@ -292,6 +292,10 @@ func (i service) CurrentP2PKey(did id.DID) (ret string, err error) {
 		return ret, err
 	}
 
+	if len(keys) == 0 {
+		return "", errors.New("missing p2p key")
+	}
+
 	lastKey := keys[len(keys)-1]
 	key, err := i.GetKey(did, lastKey.GetKey())
 	if err != nil {
--- a/testworld/missing_node_test.go
+++ b/testworld/missing_node_test.go
@@ -8,12 +8,15 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/centrifuge/go-centrifuge/anchors"
 	"github.com/centrifuge/go-centrifuge/config"
 	"github.com/centrifuge/go-centrifuge/config/configstore"
 	"github.com/centrifuge/go-centrifuge/contextutil"
 	"github.com/centrifuge/go-centrifuge/identity"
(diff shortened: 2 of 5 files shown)


testworld/missing_node_test.go
```diff
@@ -0,0 +1,33 @@
...
+func TestMissingNode_InvalidIdentity(t *testing.T) {
+	t.Parallel()
+
+	// Hosts
+	alice := doctorFord.getHostTestSuite(t, "Alice")
+
+	// RandomDID
+	randomDID := testingidentity.GenerateRandomDID()
+
+	// alice shares a document with bob and charlie
```
vimukthi-git: fix comment

vimukthi-git:
I just noticed that the second test here returns at point from the node earlier than expected, so the document is not getting anchored it seems. We can still merge the pr if this is the current buggy behaviour of the node. @eswarasai could you investigate why the node returns an error before anchoring the document?

testworld/missing_node_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Alice shares a document with a randomly generated DID with missing P2P Key
+	res := createDocument(alice.httpExpect, alice.id.String(), typeInvoice, http.StatusOK, defaultInvoicePayload([]string{randomDID.String()}))
+	txID := getTransactionID(t, res)
+	status, message := getTransactionStatusAndMessage(alice.httpExpect, alice.id.String(), txID)
+	if status != "failed" {
+		t.Error(message)
+	}
+
+	assert.Contains(t, message, "missing p2p key")
```
vimukthi-git: As explained on DM lets add the correct error string here and also a check for the anchor on chain. 
Thank you 👍

eswarasai:
@vimukthi-git @vedhavyas -- This PR should be good to go now. I've rebased with `develop` again. In case the tests fail, please re-run them by clearing cache in Travis CI. Let me know if there's anything else pending. Thanks :)

cc: @lucasvo

identity/ideth/service.go
```diff
@@ -292,6 +292,10 @@ func (i service) CurrentP2PKey(did id.DID) (ret string, err error) {
 		return ret, err
 	}
 
+	if len(keys) == 0 {
+		return "", errors.New("missing p2p key")
```
vimukthi-git: 👍

PR #862, opened 2019-03-19
Add nft owner check + tx error log

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/nft/ethereum_payment_obligation.go
+++ b/nft/ethereum_payment_obligation.go
@@ -204,6 +204,17 @@ func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID,
 			return
 		}
 
+		// Check if tokenID exists in registry and owner is deposit address
+		owner, err := s.OwnerOf(req.RegistryAddress, tokenID[:])
+		if err != nil {
+			errOut <- err
+			return
+		}
+		if owner.Hex() != req.DepositAddress.Hex() {
+			errOut <- errors.New("Owner for tokenID %s should be %s, instead got %s", tokenID.String(), req.DepositAddress.Hex(), owner.Hex())
+			return
+		}
+
 		log.Infof("Document %s minted successfully within transaction %s", hexutil.Encode(req.DocumentID), utxID)
 
 		errOut <- nil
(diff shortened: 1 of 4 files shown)


nft/ethereum_payment_obligation.go
```diff
@@ -204,6 +204,17 @@ func (s *ethereumPaymentObligation) minter(ctx context.Context, tokenID TokenID,
 			return
 		}
 
+		// Check if tokenID exists in registry and owner is deposit address
+		owner, err := s.OwnerOf(req.RegistryAddress, tokenID[:])
```
manuelpolzhofer: Just out of curiosity.
What happens if the user passes an invalid address as depositAddress?
Would it fail?

transactions/txv1/manager.go
```diff
@@ -92,6 +92,7 @@ func (s *manager) ExecuteWithinTX(ctx context.Context, accountID identity.DID, e
 			if e == nil && transactions.TxIDEqual(existingTxID, transactions.NilTxID()) {
 				tempTx.Status = transactions.Success
 			} else if e != nil {
+				tempTx.Logs = append(tempTx.Logs, transactions.NewLog(desc, e.Error()))
```
vimukthi-git: could you add something here that says log was added by `manager`?

nft/ethereum_payment_obligation.go
mikiquantum: good question, I think there must be a test case in truffle project that should verify that.

rdinicut: What does invalid mean? There is a check for a valid ETH address.

rdinicut: Valid means it passes a checksum test

lucasvo: There are no checksum tests for addresses, the checksum is only in capitalization which is not really enforced outside of user input. Theoretically any 20 byte value is correct.

lucasvo:
Is it possible to add unit tests here for the new code?

nft/ethereum_payment_obligation.go
rdinicut: Yes, but you can also enforce the checksum to have some extra validation. Web3 does that for example

PR #863, opened 2019-03-19
Fix Testworld NFT Test

vimukthi-git:
Issue Link: <Github link>

### DIFF
--- a/anchors/service.go
+++ b/anchors/service.go
@@ -54,7 +54,13 @@ func (s *service) GetAnchorData(anchorID AnchorID) (docRoot DocumentRoot, anchor
 	// Ignoring cancelFunc as code will block until response or timeout is triggered
 	opts, _ := s.client.GetGethCallOpts(false)
 	r, err := s.anchorRepositoryContract.GetAnchorById(opts, anchorID.BigInt())
+	if err != nil {
+		return docRoot, anchoredTime, err
+	}
 	blk, err := s.client.GetEthClient().BlockByNumber(context.Background(), big.NewInt(int64(r.BlockNumber)))
+	if err != nil || blk == nil {
+		return docRoot, anchoredTime, err
+	}
 	return r.DocumentRoot, time.Unix(blk.Time().Int64(), 0), err
 }
 
--- a/anchors/service.go
+++ b/anchors/service.go
@@ -5,9 +5,8 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/centrifuge/go-centrifuge/errors"
-
 	"github.com/centrifuge/go-centrifuge/contextutil"
+	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-centrifuge/ethereum"
 	"github.com/centrifuge/go-centrifuge/identity"
 	"github.com/centrifuge/go-centrifuge/queue"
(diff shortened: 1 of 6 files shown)


testworld/nft_test.go
```diff
@@ -130,10 +132,14 @@ func paymentObligationMint(t *testing.T, documentType string, grantNFTAccess, to
...
-
-	assert.Nil(t, err, "mintNFT should be successful")
+	assert.NoError(t, err, "mintNFT should be successful")
 	assert.True(t, len(response.Value("token_id").String().Raw()) > 0, "successful tokenId should have length 77")
 
+	tokenID, err := nft.TokenIDFromString(response.Value("token_id").String().Raw())
+	assert.NoError(t, err, "token ID should be correct")
+	owner, err := alice.host.tokenRegistry.OwnerOf(registry, tokenID.BigInt().Bytes())
+	assert.NoError(t, err)
+	assert.Equal(t, strings.ToLower("0x44a0579754d6c94e7bb2c26bfa7394311cc50ccb"), strings.ToLower(owner.Hex()))
```
mikiquantum: I think you meant to pass here the deposit address of the test context.

documents/invoice/model.go
```diff
@@ -487,18 +486,3 @@ func (i *Invoice) CollaboratorCanUpdate(updated documents.Model, collaborator id
 	cf := documents.GetChangedFields(oldTree, newTree, proofs.DefaultSaltsLengthSuffix)
 	return documents.ValidateTransitions(rules, cf)
 }
-
```
mikiquantum: Are these defined somewhere else? or are not needed.

anchors/service.go
```diff
@@ -54,7 +54,13 @@ func (s *service) GetAnchorData(anchorID AnchorID) (docRoot DocumentRoot, anchor
 	// Ignoring cancelFunc as code will block until response or timeout is triggered
 	opts, _ := s.client.GetGethCallOpts(false)
 	r, err := s.anchorRepositoryContract.GetAnchorById(opts, anchorID.BigInt())
+	if err != nil {
```
mikiquantum: can r be nil and not error thrown?

testworld/nft_test.go
vimukthi-git: we always deposit to our address to make to debug easily

documents/invoice/model.go
vimukthi-git: they are not needed as @vedhavyas pointed out to me today

testworld/nft_test.go
mikiquantum: even against local envs?

anchors/service.go
vimukthi-git: let me add a check for those

vimukthi-git: r cant be nil but blk can so added a check

testworld/nft_test.go
vimukthi-git: yes, its the same hardcoded address for all

anchors/service.go
```diff
@@ -5,6 +5,8 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/centrifuge/go-centrifuge/errors"
+
```
vedhavyas: sort imports pls

PR #866, opened 2019-03-20
Error handling in identity creation

charleenfei:
Just a little fix for a little bug in error handling when we create an identity. Otherwise, we will attempt to return a nil value and the error message will reflect that rather than the real error.

- also, logs your DID after successful creation of config.yaml


identity/ideth/factory.go
```diff
@@ -49,7 +49,7 @@ func (s *factory) createIdentityTX(opts *bind.TransactOpts) func(accountID id.DI
 		ethTX, err := s.client.SubmitTransactionWithRetries(s.factoryContract.CreateIdentity, opts)
 		if err != nil {
 			errOut <- err
-			log.Infof("Failed to send identity for creation [txHash: %s] : %v", ethTX.Hash(), err)
+			log.Infof("Failed to send identity for creation [txHash: %s] : %v", ethTX, err)
```
vedhavyas: This was raised by trail of bits too.
```suggestion
			log.Infof("Failed to send identity for creation: %v", err)
```

No point printing nil

PR #874, opened 2019-03-21
fix edge case

vedhavyas:
I think I covered all the edge cases. But would be great if you guys could think of a few


documents/decimal.go
```diff
@@ -17,6 +18,9 @@ const (
 	maxDecimalByteLength  = 32
 )
 
+// minDec is the min absolute value we support
+var minDec = decimal.RequireFromString("0.1")
```
vimukthi-git: this can't be a constant?

mikiquantum: questions that arise for me are:
- minimum amount of an invoice or any field in the document is 0.01. Is that good enough?
- forces us to represent % in base100 + decimals, which i think is ok.

vedhavyas: Maybe I missed something. Let me think more on this

documents/decimal.go
```diff
@@ -88,23 +93,29 @@ func (d *Decimal) Bytes() (decimal []byte, err error) {
...
 	sd := strings.Split(s, ".")
 
-	fraction := make([]byte, maxFractionByteLength)
+	v := sd[0]
+	fraction := strings.Repeat("0", decimalPrecision)
 	if len(sd) == 2 {
-		fraction, err = byteutils.IntBytesFromString(sd[1])
-		if err != nil {
-			return nil, err
+		// add extra zeros if precision is not max
```
mikiquantum: Is it possible that the user inputs a `,` instead of `.`? Does the decimal lib support it? Do we we need to fail in that case?

documents/decimal.go
```diff
@@ -117,14 +128,25 @@ func (d *Decimal) Bytes() (decimal []byte, err error) {
 
 // SetBytes parse the bytes to Decimal.
 func (d *Decimal) SetBytes(dec []byte) error {
-	if len(dec) < minDecimalByteLength {
+	if len(dec) < minDecimalByteLength || len(dec) > maxDecimalByteLength {
```
mikiquantum: if the decimal length is less than the minimum, cant we append 0s or we want to enforce always the minimum?

documents/decimal.go
vedhavyas: yeah, we will fail then. We decided to support only `.`

documents/decimal.go
vedhavyas: we want to enforce the minimum always.

PR #887, opened 2019-03-25
client API: add read/write rules

### LINKED ISSUE #873: Add new logic to write
Mainly convert from client api format to coredoc format. Give write access `prefix.*`

Write access:
- create write roles/rules/transitions
- create read roles/rules

charleenfei:
Closes #873


documents/coredocument.go
```diff
@@ -197,6 +202,47 @@ func (cd *CoreDocument) PrepareNewVersion(collaborators []string, documentPrefix
 
 }
 
+// PrepareNewVersion prepares the next version of the CoreDocument
+// if initSalts is true, salts will be generated for new version.
```
mikiquantum: Lets add a TODO that will replace the original `PrepareNewVersion ` very soon.

documents/coredocument.go
```diff
@@ -197,6 +202,47 @@ func (cd *CoreDocument) PrepareNewVersion(collaborators []string, documentPrefix
 
 }
 
+// PrepareNewVersion prepares the next version of the CoreDocument
```
mikiquantum: Let's keep the linting happy though:
```suggestion
// PrepareNewVersion1 prepares the next version of the CoreDocument
```

documents/coredocument.go
```diff
@@ -76,6 +76,11 @@ type CoreDocument struct {
 	Document coredocumentpb.CoreDocument
 }
 
+type Collaborators struct {
```
mikiquantum:
```suggestion
type CollaboratorsAccess struct {
```
At the moment is very simple as we do not care about field access and that is ok.

documents/coredocument.go
```diff
@@ -197,6 +202,47 @@ func (cd *CoreDocument) PrepareNewVersion(collaborators []string, documentPrefix
...
+		SignatureData:      new(coredocumentpb.SignatureData),
+	}
+
+	err = populateVersions(&cdp, &cd.Document)
+	if err != nil {
+		return nil, err
+	}
+
+	ncd := &CoreDocument{Document: cdp}
+	ncd.addCollaboratorsToReadSignRules(rcs)
```
mikiquantum: Unless `ncd.addCollaboratorsToTransitionRules` as well adds the read rules needed, we need to make this call as well with the write collabs:
`ncd.addCollaboratorsToReadSignRules(wcs)`

documents/coredocument.go
charleenfei: :+1:

documents/coredocument.go
```diff
@@ -76,6 +76,11 @@ type CoreDocument struct {
 	Document coredocumentpb.CoreDocument
 }
 
+type CollaboratorsAccess struct {
```
vedhavyas: Why do we need this?
Can you comment this pls

charleenfei: we need this to differentiate the types of collaborators we will be adding to the rules, when we integrate the request header as part of the ClientAPI improvements epic.

vedhavyas: Then, shouldn't this be part of the client protobufs instead?

charleenfei: The structures exist in the client protobufs as well, under the message RequestHeader. but we need this structure here in the coredoc to convert from api wire format on the create/update calls to a usable payload for coredoc methods.

vedhavyas: I think we should'nt create these but instead re-use them. Else there will be converters for just this. And these converters will get tricky once we add the more fiels to the write_acls

documents/coredocument.go
```diff
@@ -457,9 +504,82 @@ func (cd *CoreDocument) GetCollaborators(filterIDs ...identity.DID) ([]identity.
...
+	}
+
+	return collabs, nil
+}
+
+// TODO: remove this method with new GetCollaborators
 // getCollaborators returns all the collaborators who belongs to the actions passed.
 func (cd *CoreDocument) getCollaborators(actions ...coredocumentpb.Action) (ids []identity.DID, err error) {
-	findRole(cd.Document, func(_, _ int, role *coredocumentpb.Role) bool {
+	findReadRole(cd.Document, func(_, _ int, role *coredocumentpb.Role) bool {
```
vedhavyas: this shouldn't return a bool but instead the err itself. Once you return the error, we should return the error to the caller else we will just swallow the error

documents/coredocument.go
```diff
@@ -457,9 +504,82 @@ func (cd *CoreDocument) GetCollaborators(filterIDs ...identity.DID) ([]identity.
...
+			if err != nil {
+				return false
+			}
+			ids = append(ids, did)
+		}
+
+		return false
+	}, actions...)
+
+	if err != nil {
```
vedhavyas: same as above

documents/coredocument.go
```diff
@@ -457,9 +504,82 @@ func (cd *CoreDocument) GetCollaborators(filterIDs ...identity.DID) ([]identity.
...
+		return nil, err
+	}
+
+	return ids, nil
+}
+
+// getWriteCollaborators returns all the collaborators which have access to the transition actions passed in.
+func (cd *CoreDocument) getWriteCollaborators(actions ...coredocumentpb.TransitionAction) (ids []identity.DID, err error) {
+	findTransitionRole(cd.Document, func(_, _ int, role *coredocumentpb.Role) bool {
```
vedhavyas: same as above

documents/coredocument.go
mikiquantum: I think we should not expose clientapi definitions into internal methods. This way we can evolve the clientapi independent of the core logic and viceversa.

documents/coredocument.go
```diff
@@ -457,9 +504,82 @@ func (cd *CoreDocument) GetCollaborators(filterIDs ...identity.DID) ([]identity.
...
+
+	rcs, err := cd.getReadCollaborators(coredocumentpb.Action_ACTION_READ_SIGN, coredocumentpb.Action_ACTION_READ)
+	if err != nil {
+		return CollaboratorsAccess{}, err
+	}
+	wcs, err := cd.getWriteCollaborators(coredocumentpb.TransitionAction_TRANSITION_ACTION_EDIT)
+	if err != nil {
+		return CollaboratorsAccess{}, err
+	}
+	rc := filterCollaborators(rcs, filterIDs...)
```
vimukthi-git: these filter calls can be easily avoided by passing a filtered ids to the `getReadCollaborators` and  `getWriteCollaborators`.

charleenfei: those methods don't take filtered ids as a parameter as of now... are you thinking of GetCollaborators?

vimukthi-git: I mean you can modify them to take it as params, I don't know the complexity of changing them, but looks pretty straight forward here?

p2p/receiver/handler.go
```diff
@@ -247,17 +247,17 @@ func (srv *Handler) validateDocumentAccess(ctx context.Context, docReq *p2ppb.Ge
 	switch docReq.AccessType {
 	case p2ppb.AccessType_ACCESS_TYPE_REQUESTER_VERIFICATION:
 		if !m.AccountCanRead(peer) {
-			return errors.New("requester does not have access")
+			return ErrAccessDenied
```
vimukthi-git: 💯

Issue #781, opened 2019-02-26
Remove roots from the Core Document.


vedhavyas:
Remove roots - DataRoot, Sigining Root, and Document Root, from the CoreDocument as we will be calculating them everytime

lucasvo:
Take a look at https://github.com/centrifuge/go-centrifuge/pull/797 it would be good to build this on top of this PR. There will be some conflict there.

PR #890, opened 2019-03-25
Remove roots 

vimukthi-git:
Issue Link: #781


documents/coredocument.go
```diff
@@ -249,7 +239,7 @@ func (cd *CoreDocument) CreateProofs(docType string, dataTree *proofs.DocumentTr
 	if err != nil {
 		return nil, errors.New("failed to generate core document tree: %v", err)
 	}
-	srHash, err := cd.GetSigningRootHash()
+	srHash, err := cd.GetSigningRootProofHash(docType, dataTree.RootHash())
```
vedhavyas: can we rename this to `CalculateSigningRoot`? The should be consistent with remaining root calculations

vedhavyas: Actually, we dont need this. We need to replace this function call with `CalculateSingatureRoot`.

documents/model.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/centrifuge/precise-proofs/proofs"
+
```
vedhavyas: import spacing

documents/coredocument.go
vimukthi-git: it just leaks info about the tree to outside, I would keep it the way it is, so that we can change/evolve the implementations as we need.

Issue #713, opened 2019-02-02
Add Attribute struct with support for key hashing


pstehlik:
Adding custom fields (arbitrary field-value pairs for functional documents).

Field definitions and suggested structure for custom fields are described in https://hackmd.io/yJvky6omQu2yOU4FHm3B9w?both

# Testing
Custom fields tests for all possible field values:
* string
* decimal/int
* binary
* boolean

Test with small and large-ish binaries in Testworld.

# Dependencies and affected external components
P2P Message modifications will lead to Client API changes.

PR #901, opened 2019-03-27
Adding attributes to models

vimukthi-git:
Issue Link: #713


documents/model.go
```diff
@@ -109,6 +109,12 @@ type Model interface {
 
 	// CollaboratorCanUpdate returns an error if indicated identity does not have the capacity to update the document.
 	CollaboratorCanUpdate(updated Model, collaborator identity.DID) error
+
+	// AddAttribute adds a custom attribute to the model with the given value. If an attribute with the given name already exists its updated.
```
charleenfei: typo: already exists, it is updated

documents/coredocument_test.go
```diff
@@ -488,3 +488,21 @@ func TestCoreDocument_GetSignCollaborators(t *testing.T) {
...
 }
+
+func TestCoreDocument_AddAttribute(t *testing.T) {
+	id1 := testingidentity.GenerateRandomDID()
+	cas := CollaboratorsAccess{ReadWriteCollaborators: []identity.DID{id1}}
+	cd, err := NewCoreDocumentWithCollaborators(nil, cas)
+	assert.NoError(t, err)
+	err = cd.AddAttribute("com.basf.deliverynote.chemicalnumber", Str, "100")
+	assert.NoError(t, err)
+	assert.True(t, len(cd.Attributes) == 1)
```
charleenfei: assert.Len?

documents/coredocument.go
```diff
@@ -587,6 +590,30 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
 	return utils.FromTimestamp(cd.Document.Timestamp)
 }
 
+// AddAttribute adds a custom attribute to the model with the given value. If an attribute with the given name already exists its updated.
```
charleenfei: typo: already exists, it's/ it is updated

documents/attribute.go
```diff
@@ -0,0 +1,92 @@
...
+type AllowedAttributeType string
+
+// String string repr
+func (a AllowedAttributeType) String() string {
+	return string(a)
+}
+
+const (
+	// Int256 is the standard integer custom attribute type
+	Int256 AllowedAttributeType = "int256"
```
mikiquantum: Can we use suffix/prefix `Type`? For example: `Int256Type` or `TypeInt256`

documents/coredocument_test.go
charleenfei: could we also test cases with empty readableKey or nil value?

crypto/generate.go
```diff
@@ -39,11 +39,17 @@ func GenerateSigningKeyPair(publicFileName, privateFileName, curveType string) (
...
 	h := sha256.New()
-	_, err = h.Write(preimage)
+	_, err = h.Write(value)
 	if err != nil {
-		return []byte{}, []byte{}, err
+		return []byte{}, err
 	}
-	hash = h.Sum(hash)
-	return preimage, hash, nil
+	hash = h.Sum(nil)
```
charleenfei: why is this line necessary if we are just appending nil?

documents/coredocument_test.go
vimukthi-git: As the `TODO` says below these would be updated when converters are ready. Right now its just a sanity check.

crypto/generate.go
vimukthi-git: yes we can just make the following line
`return h.Sum(nil), nil`

documents/coredocument.go
```diff
@@ -587,6 +590,30 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
...
+
+// GetAttribute gets the attribute with the given name from the model together with its type, it returns a non-nil error if the attribute doesn't exist or can't be retrieved.
+func (cd *CoreDocument) GetAttribute(name string) (hashedKey []byte, attrType string, value interface{}, valueStr string, err error) {
+	if attr, ok := cd.Attributes[name]; ok {
+		// TODO convert value to its string repr
+		return attr.hashedKey, string(attr.attrType), attr.value, "", nil
+	}
+	return hashedKey, attrType, value, valueStr, errors.NewTypedError(ErrCDAttribute, errors.New("attribute does not exist"))
+}
+
```
mikiquantum: Should we add a deleteAttribute?

crypto/generate.go
charleenfei: I mean, we can also just return the hash, no? Sum() does not change the underlying hash state, just appends the current hash to the parameter and returns.

documents/coredocument_test.go
charleenfei: 👍

crypto/generate.go
vimukthi-git: As far as I know there is no other function to get the underlying hash anyway, so we have to call this.

vimukthi-git:
@mikiquantum @charleenfei thanks for the reviews, I will address your concerns ASAP.
After this is merged we will park the feature temporarily for Amphitrite release. Since @lucasvo wants me to focus on other more critical stuff.

crypto/generate.go
charleenfei: okay, I see

documents/coredocument.go
```diff
@@ -587,6 +590,36 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
...
+// DeleteAttribute deletes a custom attribute from the model
+func (cd *CoreDocument) DeleteAttribute(name string) error {
```
vedhavyas: Any mutations to CoreDocument should return the new model. This will ensure that we dont mutate the current one

PR #903, opened 2019-03-28
Mainnet fixes

vimukthi-git:
## List of issues faced
1. Config _ issue, sent bunch of TXs to 0x0000..
2. Transactions timing out
3. Found that 2 might be because number of queue retries are too small
4. Adjusted retries
5. Then the slower transactions were blocking the pool and timing out entire test
6. Then adjusted gas price, didn't work because `gas price * gas limit > account balance`, geth rejects it.
7. Gas limit too low for NFT minting, had adjust to 1,000,000
7. Found that geth client can automatically adjust gas price(oracle), finally NFT went through with this.

cc @pstehlik


build/configs/default_config.yaml
```diff
@@ -65,9 +65,24 @@ networks:
...
-      anchorRepository: "0xff5f35f6f3910ed66a24b6e55542e66b868416a3"
-      paymentObligation: "0x48c25e7639e888e667f20d30b96653de054251d8"
+      identityFactory: "0xF86177b275E4Fcb2Ab5A3FcF5E46aab51f73D7a6"
+      anchorRepository: "0xf630d8bE5664C7FB1234CEf48b6A729cEe3B71D8"
+      paymentObligation: "0x9eee8FbE3E29c96cC944F0eeC1594fe1b633dABe"
+
+  # Main net test network
+  embarcaderotest:
+    # Numeric ID of the Centrifuge network
+    id: 54
```
mikiquantum: Can we use here number 1 as well? Or reserve a decimal like 10?

build/scripts/tests/run_testworld.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Running Testworld"
 
-cleanup="ls testworld/hostconfigs/* | grep testworld | grep -v README.md | tr -d : | xargs rm -rf"
+cleanup="pkill -f testworld; ls testworld/hostconfigs/* | grep testworld | grep -v README.md | tr -d : | xargs rm -rf"
```
mikiquantum: What is the purpose of `pkill` here? Just kill a previous process if it was running? Or something else?

build/configs/default_config.yaml
vimukthi-git: sure

build/scripts/tests/run_testworld.sh
vimukthi-git: it will kill everything testworld related on the previous run.

vimukthi-git: haha including it self it seems :D

PR #904, opened 2019-03-28
Feat/write acl api

### LINKED ISSUE #869: Integrate Write Access conversion
Add a method to the CoreDocument model (model/coredocument.go) to read values in CoreDocument and use them as input for the init rules/transitions...
This will break functional tests, which they will need to be updated as part of this work.

vedhavyas:
Closes #869


documents/coredocument.go
```diff
@@ -486,70 +434,33 @@ func (cd *CoreDocument) coredocTree(docType string) (tree *proofs.DocumentTree,
...
 	wcs, err := cd.getWriteCollaborators(coredocumentpb.TransitionAction_TRANSITION_ACTION_EDIT)
 	if err != nil {
 		return CollaboratorsAccess{}, err
 	}
+
 	rc := filterCollaborators(rcs, filterIDs...)
 	wc := filterCollaborators(wcs, filterIDs...)
-
-	collabs := CollaboratorsAccess{
+	rc = filterCollaborators(rc, wc...)
```
vimukthi-git: switch the order of filtering rc and wc then you can avoid the 3rd filter here

documents/converters.go
```diff
@@ -271,3 +272,87 @@ func FromP2PPaymentDetails(pdetails []*commonpb.PaymentDetails) ([]*PaymentDetai
...
+			if _, ok := wmap[c]; ok {
+				continue
+			}
+
+			wmap[c] = struct{}{}
+			wcs = append(wcs, c)
+		}
+	}
+
+	if racess != nil {
```
vimukthi-git: same logic as `waccess` so repeated code, Refactor to a function.

protobufs/invoice/service.proto
```diff
@@ -60,30 +60,23 @@ message GetVersionRequest {
...
-  repeated string collaborators = 2;
-  InvoiceData data = 3;
+  document.ReadAccess read_access = 2;
+  document.WriteAccess write_access = 3;
+  InvoiceData data = 4;
 }
 
 message InvoiceResponse {
-  ResponseHeader header = 1;
+  document.ResponseHeader header = 1;
```
vimukthi-git: 👍

documents/converters.go
vedhavyas: not the same logic ;)

PR #905, opened 2019-03-29
Entity Relationship Create (Share) service

### LINKED ISSUE #894: Entity Relationship Service/Create

Create Services EntityRelationship for a given Entity Object

charleenfei:
Closes #894


documents/entityrelationship/model.go
```diff
@@ -215,12 +215,12 @@ func (e *EntityRelationship) CreateProofs(fields []string) (proofs []*proofspb.P
...
 }
 
-// DocumentType returns the entity relationship document type. THIS IS NOT IMPLEMENTED FOR ENTITY RELATIONSHIP.
+// DocumentType returns the entity relationship document type.
 func (*EntityRelationship) DocumentType() string {
 	return documenttypes.EntityRelationshipDocumentTypeUrl
 }
 
-// AddNFT adds NFT to the EntityRelationship.
+// AddNFT adds NFT to the EntityRelationship. THIS IS NOT IMPLEMENTED FOR ENTITY RELATIONSHIP.
```
vimukthi-git: I didn't notice this before, 
Please don't panic on production code specially since this something someone can try by calling nft endpoint, just log and return an error. 
Only places where we call panic is when the node shouldn't be alive afterwards such as when there is a config error.

vedhavyas:
```suggestion
// AddNFT is not implemented for EntityRelationship.
```

vimukthi-git: Panic is still there right?

charleenfei:
> Lets bring the test coverage to more than 80%
> 
> Minor suggestions. LGTM

I've added some more coverage, but I think the TODO services (update) are throwing off code coverage also. When I add these in the next PR and complete the EntityRelationship service, I'll bring up test coverage to over 80%.

PR #916, opened 2019-04-01
take password interactively

vedhavyas:


mikiquantum:
@vedhavyas what happened to codecov here?

vedhavyas:
@mikiquantum previous develop build didn't run. Running it now. Lets wait and see

cmd/centrifuge/create_config.go
```diff
@@ -30,11 +32,21 @@ func init() {
...
-			err := cmd.CreateConfig(targetDataDir, ethNodeURL, accountKeyPath, accountPassword, network, apiPort, p2pPort, bootstraps, txPoolAccess, false, "", nil)
+			_, err := fmt.Fprintln(os.Stderr, "Enter your Ethereum Account Password:")
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			pwd, err := terminal.ReadPassword(int(syscall.Stdin))
+			if err != nil {
+				// lets take empty password
+				log.Error(err)
```
vimukthi-git: what happens to the value of password here if there is a different error than empty password?

vedhavyas: its not empty password issue here. error happens when its not a terminal per se. like our scripts running the tests on travis etc...

vimukthi-git: but then `string(pwd)` will fail right? because `pwd` is `nil`

PR #928, opened 2019-04-02
Remove certs

### LINKED ISSUE #909: Remove certificates for HTTP and listen on local interface only

vedhavyas:
Closes #909 

Changes:
- Removed the certificates
- GRPC server will be run on a different port and `127.0.0.1`
- API host is `127.0.0.1` by default but can be changed through `createconfig`


testworld/park.go
```diff
@@ -195,16 +195,17 @@ func (r *hostManager) startTempHost(name string) error {
...
 		name,
 		r.ethNodeUrl,
 		r.accountKeyPath,
 		r.accountPassword,
 		r.network,
+		"127.0.0.1",
 		twConfigName,
 		p2pTimeout,
-		apiPort, p2pPort, bootstraps,
+		apiPort, p2pPort, grpcPort, bootstraps,
```
vimukthi-git: I usually don't care much about whats here but let's not have this on testworld hosts, it just creates unnecessary dependancies. Specially since we will remove grpc soon.

build/configs/default_config.yaml
```diff
@@ -99,10 +99,13 @@ accounts:
   keystore: /tmp/accounts
 
 # Interface where the API and P2P Server listens to
-nodeHostname: 0.0.0.0
+nodeHostname: 127.0.0.1
 # Port where API Server listens to
 nodePort: 8082
 
+# Port where GRPC server listens to
+grpcPort: 28202
```
vimukthi-git: why do we need this in the config?
Does a client need to configure the grpc server for any reason? If so its not desirable since we are exposing something that we are going to remove soon.

testworld/park.go
vedhavyas: I dont understand what you mean by lets not have it as part of testworld. We are seperating the HTTP api from GRPC. So wee need to have a seperate port for GRPC. I'm open any solutions you have

build/configs/default_config.yaml
vedhavyas: We already were exposing GRPC thriugh HTTP API. That means, we were exposing the GRPC to the public. This PR is changes that and exposes GRPC server on localhost. And we cannot assume the GRPC port we chose to be avaible during the start of the cent-node. So, we will have to expose it here

utils/httputils.go
```diff
@@ -17,3 +20,15 @@ func SendPOSTRequest(url string, contentType string, payload []byte) (statusCode
 
 	return resp.StatusCode(), nil
 }
+
+// GetFreeAddrPort returns a loopback address and port that can be listened from.
+// Note: port is included in the address.
+func GetFreeAddrPort() (string, int, error) {
+	l, err := net.Listen("tcp", "localhost:0")
```
vimukthi-git: Cool, is this like a standard way to find a free port?

vedhavyas: yeap. Re-usable if required

PR #931, opened 2019-04-03
create file with 600 permission

### LINKED ISSUE #908: Generated private keys are world-readable by default
Make sure the file permissions in the CreateConfig command are 600

vedhavyas:
Closes #908

### DIFF
--- a/utils/io.go
+++ b/utils/io.go
@@ -10,10 +10,13 @@ import (
 
 // WriteKeyToPemFile writes encode of key and purpose to the file
 func WriteKeyToPemFile(fileName string, keyPurpose string, key []byte) error {
-	f, err := os.Create(fileName)
+	// we are going with 0622 so that when umask is applied, we will get 0600 file permission
+	f, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0622)
 	if err != nil {
 		return err
 	}
+
+	defer f.Close()
 
 	block := &pem.Block{
 		Type:  keyPurpose,
@@ -22,7 +25,7 @@ func WriteKeyToPemFile(fileName string, keyPurpose string, key []byte) error {
 	if err := pem.Encode(f, block); err != nil {
 		return err
 	}
-	f.Close()
+
 	return nil
 }
 
--- a/utils/io.go
+++ b/utils/io.go
@@ -9,13 +9,20 @@ import (
 )
 
 // WriteKeyToPemFile writes encode of key and purpose to the file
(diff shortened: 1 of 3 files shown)


utils/io.go
```diff
@@ -10,19 +10,22 @@ import (
 
 // WriteKeyToPemFile writes encode of key and purpose to the file
 func WriteKeyToPemFile(fileName string, keyPurpose string, key []byte) error {
```
mikiquantum: Can we extend the existing `io_test.go` to check that the file permissions are the expected ones?

utils/io.go
```diff
@@ -10,19 +10,22 @@ import (
...
-	f, err := os.Create(fileName)
+	// we are going with 0622 so that when umask is applied, we will get 0600 file permission
+	f, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0622)
```
mikiquantum: Does it work the same across OSs?

vedhavyas: There is no clear way of confirming this. The default Umask is 022 for most OS's. If someone change the umask to be more strict, then we will have problems with current(0644) or old (0666) as well. So I guess, it is okay

mikiquantum: wouldnt it be safer to create file first and then change the permissions? That way we overcome possible restrictive values of Umask?

vedhavyas: I think it should work

utils/io.go
```diff
@@ -15,15 +15,16 @@ func WriteKeyToPemFile(fileName string, keyPurpose string, key []byte) error {
 		return err
 	}
 
+	defer f.Close()
```
vimukthi-git: is there no way to handle error here?

utils/io.go
```diff
@@ -15,15 +15,16 @@ func WriteKeyToPemFile(fileName string, keyPurpose string, key []byte) error {
...
 		Type:  keyPurpose,
 		Bytes: key,
 	}
 	if err := pem.Encode(f, block); err != nil {
 		return err
 	}
-	f.Close()
-	return nil
+
+	return f.Chmod(os.FileMode(0600))
```
vimukthi-git: is there no standard constant defined for this?

vedhavyas: There will be million combinations if they go with constants ;)

vimukthi-git: haha but there are bit masks that can be used right?

vimukthi-git: not critical ;)

vedhavyas: If you mean by using the standard perm 0777 and masking it. I think we will lose readability there. Would prefer keeping it as `0600`

PR #932, opened 2019-04-03
Allow setting Ethereum max gas price

vimukthi-git:
Issue Link: <Github link>


ethereum/geth_client.go
```diff
@@ -180,23 +175,41 @@ func (gc *gethClient) TransactionReceipt(ctx context.Context, txHash common.Hash
...
+		return nil, errors.NewTypedError(ErrEthTransaction, errors.New("failed to create new transaction opts: %v", err))
+	}
 	opts.GasLimit = gc.config.GetEthereumGasLimit()
-	opts.Context = context.Background()
+	opts.Context = ctx
 	return opts, nil
 }
 
+// getOptimalGasPrice get the optimal current gas price from eth client
+func (gc *gethClient) getOptimalGasPrice(ctx context.Context) (*big.Int, error) {
```
vimukthi-git: main change

ethereum/geth_client_integration_test.go
```diff
@@ -93,8 +93,22 @@ func TestGetConnection_returnsSameConnection(t *testing.T) {
...
+	opts, err := gc.GetTxOpts(context.Background(), "main")
+	assert.NoError(t, err)
+	assert.True(t, opts.GasPrice.Cmp(big.NewInt(20000000000)) == 0)
+
+	cfg.Set("ethereum.maxGasPrice", 10000000000)
+	gc, err = ethereum.NewGethClient(cfg)
+	assert.NoError(t, err)
+	assert.NotNil(t, gc)
+	opts, err = gc.GetTxOpts(context.Background(), "main")
+	if assert.Error(t, err) {
```
mikiquantum: we want to make sure that an error is triggered in this scenario, can we force it, instead of having a if clause?

config/configuration.go
```diff
@@ -320,9 +320,14 @@ func (c *configuration) GetEthereumMaxRetries() int {
...
 
-// GetEthereumGasPrice returns the gas price to use for a ethereum transaction.
-func (c *configuration) GetEthereumGasPrice() *big.Int {
-	return big.NewInt(cast.ToInt64(c.get("ethereum.gasPrice")))
+// GetEthereumMaxGasPrice returns the gas price to use for a ethereum transaction.
+func (c *configuration) GetEthereumMaxGasPrice() *big.Int {
+	n := new(big.Int)
+	n, ok := n.SetString(c.GetString("ethereum.maxGasPrice"), 10)
+	if !ok {
+		log.Error("could not read ethereum.maxGasPrice")
```
mikiquantum: Is 0 the default value if not set in config?

ethereum/geth_client_integration_test.go
vimukthi-git: it is forced in this case. its the recommended way to assert an error :)

config/configuration.go
vimukthi-git: its just nil, so it will cause a nil pointer error and node will crash. I guess thats better than continuing to run since its misconfigured.

vimukthi-git: I think I will make this a log.Fatal 
what do you think?

vimukthi-git:
Tested on mainnet :)

PR #934, opened 2019-04-03
entity: GET method

manuelpolzhofer:
- added logic for the two cases
  - collaborator true
      -  return document
  - collaborator false
      - request document with a entityRelationship via p2p


documents/entity/service.go
```diff
@@ -220,3 +220,65 @@ func (s service) DeriveFromUpdatePayload(ctx context.Context, payload *clientent
 
 	return entity, nil
 }
+
+func (s service) GetVersion(ctx context.Context, documentID []byte, version []byte) (documents.Model, error) {
+	return s.get(ctx, documentID, version)
+}
+
+// isDIDaCollaborator returns true if the did is a collaborator of the document
+func (s service) isDIDaCollaborator(ctx context.Context, did identity.DID, model documents.Model) (bool, error) {
```
vedhavyas:
```suggestion
func (s service) isDIDCollaborator(ctx context.Context, did identity.DID, model documents.Model) (bool, error) {
```

documents/entity/service.go
```diff
@@ -220,3 +220,65 @@ func (s service) DeriveFromUpdatePayload(ctx context.Context, payload *clientent
...
+	if err != nil {
+		return false, err
+	}
+
+	for _, d := range collAccess.ReadWriteCollaborators {
+		if d == did {
+			return true, nil
+		}
+	}
+	for _, d := range collAccess.ReadCollaborators {
```
vedhavyas: lets add this logic as part of the Model. Can be re-used
```
type Model interface {
	...
	IsCollaborator(did) (bool(true or false))
	...
}
```

documents/entity/service_test.go
```diff
@@ -5,6 +5,8 @@ package entity
 import (
 	"testing"
 
+	"github.com/centrifuge/go-centrifuge/identity"
+
```
vedhavyas: spacing

documents/entity/service.go
```diff
@@ -220,3 +220,65 @@ func (s service) DeriveFromUpdatePayload(ctx context.Context, payload *clientent
...
+func (s service) get(ctx context.Context, documentID, version []byte) (documents.Model, error) {
+	selfDID, err := contextutil.AccountDID(ctx)
+	if err != nil {
+		return nil, errors.NewTypedError(documents.ErrDocumentConfigAccountID, err)
+	}
+
+	isCollaborator := false
+	var entity documents.Model
+
+	if s.Service.Exists(ctx, documentID) {
```
vedhavyas: You dont need to do this check, `GetCurrentVersion` and `GetVersion` should error out if the document is not found.

documents/entity/service.go
```diff
@@ -220,3 +220,65 @@ func (s service) DeriveFromUpdatePayload(ctx context.Context, payload *clientent
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	if isCollaborator {
+		// todo add relationship array
+		return entity, nil
+	}
+
+	// todo call entityRelationship service and request Entity document from other collaborators
```
vedhavyas: These todos will be in a different PR ?

documents/entity/service.go
vedhavyas: We can move this function to model itself

documents/entity/service.go
manuelpolzhofer: yes.
Waiting for Charlies PR

documents/entity/service.go
manuelpolzhofer: I know but the common case for an entity document is that the document doesn't exist. 

I don't like the pattern

```
 if err != nil {
 //continue with business logic
}
```

documents/entity/service.go
manuelpolzhofer: First I had it there. Afterwards, I realized I would need to add it to the interface and for me, it is more a helper function.

I agree the method should be re-usable for all documents.

manuelpolzhofer:
@vedhavyas  I addressed you changes request. I will merge

PR #937, opened 2019-04-03
Add multiple gas limits

vimukthi-git:
Issue Link: <Github link>


build/configs/default_config.yaml
```diff
@@ -130,8 +130,14 @@ ethereum:
   # maximum allowed gas price to pay for a transaction in wei
   # 20 Gwei now
   maxGasPrice: 20000000000
-  # Default gas limit
-  gasLimit: 1000000
+  # Gas limits for different Ethereum operations
+  gasLimits:
+    idCreate: 1000000
+    idAddKey: 300000
+    idRevocKey: 55000
```
mikiquantum: idRevokeKey

config/configstore/model.go
```diff
@@ -323,7 +323,7 @@ func (nc *NodeConfig) CreateProtobuf() *configpb.ConfigData {
 		EthContextWaitTimeout:     &duration.Duration{Seconds: int64(nc.EthereumContextWaitTimeout.Seconds())},
 		EthIntervalRetry:          &duration.Duration{Seconds: int64(nc.EthereumIntervalRetry.Seconds())},
 		EthGasPrice:               nc.EthereumMaxGasPrice.Uint64(),
-		EthGasLimit:               nc.EthereumGasLimit,
+		EthGasLimit:               0,
```
mikiquantum: Aren't we removing this field?

config/configuration.go
```diff
@@ -53,13 +56,36 @@ const (
...
 	InvoiceUnpaidNFT ContractName = "invoiceUnpaid"
+
+	// IDCreate identity creation operation
+	IDCreate ContractOp = "idCreate"
+
+	// IDAddKey identity add key operation
+	IDAddKey ContractOp = "idAddKey"
+
+	// IDRevocKey identity key revocation operation
+	IDRevocKey ContractOp = "idRevocKey"
```
mikiquantum:
```suggestion
	IDRevokeKey ContractOp = "idRevokeKey"
```

ethereum/geth_client.go
```diff
@@ -107,9 +104,7 @@ func NewGethClient(config Config) (Client, error) {
 		client:    ethclient.NewClient(c),
 		rpcClient: c,
 		host:      u,
-		accounts:  make(map[string]*bind.TransactOpts),
```
mikiquantum: Why are we removing this logic? It saves ~15secs on every tx prep.

config/configstore/model.go
vimukthi-git: next pr :)

ethereum/geth_client.go
vimukthi-git: 1. We shouldn't have cached this since it contains all latest transaction level details including gasPrice
2. Waiting 15 sec is better than not getting the tx throught at all :) 
3. Cache creates race conditions even with the lock since opts are being modified in multiple places now (the reason that made me look at this, but I thought why optimal gas price was static all the time when testing with mainnet before, this was the reason)

vimukthi-git:
@mikiquantum I will ad an ignore
But as explained we must not cache the opts.

ethereum/geth_client.go
mikiquantum: I disagree on the removal of this. We can adapt the gasLimit but not the transactor per request. This cache is only per account and since the field that will be updated is the gasLimit then if we make a copy of the opts per request we do not have that potential problem.
If you want we can discuss this offline

vimukthi-git: As agreed I will keep the cache while adding the other info ondemand on a copy.

ethereum/geth_client.go
```diff
@@ -136,21 +135,37 @@ func (gc *gethClient) defaultReadContext() (ctx context.Context, cancelFunc cont
...
 		return nil, err
 	}
 
 	gc.accounts[accountName] = txOpts
-	return txOpts, nil
+	return gc.copyOpts(context.Background(), txOpts)
+}
+
+// copyOpts copies tx opts each time to avoid any race conditions when modifying, for example gasLimit
+func (gc *gethClient) copyOpts(ctx context.Context, original *bind.TransactOpts) (nOpts *bind.TransactOpts, err error) {
```
mikiquantum: Thanks!

PR #940, opened 2019-04-04
Entity handler: Share and Revoke

### LINKED ISSUE #933: Entity Relationship Service/Update (Revoke)

service supporting the Revoke endpoint

charleenfei:
Closes #933


documents/entityrelationship/repository.go
```diff
@@ -29,14 +32,63 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
...
+		return nil, documents.ErrDocumentNotFound
+	}
+
+	var all []EntityRelationship
+	for _, r := range relationships {
+		e, ok := r.(*EntityRelationship)
+		if !ok {
+			continue
+		}
+		if bytes.Equal(e.EntityIdentifier, entityIdentifier) {
```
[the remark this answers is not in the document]
charleenfei: db shouldn't need to know about this, I will revise above method as discussed.

documents/entityrelationship/service.go
```diff
@@ -190,33 +190,93 @@ func (s service) DeriveEntityRelationshipResponse(model documents.Model) (*entit
...
+	}
+
+	relevant, err := s.repo.ListAllRelationships(entityID, selfDID)
+	if err != nil {
+		return nil, err
+	}
+
+	if version != nil {
+		for _, v := range relevant {
+			r, err := s.GetVersion(ctx, v, version)
```
manuelpolzhofer: Let say I have 3 ER.
Every ER would have a different `current_version` identifier. GetVersion would return the document where `current_version == version`.  It would only work for one ER but would fail for the two others.

We currently don't have a method where you could say return version number 1 or 2.
Do we need the version part at all in the method?

charleenfei: The db method ListAllEntityRelationships lists unique relationships, so this shouldn't happen.

charleenfei: let's discuss this

documents/entityrelationship/repository.go
```diff
@@ -29,14 +31,58 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
...
+			continue
+		}
+		_, found := all[string(e.Document.DocumentIdentifier)]
+		if bytes.Equal(e.EntityIdentifier, entityIdentifier) && !found {
+			all[string(e.Document.DocumentIdentifier)] = e.Document.DocumentIdentifier
+		}
+	}
+
+	if len(all) == 0 {
+		return nil, documents.ErrDocumentNotFound
```
mikiquantum: I do not think this is necessarily and error. I would return an empty map and no error

documents/entityrelationship/repository.go
```diff
@@ -29,14 +31,58 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
...
+	if relationships == nil {
+		return nil, documents.ErrDocumentNotFound
+	}
+
+	for _, r := range relationships {
+		e, ok := r.(*EntityRelationship)
+		if !ok {
+			continue
+		}
+		if bytes.Equal(e.EntityIdentifier, entityIdentifier) {
```
mikiquantum: Can you inline it like:

```suggestion
		if bytes.Equal(e.EntityIdentifier, entityIdentifier) && targetDID.Equal(*e.TargetIdentity) {
```

documents/entityrelationship/repository.go
```diff
@@ -29,14 +31,58 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
...
+	return nil, documents.ErrDocumentNotFound
 }
 
-// EntityExists returns true if a entity relationship exists for a given entity identifier and target did
-func (r *repo) EntityExists(entityIdentifier []byte, targetDID identity.DID) (bool, error) {
-	// todo not implemented
-	return false, nil
+// ListAllRelationships returns a list of all entity relationship identifiers in which a given entity is involved
+func (r *repo) ListAllRelationships(entityIdentifier []byte, ownerDID identity.DID) (map[string][]byte, error) {
+	relationships, err := r.db.GetAllByPrefix(string(ownerDID[:]))
```
mikiquantum:
```suggestion
	everything, err := r.db.GetAllByPrefix(string(ownerDID[:]))
```

documents/entityrelationship/repository.go
```diff
@@ -29,14 +31,58 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
...
+	if err != nil {
+		return nil, err
+	}
+
+	if relationships == nil {
```
mikiquantum: I would return an empty map here as well, instead of error.

documents/entityrelationship/service.go
```diff
@@ -190,33 +190,93 @@ func (s service) DeriveEntityRelationshipResponse(model documents.Model) (*entit
...
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := s.GetCurrentVersion(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	model, err := r.(*EntityRelationship).DeleteAccessToken(ctx, hexutil.Encode(did[0][:]))
```
mikiquantum: At the moment is fine, as there is no other operations over the ER. In the near future, this logic is not as agnostic as the `Derive` method should be.

documents/entityrelationship/repository.go
charleenfei: okay, will change, makes sense

documents/entityrelationship/service.go
charleenfei: agree, in the future we can perhaps specify to DeriveFromRevokePayload which would differ from DeriveFromUpdatePayload

documents/entityrelationship/repository.go
```diff
@@ -29,14 +31,56 @@ func newDBRepository(db storage.Repository, docRepo documents.Repository) reposi
 	return r
 }
 
-// Find returns a EntityRelationship based on a entity id and a targetDID
-func (r *repo) Find(entityIdentifier []byte, targetDID identity.DID) (*EntityRelationship, error) {
-	// todo not implemented
-	return nil, nil
+// Find returns the latest (second if revoked) version of an EntityRelationship based on a entity id and a targetDID
```
manuelpolzhofer: The comment needs to be adjusted for the new purpose

charleenfei: what new purpose?

charleenfei: 👍

PR #944, opened 2019-04-04
P2P error masking

vimukthi-git:
Issue Link: <Github link>


build/configs/default_config.yaml
```diff
@@ -154,5 +154,8 @@ debug:
   # pprof for debugging
   pprof: false
 
+  # whether to mask errors returned in p2p responses for security
+  maskErrors: false
```
lucasvo: That shouldn't be an option. Options that make a product less secure are bad options.

p2p/receiver/handler.go
```diff
@@ -47,34 +47,34 @@ func New(
 // HandleInterceptor acts as main entry point for all message types, routes the request to the correct handler
 func (srv *Handler) HandleInterceptor(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *pb.P2PEnvelope) (*pb.P2PEnvelope, error) {
 	if msg == nil {
-		return convertToErrorEnvelop(errors.New("nil payload provided"))
+		return srv.convertToErrorEnvelop(errors.New("nil payload provided"))
```
lucasvo: This shouldn't be masked as it doesn't not result in any data leak if returned.

p2p/receiver/handler.go
```diff
@@ -47,34 +47,34 @@ func New(
...
 	}
 	envelope, err := p2pcommon.ResolveDataEnvelope(msg)
 	if err != nil {
-		return convertToErrorEnvelop(err)
+		return srv.convertToErrorEnvelop(err)
```
lucasvo: I don't think this does either, right?

vimukthi-git:
@lucasvo Thanks for the good review ;)
as agreed on DM, I will
- remove the config option and replace with some compile time flag
- just keep masking for all errors
- create a ticket to research parsing logs in testworld

PR #948, opened 2019-04-08
p2p: getDocumentRequest for entity

manuelpolzhofer:


manuelpolzhofer:
I commit the first draft.
I have some questions in general:

- should RequestDocumentWithToken method be in the processor?
  - currently, the processor is more related to anchoring
- in the service, we have a method called ReceiveAnchoredDocument
   - I will make a more generic method out of it. because some parts I can reuse for GetDocument.
   - ReceivedAnchoredDocumentValidator etc.

Currently, the differences between ReceiveAnchoredDocument and GetDocument are:
- no webhook needed because GetDocument is sync
- GetDocument does not always need Update for saving the document

charleenfei:
> I commit the first draft.
> I have some questions in general:
> 
> * should RequestDocumentWithToken method be in the processor?
>   
couldn't we have a more generic request method which takes multiple access types, like how the receiver is set up?

documents/entity/bootstrapper.go
```diff
@@ -63,11 +63,16 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return errors.New("entity relation service not initialised")
 	}
 
+	processor, _ := ctx[documents.BootstrappedAnchorProcessor].(documents.AnchorProcessor)
+	/*if !ok {
```
manuelpolzhofer: I want to move the processor from documents.PostBootstrapper to documents.Bootstrapper

documents/anchor.go
```diff
@@ -18,6 +20,7 @@ type AnchorProcessor interface {
 	PreAnchorDocument(ctx context.Context, model Model) error
 	AnchorDocument(ctx context.Context, model Model) error
 	SendDocument(ctx context.Context, model Model) error
+	RequestDocumentWithToken(ctx context.Context, requesterID identity.DID, tokenIdentifier, entityIdentifier, entityRelationIdentifier []byte) (*p2ppb.GetDocumentResponse, error)
```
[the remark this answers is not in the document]
manuelpolzhofer: I agree the processor should handle the request

documents/coredocument.go
```diff
@@ -122,16 +122,16 @@ func NewCoreDocumentWithCollaborators(documentPrefix []byte, collaborators Colla
 
 // NewCoreDocumentWithAccessToken generates a new core document with a document type specified by the prefix.
 // It also adds the targetID as a read collaborator, and adds an access token on this document for the document specified in the documentID parameter
-func NewCoreDocumentWithAccessToken(ctx context.Context, documentPrefix []byte, params documentpb.AccessTokenParams) (*CoreDocument, error) {
+func NewCoreDocumentWithAccessToken(ctx context.Context, documentPrefix []byte, params documentpb.AccessTokenParams, selfDID identity.DID) (*CoreDocument, error) {
```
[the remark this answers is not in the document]
manuelpolzhofer: 👍

documents/coredocument.go
```diff
@@ -463,12 +463,12 @@ func (cd *CoreDocument) coredocTree(docType string) (tree *proofs.DocumentTree,
 // GetSignerCollaborators returns the collaborators excluding the filteredIDs
 // returns collaborators with Read_Sign permissions.
 func (cd *CoreDocument) GetSignerCollaborators(filterIDs ...identity.DID) ([]identity.DID, error) {
-	cs, err := cd.GetCollaborators(filterIDs...)
+	sign, err := cd.getReadCollaborators(coredocumentpb.Action_ACTION_READ_SIGN)
```
manuelpolzhofer: @vedhavyas I needed to change this method. 
You changed it to `cs.ReadWriteCollaborators`. Why?

A `grantee` for an entity relationship is only read_sign. That's why I changed it.

testworld/p2p_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// alice anchors entity
+	res := createDocument(alice.httpExpect, alice.id.String(), typeEntity, http.StatusOK, defaultEntityPayload(alice.id.String(), []string{}))
+	txID := getTransactionID(t, res)
+	status, message := getTransactionStatusAndMessage(alice.httpExpect, alice.id.String(), txID)
+	if status != "success" {
+		t.Error(message)
+	}
+	entityIdentifier := getDocumentIdentifier(t, res)
+
```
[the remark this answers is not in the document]
manuelpolzhofer: Yes. we should add more tests.

documents/anchor.go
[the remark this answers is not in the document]
manuelpolzhofer: I think I prefer three methods otherwise you always need to pass a lot of empty parameters. (nft registry, etc).

We could make a more generic method in the service itself.

documents/anchor.go
```diff
@@ -18,6 +19,7 @@ type AnchorProcessor interface {
 	PreAnchorDocument(ctx context.Context, model Model) error
 	AnchorDocument(ctx context.Context, model Model) error
 	SendDocument(ctx context.Context, model Model) error
+	RequestDocumentWithToken(ctx context.Context, tokenIdentifier, entityIdentifier, entityRelationIdentifier []byte) (*p2ppb.GetDocumentResponse, error)
```
mikiquantum:
```suggestion
	RequestDocumentWithAccessToken(ctx context.Context, tokenIdentifier, documentIdentifier, delegatingDocumentIdentifier []byte) (*p2ppb.GetDocumentResponse, error)
```
As a broader question, why do we need this definition in the `Anchor.go` file? I do not think has to do anything with `anchoring`. I would like to not expose any `p2ppb.*` here.
If you want to have a common method that is used by many other documents, I would create a new interface, maybe DocumentProcessor? or DocumentGetter? in a different file or under the processor.go that encapsulates the proto file calls.

manuelpolzhofer: I agree. The processor was previously only used for anchoring but I would argue we can see it in a more generic way. For me, the processor communicates with the p2p layer.

`service -> processor -> p2p`

The method for transforming a `p2ppb.*` response to a model would be in the service.
I can't access the service in the processor it would create a cyclic dependency, that is the reason why I am returning the `p2ppb.*`

mikiquantum: Yes, that is fair. My suggestion is to move that definition to a different interface implemented by the defaultProcessor similar to `Send` that abstracts the p2ppb logic.

documents/entity/service.go
```diff
@@ -3,10 +3,11 @@ package entity
 import (
 	"context"
 
+	"github.com/centrifuge/go-centrifuge/documents/entityrelationship"
+
```
charleenfei: spacing

documents/coredocument.go
vedhavyas: because ReadWrite Will have all the collaborators that are required to sign the document update

vedhavyas: > A grantee for an entity relationship is only read_sign. That's why I changed it.
Not sure what you mean by grantee but this function on CoreDocument should ot tie with entity or any specific document

manuelpolzhofer: Resolved in discussion with Ved

PR #949, opened 2019-04-08
 Add nft index to api response

vimukthi-git:
Issue Link: <Github link>


build/configs/default_config.yaml
```diff
@@ -65,9 +65,9 @@ networks:
     ethereumNetworkId: 3
     # Latest deployed Smart Contracts for the given testnet
     contractAddresses:
-      identityFactory: "0xF86177b275E4Fcb2Ab5A3FcF5E46aab51f73D7a6"
-      anchorRepository: "0xf630d8bE5664C7FB1234CEf48b6A729cEe3B71D8"
-      invoiceUnpaid: "0x9eee8FbE3E29c96cC944F0eeC1594fe1b633dABe"
+      identityFactory: "0x0d391E72325CedC41E48053E77288a89F2E31115"
```
mikiquantum: Rinkeby/kovan addresses didnt change?

documents/converters.go
```diff
@@ -329,30 +331,66 @@ func ToClientCollaboratorAccess(ca CollaboratorsAccess) (*documentpb.ReadAccess,
...
-	//	return nil, err
-	//}
-	//
-	//time, err := model.Timestamp()
-	//if err != nil {
-	//	return nil, err
-	//}
+	author := ""
+	a, err := model.Author()
+	// we ignore error here because it can happen when a model us first created but its not anchored yet
```
mikiquantum:
```suggestion
	// we ignore error here because it can happen when a model is first created but its not anchored yet
```

mikiquantum: In which case why do we need the err check? why not just assign author=a.String()

documents/converters.go
```diff
@@ -329,30 +331,66 @@ func ToClientCollaboratorAccess(ca CollaboratorsAccess) (*documentpb.ReadAccess,
...
+	time := ""
+	t, err := model.Timestamp()
+	if err == nil {
+		time = t.UTC().String()
+	}
 
+	nfts := model.NFTs()
+	cnfts, err := convertNFTs(tokenRegistry, nfts)
+	if err != nil {
+		// this could be a temporary failure, so we ignore but warn about the error
```
mikiquantum: what is the impact of this erroring out?

documents/converters.go
```diff
@@ -329,30 +331,66 @@ func ToClientCollaboratorAccess(ca CollaboratorsAccess) (*documentpb.ReadAccess,
...
+		CreatedAt:   time,
 		ReadAccess:  rcs,
 		WriteAccess: wcs,
+		Nfts:        cnfts,
 	}, nil
 }
+
+func convertNFTs(tokenRegistry TokenRegistry, nfts []*coredocumentpb.NFT) (nnfts []*documentpb.NFT, err error) {
+	for _, n := range nfts {
+		regAddress := common.BytesToAddress(n.RegistryId[:20])
```
mikiquantum: Don't we have a constant for the registryLength`20` ?

documents/converters_test.go
```diff
@@ -3,8 +3,13 @@
 package documents
 
 import (
+	"math/big"
+	"strings"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
+
```
mikiquantum: spacing

documents/invoice/bootstrapper.go
```diff
@@ -55,7 +55,13 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	srv := DefaultService(
 		docSrv,
 		repo,
-		queueSrv, txManager)
+		queueSrv, txManager, func() documents.TokenRegistry {
+			tokenRegistry, ok := ctx[bootstrap.BootstrappedInvoiceUnpaid].(documents.TokenRegistry)
+			if !ok {
+				panic("token registry initialisation error")
```
mikiquantum: Don't we usually return error instead? it is easier to build tests for bootstrappers.

documents/purchaseorder/bootstrapper.go
```diff
@@ -52,7 +52,14 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	}
 
 	// register service
-	srv := DefaultService(docSrv, repo, queueSrv, txManager)
+	srv := DefaultService(docSrv, repo, queueSrv, txManager, func() documents.TokenRegistry {
+		tokenRegistry, ok := ctx[bootstrap.BootstrappedInvoiceUnpaid].(documents.TokenRegistry)
+		if !ok {
```
mikiquantum: same here

documents/converters.go
vimukthi-git: I don't think we have, i will add

documents/converters.go
vimukthi-git: its would prevent someone from getting a document through API

documents/converters.go
vimukthi-git: error check is just there to warn. But I realise we can directly us a.String in either case.

build/configs/default_config.yaml
vimukthi-git: Will do so in my next commit after testing locally.

documents/invoice/bootstrapper.go
vimukthi-git: that would be problematic, since we must not run the node in a misconfigured state. For config errors we always panic.

documents/purchaseorder/bootstrapper.go
vimukthi-git: same here :)

documents/invoice/bootstrapper.go
mikiquantum: Isn't it the duty of the main boostrapper? to panic when errors while iterating over all the individual bootstrappers?

documents/converters.go
vimukthi-git: shit I realise I removed the warn anyway

documents/invoice/bootstrapper.go
vimukthi-git: this is a special case since the token registry needs to be injected at a later point because nft package is bootstrapped after the invoice or po packages, hence the closure.

mikiquantum: ah didnt catch that, thanks.

build/configs/default_config.yaml
```diff
@@ -26,9 +26,9 @@ networks:
     ethereumNetworkId: 4
     # Latest deployed Smart Contracts for the given testnet
     contractAddresses: 
-      identityFactory: "0x9ebDD8B7b2e701c9f7d90F81a9716562cAC554b3"
-      anchorRepository: "0x36e3b8b9fff60A1e031fd07671298d05745cfBa2"
-      invoiceUnpaid: "0xE1624AE4766ef3aA4300122D5EAa5a4eA7fC5978"
+      identityFactory: "0xf880b860103bd45a523574AeeC07C492031f4781"
```
mikiquantum: nice

nft/handler.go
```diff
@@ -59,8 +59,7 @@ func (g grpcHandler) MintNFT(ctx context.Context, request *nftpb.NFTMintRequest)
 	}
 
 	return &nftpb.NFTMintResponse{
-		TokenId: resp.TokenID,
-		Header:  &nftpb.ResponseHeader{TransactionId: resp.TransactionID},
+		Header: &nftpb.ResponseHeader{TransactionId: resp.TransactionID},
```
mikiquantum: Cool, cleaner. Lets make sure that the functional tests are updated if needed :)

version/version.go
```diff
@@ -10,7 +10,7 @@ import (
 var gitCommit = "master"
 
 // CentrifugeNodeVersion is the current version of the app
-const CentrifugeNodeVersion = "0.0.3-alpha4"
+const CentrifugeNodeVersion = "0.0.4-alpha4"
```
mikiquantum:
```suggestion
const CentrifugeNodeVersion = "0.0.4"
```

build/swagger_config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
     info: {
-        version: "0.0.3-alpha4",
+        version: "0.0.4-alpha4",
```
mikiquantum:
```suggestion
        version: "0.0.4",
```

PR #952, opened 2019-04-09
updated entity relationship protobuf

charleenfei:


charleenfei:
> LGTM.
> 
> only one small thing. I prefer more describing PR names.
> like "updated entityrelationship protobuf" or "added id to entityrelationship protobuf".
> 
> I case something breaks it easier to find out.

cool, I can change this!

protobufs/entity/service.proto
```diff
@@ -131,6 +131,8 @@ message RelationshipData {
   string owner_identity = 1;
```
manuelpolzhofer: Do we need the `owner_identity`?

We already have the owner DID (selfDID) from the context or is there a case where `owner_identity != selfDID`?

PR #956, opened 2019-04-09
hex string DB index

mikiquantum:
Issue Link: <Github link>


identity/did.go
```diff
@@ -108,6 +108,26 @@ type DID common.Address
 // DIDLength contains the length of a DID
 const DIDLength = common.AddressLength
 
+// MarshalJSON marshals DID to json bytes.
+func (d DID) MarshalJSON() ([]byte, error) {
```
mikiquantum: Added these custom marshaler, so this field appear human readable in the DB

identity/did.go
```diff
@@ -108,6 +108,26 @@ type DID common.Address
...
+	str := "\"" + d.String() + "\""
```
vimukthi-git: why we need the quotes?

mikiquantum: It is the JSON standard way of serializing as string vs other types.

identity/did.go
```diff
@@ -108,6 +108,26 @@ type DID common.Address
...
+	return []byte(str), nil
+}
+
+// UnmarshalJSON loads json bytes to DID
+func (d *DID) UnmarshalJSON(data []byte) error {
+	cleanedData := data
+	if len(data) > 2 && data[0] == '"' && data[len(data)-1] == '"' {
+		cleanedData = data[1 : len(data)-1]
+	}
```
vedhavyas: I think we should use the trim prefix and suffix methods from the `strings` package

PR #957, opened 2019-04-09
Start message for node

vimukthi-git:
- This fixes duplicate protobuf errors when node starts
- Adds a "beta software" message
- Adds a ascii art


protobufs/invoice/service.proto
```diff
@@ -1,11 +1,11 @@
 syntax = "proto3";
 
-package invoice;
+package inv;
 
-option go_package = "invoicepb";
+option go_package = "invpb";
```
mikiquantum: can we add the key `api` or `client` somewhere instead?

cmd/centrifuge/root.go
```diff
@@ -39,6 +41,12 @@ func Execute() {
 	}
 }
 
+func printStartMessage() {
+	c := figure.NewFigure("Centrifuge OS", "", true)
```
mikiquantum: Do you have a screenshot you can attach of how it would look like?

cmd/centrifuge/root.go
```diff
@@ -39,6 +41,12 @@ func Execute() {
...
+	c.Print()
+	fmt.Println("Centrifuge is beta software !!!")
```
mikiquantum: Do we need to add more information about what this beta version means?

protobufs/invoice/service.proto
vimukthi-git: that was the first thing I tried, its just too long.. and we are removing these soon anyway

cmd/centrifuge/root.go
vimukthi-git: Screenshot 2019-04-10 at 11 30 17

cmd/centrifuge/root.go
vimukthi-git: I also thought of this, but for the console its too much information. I think we can explain that in the disclaimer on develop website.

PR #961, opened 2019-04-10
entity: GET EntityWithRelationship 

manuelpolzhofer:
- added second GET endpoint for relationships
- added processor dependency as closure
- added method `entity.service.RequestEntityWithRelationship(...)`
- added new entity testworld test


documents/entity/service_test.go
```diff
@@ -41,7 +41,7 @@ func getServiceWithMockedLayers() (testingcommons.MockIdentityService, *testingc
 		repo,
```
manuelpolzhofer: I will add some unit tests

documents/entity/bootstrapper.go
```diff
@@ -63,16 +64,29 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 		return errors.New("entity relation service not initialised")
 	}
 
-	processor, _ := ctx[documents.BootstrappedAnchorProcessor].(documents.DocumentRequestProcessor)
-	/*if !ok {
-		return errors.New("processor not initialised")
-	}*/
+	processorFinder := func() documents.DocumentRequestProcessor {
```
manuelpolzhofer: added closure pattern to resolve dependencies.

protobufs/entity/service.proto
```diff
@@ -48,6 +48,15 @@ service DocumentService {
       description: "Get the current entity"
     };
   }
+  // Entity Relation Get
+  rpc GetEntityByRelationship(GetRequestRelationship) returns (EntityResponse) {
+    option (google.api.http) = {
+      get: "/entity/{identifier}/relationship/{relationship_identifier}"
```
manuelpolzhofer: First I wanted to pass it as normal GET parameters. 
`/relationship?relationship_identitfier=val1`

What do we prefer?

documents/entity/service_test.go
manuelpolzhofer: We have one testworld test which passes

documents/entity/handler.go
```diff
@@ -156,6 +156,41 @@ func (h *grpcHandler) Get(ctx context.Context, getRequest *cliententitypb.GetReq
...
+		return nil, centerrors.Wrap(err, "identifier is an invalid hex string")
+	}
+
+	erIdentifier, err := hexutil.Decode(getRequest.RelationshipIdentifier)
+	if err != nil {
+		apiLog.Error(err)
+		return nil, centerrors.Wrap(err, "identifier is an invalid hex string")
+	}
+
+	model, err := h.service.RequestEntityWithRelationship(ctxHeader, entityIdentifier, erIdentifier)
```
manuelpolzhofer: I still need to add the following check:
Only request via p2p if the entity is not in the database or if the stored entity is not the latest version.

documents/entity/service.go
```diff
@@ -252,15 +267,31 @@ func (s service) GetVersion(ctx context.Context, documentID []byte, version []by
 	return s.get(ctx, documentID, version)
 }
 
+func (s service) GetEntityByRelationship(ctx context.Context, entityIdentifier, erIdentifier []byte) (documents.Model, error) {
```
charleenfei: maybe you can rename erIdentifier to relationshipIdentifier, for clarity

protobufs/gen/swagger/entity/service.swagger.json
```diff
@@ -98,6 +98,38 @@
...
+          },
+          {
+            "name": "relationship_identifier",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "DocumentService"
```
charleenfei: we could rename this in the corresponding proto file to EntityService, as per the housekeeping discussion yesterday

documents/entity/service.go
```diff
@@ -32,6 +35,12 @@ type Service interface {
 	// DeriveEntityResponse returns the entity model in our standard client format
 	DeriveEntityResponse(entity documents.Model) (*cliententitypb.EntityResponse, error)
 
+	// GetEntityByRelationship returns the entity model from database or requests from granter
+	GetEntityByRelationship(ctx context.Context, entityIdentifier, erIdentifier []byte) (documents.Model, error)
```
charleenfei: same as comments about naming for erIdentifier below

protobufs/entity/service.proto
charleenfei: couldn't we do `get/relationship/{relationship_id}/entity` ?

manuelpolzhofer:
Clarified with @charleenfei . `requesterID` will be renamed to `granterDID`.

protobufs/entity/service.proto
manuelpolzhofer: mhm. We would break the pattern a bit. 

@lucasvo do you have a preference?

documents/entity/service.go
```diff
@@ -32,6 +33,12 @@ type Service interface {
 	// DeriveEntityResponse returns the entity model in our standard client format
 	DeriveEntityResponse(entity documents.Model) (*cliententitypb.EntityResponse, error)
 
+	// GetEntityByRelationship returns the entity model from database or requests from granter
+	GetEntityByRelationship(ctx context.Context, entityIdentifier, relationshipIdentifier []byte) (documents.Model, error)
+
+	// RequestEntityFromCollaborator requests an entity with an entity relationship from granter
```
mikiquantum:
```suggestion
	// RequestEntityWithRelationship requests an entity with an entity relationship from granter
```

documents/entity/service.go
```diff
@@ -252,15 +265,31 @@ func (s service) GetVersion(ctx context.Context, documentID []byte, version []by
...
+func (s service) GetEntityByRelationship(ctx context.Context, entityIdentifier, relationshipIdentifier []byte) (documents.Model, error) {
+	if s.Service.Exists(ctx, entityIdentifier) {
+		entity, err := s.Service.GetCurrentVersion(ctx, entityIdentifier)
+		if err != nil {
+			// in case of an error try to get document from collaborator
+			return s.RequestEntityWithRelationship(ctx, entityIdentifier, relationshipIdentifier)
+		}
+
+		// check if stored document is the latest version
+		if err := documents.PostAnchoredValidator(s.idService, s.anchorRepo).Validate(nil, entity); err != nil {
```
mikiquantum: Here you are implying that with any validation error (not explicitely the latest document) you will request the latest. Is this just a placeholder, until you add the latest logic in following PR?

protobufs/entity/service.proto
mikiquantum: This is what we discussed in a meeting. We agreed on:
`/entity/{identifier}/relationship/{relationship_identifier}`

documents/entity/bootstrapper.go
```diff
@@ -63,16 +64,29 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
...
+		processor, ok := ctx[documents.BootstrappedAnchorProcessor].(documents.DocumentRequestProcessor)
+		if !ok {
+			panic("processor initialisation error")
```
mikiquantum: I am not super keen on this type of handling as they will fail as panic when the node is running...which can cause other unforseen inconsistencies on already running processes. Why we cannot check that in bootstrap?

documents/entity/service.go
manuelpolzhofer: I opened a second PR for latestDocumentValidator.

We could only check the latest document or doing the complete `PostAnchorValidator` checking the roots etc.

documents/entity/service.go
```diff
@@ -286,17 +315,50 @@ func (s service) get(ctx context.Context, documentID, version []byte) (documents
...
+
+	model, err := s.Service.DeriveFromCoreDocument(*response.Document)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := documents.PostAnchoredValidator(s.idService, s.anchorRepo).Validate(nil, model); err != nil {
+		return nil, errors.NewTypedError(documents.ErrDocumentInvalid, err)
+	}
+
```
manuelpolzhofer: I forgot the store the model in the db

documents/entity/bootstrapper.go
manuelpolzhofer: I adopted the pattern from the invoice bootstrapper with the NFT dependency. 

Currently, the processor is bootstrapped in the PostDocumentBoostrapper because the processor depends on p2p etc.

protobufs/entity/service.proto
mikiquantum: Actually, just talking to @charleenfei i misunderstood the http method. The call should be:
`relationship/{relationship_identifier}/entity`

documents/entity/service.go
mikiquantum: If we want to perform all validators is fine, but we need to only get the latest document when we failed checking the doc locally only on "not latest doc" validation error, not for all of them. At least at the moment.

documents/entity/bootstrapper.go
mikiquantum: Couldnt we have the entity.Bootstrapper{} invoked after the documents.PostDocumentBootstrapper{}?

manuelpolzhofer: @vimukthi-git Introduced the closure idea.

In some way I like it, because we don't have to care about the dependencies anymore.

manuelpolzhofer: Instead of trying to rearrange the order of the bootstrappers

vimukthi-git: @mikiquantum as explained offline its only needed in specific situations, I think it should be fine to do it here, as there is a legitimate init cycle here :)

PR #963, opened 2019-04-10
add stack traces to error

### LINKED ISSUE #883: Attach stack trace with error
- we need to attach stacktraces to the errors when they are defined
- Make sure we also log the stack trace as `Debug` on log.
This should be done based on a config so that we don't leak unnecessary information on a production deployment.

vedhavyas:
Closes #883 

Changes:
- `errors.New` will attach stack trace by default
- Call `errors.StackTrace(err)` to get the stack trace

Note: this will not change the current `error.Error` log behavior. Stack traces should be retrieved and printed explicitly

Example stack trace:
```
github.com/centrifuge/go-centrifuge/errors.TestWithStackTrace
	/Users/vedhavyas/Projects/Go/src/github.com/centrifuge/go-centrifuge/errors/errors_test.go:169
testing.tRunner
	/usr/local/Cellar/go@1.11/1.11.6/libexec/src/testing/testing.go:827
runtime.goexit
	/usr/local/Cellar/go@1.11/1.11.6/libexec/src/runtime/asm_amd64.s:1333
```


errors/errors.go
```diff
@@ -32,7 +32,8 @@ func (e Error) Error() string {
 // New("some error") returns error with exact string passed
 // New("some error: %v", "some context") returns error with message "some error: some context"
 func New(format string, args ...interface{}) error {
-	return Error(fmt.Sprintf(format, args...))
+	err := Error(fmt.Sprintf(format, args...))
```
vimukthi-git: could you make a bench mark test for an error with and without stack trace? I'm curious about the performance impact of getting the stack trace each time this is called. If the cost is too high I would prefer to have a way to switch it off. Other than that LGTM

vedhavyas: Each error would have around 56 bytes added to them. Since these will be created only at times we need to raise an error, this is an acceptable cost

vimukthi-git: I don't mind about the storage, its not a concern at all.

PR #964, opened 2019-04-10
Fix/prefix and develop

mikiquantum:
Issue Link: <Github link>


documents/entityrelationship/repository.go
```diff
@@ -2,6 +2,7 @@ package entityrelationship
 
 import (
 	"bytes"
+	"github.com/ethereum/go-ethereum/common/hexutil"
```
mikiquantum: fixing this now

PR #966, opened 2019-04-11
swagger API service naming update

### LINKED ISSUE #960: service.swagger.json should have tags specific to the entity
All service.swagger.json have the the tags set to DocumentService. 
The tags is used to generate service clients with a code generator and this means when you a generate a client all methods will go under DocumentServiceApi which will result in method collision:
```
DocumentServiceApi.create() 
DocumentServiceApi.create_1()
DocumentServiceApi.create_2()
```
To fix this we need to define entity specific tags:
1. /invoice/service.swagger.json 
```
"tags": [
   "InvoiceService"
]
```
2. /account/service.swagger.json 
```
"tags": [
   "AccountService"
]
```

charleenfei:
Close #960.


api/service.go
```diff
@@ -105,35 +105,35 @@ func registerAPIs(ctx context.Context, cfg Config, InvoiceUnpaidService nft.Invo
 
 func registerDocumentTypes(ctx context.Context, nodeObjReg map[string]interface{}, grpcServer *grpc.Server, gwmux *runtime.ServeMux, addr string, dopts []grpc.DialOption) error {
 	// register invoice
-	invHandler, ok := nodeObjReg[invoice.BootstrappedInvoiceHandler].(invoicepb.DocumentServiceServer)
+	invHandler, ok := nodeObjReg[invoice.BootstrappedInvoiceHandler].(invoicepb.InvoiceServiceServer)
```
vedhavyas: I feel `invoicepb.InvoiceService` is redundant but I see this is the only way.

protobufs/nft/service.proto
```diff
@@ -10,7 +10,7 @@ option java_package = "com.nft";
 import "google/api/annotations.proto";
 import "protoc-gen-swagger/options/annotations.proto";
 
-// NFTService let's you mint NFTs for a document
+// // NFTService contains all common interactions for NFTs
```
vedhavyas:
```suggestion
// NFTService contains all common interactions for NFTs
```

protobufs/account/service.proto
```diff
@@ -11,7 +11,7 @@ import "google/api/annotations.proto";
 import "google/protobuf/empty.proto";
 import "protoc-gen-swagger/options/annotations.proto";
 
-// AccountService allows to manage accounts in the node
+// AccountService allows for the management of accounts in the node
```
vedhavyas:
```suggestion
// AccountService allows account management on the Node.
```

api/service.go
charleenfei: yea, I agree. But I also can't think of a better way :(

PR #967, opened 2019-04-11
added latestDocumentValidator to PostAnchoredValidator

manuelpolzhofer:
- added validator


documents/validator.go
```diff
@@ -300,6 +300,29 @@ func anchoredValidator(repo anchors.AnchorRepository) Validator {
...
+
+		var empty [anchors.DocumentRootLength]byte
+		var rootByte [anchors.DocumentRootLength]byte
+
+		root, _, err := repo.GetAnchorData(nextID)
+		if err != nil {
+			return errors.New("failed to get document anchor %s from chain: %v", nextID.String(), err)
+		}
+		rootByte = root
+		if rootByte != empty {
```
vimukthi-git: check length instead of creating an empty slice

documents/validator.go
```diff
@@ -300,6 +300,29 @@ func anchoredValidator(repo anchors.AnchorRepository) Validator {
...
+		if err != nil {
+			return errors.New("failed to get anchorID: %v", err)
+		}
+
+		var empty [anchors.DocumentRootLength]byte
+		var rootByte [anchors.DocumentRootLength]byte
+
+		root, _, err := repo.GetAnchorData(nextID)
+		if err != nil {
+			return errors.New("failed to get document anchor %s from chain: %v", nextID.String(), err)
```
vimukthi-git: use typed errors

vimukthi-git:
also add tests :)

documents/validator.go
manuelpolzhofer: the length is always 32 bytes

manuelpolzhofer: I will rename into `empty` in `zeros`

vimukthi-git: Capacity is 32 bytes

manuelpolzhofer:
Yes. unit tests are on the way 
Nearly every testworld test calls the method already :)

documents/validator.go
vimukthi-git: If you are getting len=32 thats a different issue since it returns garbage.

manuelpolzhofer: Solidity returns a 32 byte-array filled with zeros for a non-existing anchor

documents/validator.go
```diff
@@ -350,6 +373,7 @@ func PostAnchoredValidator(idService identity.ServiceDID, repo anchors.AnchorRep
 	return ValidatorGroup{
 		PreAnchorValidator(idService),
 		anchoredValidator(repo),
+		latestVersionValidator(repo),
```
vedhavyas: not sure adding the validator is a good idea. This is not re-usable if I want to use this for old documents. Create a new validator group and use that where you need.

vimukthi-git: since this is called `PostAnchoredValidator, its should be ok no?

documents/validator.go
vimukthi-git: ok I see we have a function that you could use already :) 
`utils.IsEmptyByteSlice(rootByte[:])`

documents/validator.go
mikiquantum: I think is fine to have it inside the `PostAnchoredValidator`, `latestVersionValidator ` should be public, as I believe it will be invoked as well independently.

documents/validator.go
```diff
@@ -300,6 +300,29 @@ func anchoredValidator(repo anchors.AnchorRepository) Validator {
 	})
 }
 
+// latestVersionValidator checks if the document is the latest version
+func latestVersionValidator(repo anchors.AnchorRepository) Validator {
+	return ValidatorFunc(func(_, model Model) error {
+		nextID, err := anchors.ToAnchorID(model.NextVersion())
```
mikiquantum: this will panic if we pass `nil` as model

documents/validator.go
```diff
@@ -300,6 +300,29 @@ func anchoredValidator(repo anchors.AnchorRepository) Validator {
...
+		if err != nil {
+			return errors.New("failed to get anchorID: %v", err)
+		}
+
+		var empty [anchors.DocumentRootLength]byte
+		var rootByte [anchors.DocumentRootLength]byte
+
+		root, _, err := repo.GetAnchorData(nextID)
```
mikiquantum: Are we sure that the GetAnchorData returns nil error if anchor not found?

manuelpolzhofer: yes. testworld test already passed

manuelpolzhofer:
@vimukthi-git implemented your change requests and added unit tests.

PR #968, opened 2019-04-11
Added embarcadero boot nodes

mikiquantum:
Issue Link: <Github link>


build/configs/default_config.yaml
```diff
@@ -69,13 +69,13 @@ networks:
       anchorRepository: "0x2C80F1DafDa7aADeb2C175d4B43A858473E11713"
       invoiceUnpaid: "0xE73841140Fce93912Ec769dAd8F482Bf5c768DA5"
 
-  # Main net test network
-  embarcaderotest:
+  # Mainnet network
+  embarcadero:
```
vimukthi-git: hmm I wonder how we can test with new deployments with this change, as we need a separate set of contracts as test addresses always.

PR #973, opened 2019-04-11
GET relationship: correct endpoint + db store

manuelpolzhofer:
- correct url endpoint
- removed closure bootstrapper for processor
- private requestEntityReleationship method in entity service
- store received entity in db


documents/entity/service.go
```diff
@@ -265,22 +262,30 @@ func (s service) GetVersion(ctx context.Context, documentID []byte, version []by
...
 }
 
-func (s service) GetEntityByRelationship(ctx context.Context, entityIdentifier, relationshipIdentifier []byte) (documents.Model, error) {
+func (s service) GetEntityByRelationship(ctx context.Context, relationshipIdentifier []byte) (documents.Model, error) {
+	model, err := s.erService.GetCurrentVersion(ctx, relationshipIdentifier)
+	if err != nil {
+		return nil, entityrelationship.ErrERNotFound
+	}
+
+	relationship := model.(*entityrelationship.EntityRelationship)
```
charleenfei: maybe put a relationship, ok check here in case casting the model to EntityRelationship fails (if an entityID is used instead of a relationshipID, for example)

documents/entity/service.go
```diff
@@ -265,22 +262,30 @@ func (s service) GetVersion(ctx context.Context, documentID []byte, version []by
 	return s.get(ctx, documentID, version)
 }
 
-func (s service) GetEntityByRelationship(ctx context.Context, entityIdentifier, relationshipIdentifier []byte) (documents.Model, error) {
+func (s service) GetEntityByRelationship(ctx context.Context, relationshipIdentifier []byte) (documents.Model, error) {
```
charleenfei: add a comment to this method?

documents/entity/service.go
```diff
@@ -358,9 +358,36 @@ func (s service) RequestEntityWithRelationship(ctx context.Context, entityIdenti
 		return nil, errors.NewTypedError(documents.ErrDocumentInvalid, err)
 	}
 
+	err = s.store(ctx, model)
```
charleenfei: if err = s.store(ctx, model); err != nil ...

documents/entity/service.go
```diff
@@ -358,9 +358,36 @@ func (s service) RequestEntityWithRelationship(ctx context.Context, entityIdenti
...
+
```
vedhavyas: spacing

PR #977, opened 2019-04-12
Minor UX things

vimukthi-git:
Issue Link: <Github link>


config/configuration.go
```diff
@@ -452,7 +452,7 @@ func (c *configuration) GetPrecommitEnabled() bool {
 
 // GetLowEntropyNFTTokenEnabled returns true if low entropy nft token IDs are not enabled
 func (c *configuration) GetLowEntropyNFTTokenEnabled() bool {
```
pstehlik: preference would be to name the `getter` same as the config variable to avoid having to make another mental step. not worth it right now to make another PR though.

PR #981, opened 2019-04-12
fix p2p return errors

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/testworld/park.go
+++ b/testworld/park.go
@@ -144,6 +144,8 @@ func (r *hostManager) init(createConfig bool) error {
 	}
 	// make sure hosts are alive and print host centIDs
 	for name, host := range r.niceHosts {
+		// Temporary until we have a proper healthcheck in place
+		time.Sleep(2 * time.Second)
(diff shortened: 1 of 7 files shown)


testworld/park.go
```diff
@@ -144,6 +144,8 @@ func (r *hostManager) init(createConfig bool) error {
 	}
 	// make sure hosts are alive and print host centIDs
 	for name, host := range r.niceHosts {
+		// Temporary until we have a proper healthcheck in place
+		time.Sleep(2 * time.Second)
```
mikiquantum: Found timing error when running locally for subsequent times after first one.

testworld/park.go
```diff
@@ -281,6 +283,16 @@ func (h *host) init() error {
 		if err != nil {
 			return err
 		}
+
+		values := map[string]interface{}{
```
mikiquantum: @vimukthi-git we update the config and add the keys only for testworld. So we can run tests without having to rerun the host creation.

PR #999, opened 2019-04-23
New big int type

vimukthi-git:
Issue Link: <Github link>

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -224,7 +224,7 @@ func (cd *CoreDocument) PrepareNewVersion(documentPrefix []byte, collaborators C
 func copyAttrMap(attributes map[string]*attribute) map[string]*attribute {
 	m := make(map[string]*attribute)
 	for k, v := range attributes {
-		m[k] = v
+		m[k] = v.copy()
 	}
 	return m
 }
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -214,9 +214,19 @@ func (cd *CoreDocument) PrepareNewVersion(documentPrefix []byte, collaborators C
 	ncd := &CoreDocument{Document: cdp}
 	ncd.addCollaboratorsToReadSignRules(rcs)
 	ncd.addCollaboratorsToTransitionRules(documentPrefix, wcs)
+	ncd.Attributes = copyAttrMap(cd.Attributes)
 	ncd.Modified = true
 	return ncd, nil
 
+}
+
+// copyAttrMap copies the attributes map
+func copyAttrMap(attributes map[string]*attribute) map[string]*attribute {
+	m := make(map[string]*attribute)
+	for k, v := range attributes {
+		m[k] = v
+	}
+	return m
 }
 
 // newRole returns a new role with random role key
@@ -623,18 +633,23 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
 }
 
 // AddAttribute adds a custom attribute to the model with the given value. If an attribute with the given name already exists, it's updated.
-func (cd *CoreDocument) AddAttribute(name string, attributeType AllowedAttributeType, value string) error {
+func (cd *CoreDocument) AddAttribute(name string, attributeType attributeType, value string) (*CoreDocument, error) {
+	ncd, err := cd.PrepareNewVersion(nil, CollaboratorsAccess{})
+	if err != nil {
+		return nil, errors.NewTypedError(ErrCDAttribute, errors.New("failed to prepare new version: %v", err))
+	}
 	// TODO convert value from string to correct type
 	// For now its only string that is supported
 	nAttr, err := newAttribute(name, attributeType, value)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	if cd.Attributes == nil {
-		cd.Attributes = make(map[string]*attribute)
+	if ncd.Attributes == nil {
+		ncd.Attributes = make(map[string]*attribute)
 	}
-	cd.Attributes[name] = nAttr
-	return nil
+	ncd.Attributes[name] = nAttr
+	ncd.Modified = true
+	return ncd, nil
 }
 
 // GetAttribute gets the attribute with the given name from the model together with its type, it returns a non-nil error if the attribute doesn't exist or can't be retrieved.
@@ -647,9 +662,14 @@ func (cd *CoreDocument) GetAttribute(name string) (hashedKey []byte, attrType st
 }
 
 // DeleteAttribute deletes a custom attribute from the model
-func (cd *CoreDocument) DeleteAttribute(name string) error {
-	delete(cd.Attributes, name)
-	return nil
+func (cd *CoreDocument) DeleteAttribute(name string) (*CoreDocument, error) {
+	ncd, err := cd.PrepareNewVersion(nil, CollaboratorsAccess{})
+	if err != nil {
+		return nil, errors.NewTypedError(ErrCDAttribute, errors.New("failed to prepare new version: %v", err))
+	}
+	delete(ncd.Attributes, name)
+	ncd.Modified = true
+	return ncd, nil
 }
 
 func populateVersions(cd *coredocumentpb.CoreDocument, prevCD *coredocumentpb.CoreDocument) (err error) {
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -189,7 +189,7 @@ func (cd *CoreDocument) PrepareNewVersion(documentPrefix []byte, collaborators C
 	// get all the old collaborators
 	oldCs, err := cd.GetCollaborators()
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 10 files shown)


documents/coredocument.go
```diff
@@ -208,17 +208,27 @@ func (cd *CoreDocument) PrepareNewVersion(documentPrefix []byte, collaborators C
...
 	ncd.Modified = true
 	return ncd, nil
 
 }
 
+// copyAttrMap copies the attributes map
+func copyAttrMap(attributes map[string]*attribute) map[string]*attribute {
+	m := make(map[string]*attribute)
+	for k, v := range attributes {
+		m[k] = v
```
vedhavyas: This is still a shallow copy since attribute `v` is a pointer. Lets make a copy of the value attribute and assign its pointer

documents/int256.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"math/big"
+
+	"github.com/centrifuge/go-centrifuge/errors"
+)
+
+// Int256 represents a signed 256 bit integer
+type Int256 struct {
+	v *big.Int
```
vedhavyas: lets make the `v` a value since `Int256` will always be pointer.

documents/int256_test.go
```diff
@@ -0,0 +1,66 @@
...
+			"-57896044618658097711785492504343953926634992332820282019728792003956564819968",
+			true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			i, err := NewInt256(test.val)
+			if test.isErr {
+				assert.Error(t, err)
+			} else {
```
vedhavyas: you can avoid the else by using `continue` in the `if` and de-indenting the else case

documents/int256_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{
+			"more than allowed int256 (+1)",
+			"57896044618658097711785492504343953926634992332820282019728792003956564819968",
+			true,
+		},
+		{
+			"less than allowed int256 (-1)",
+			"-57896044618658097711785492504343953926634992332820282019728792003956564819968",
+			true,
+		},
```
vedhavyas: lets add a decimal failure scenario as well ?

documents/int256.go
```diff
@@ -0,0 +1,44 @@
...
+	nn := new(big.Int)
+	nn.SetString(n, 10)
+	if !isValidInt256(*nn) {
+		return nil, errors.NewTypedError(ErrInvalidInt256, errors.New("value: %s", n))
+	}
+	return &Int256{nn}, nil
+}
+
+// Bytes returns the byte representation of this int256
+func (i *Int256) Bytes() []byte {
```
vedhavyas: What about the `FromBytes` since this will be needed to convert back and forth from p2p layer to model layer

documents/int256_test.go
vimukthi-git: I find the above more readable

vimukthi-git:
@vedhavyas the coverage would not increase until we start using attribute stuff, I will be making that pr soon. For now the most important parts are covered.

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+var validInt = regexp.MustCompile(`^[-+]?\d+$`)
+
+// Int256 represents a signed 256 bit integer
+type Int256 struct {
+	v big.Int
+}
+
+// NewInt256 creates a new Int256 given a string
+func NewInt256(n string) (*Int256, error) {
+	if !validInt.MatchString(n) {
```
vimukthi-git: @vedhavyas had to add this because big.NewInt silently ignores decimals

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if !isValidInt256(*nn) {
+		return nil, errors.NewTypedError(ErrInvalidInt256, errors.New("value: %s", n))
+	}
+	return &Int256{*nn}, nil
+}
+
+// FromBytes converts the a big endian byte slice to an Int256
+func FromBytes(b []byte) (*Int256, error) {
+	if len(b) != 32 {
```
vedhavyas: Do we have to make it byte 32 all the time? it too much space IMO

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Equals checks if the given two Int256s are equal
+func (i *Int256) Equals(o *Int256) bool {
+	if i.v.Cmp(&o.v) != 0 {
+		return false
+	}
+	return true
+}
+
+func isValidInt256(n big.Int) bool {
```
vedhavyas: can you add more comments to this with some examples here. Its not very clear in the first look

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+	x := n.BitLen()
+	if x > 256 {
+		return false
+	}
+
+	// check max
+	two := big.NewInt(2)
+	exp := two.Exp(two, big.NewInt(255), nil)
+	maxI256 := exp.Sub(exp, big.NewInt(1))
```
vedhavyas: this can be generated only once instead of creating everytime. What do you think ?

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+func (i *Int256) Equals(o *Int256) bool {
+	if i.v.Cmp(&o.v) != 0 {
+		return false
+	}
+	return true
+}
+
+func isValidInt256(n big.Int) bool {
+	x := n.BitLen()
+	if x > 256 {
```
vedhavyas: Dont think you need this check if the we are comparing the maxint in the bottom

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+// Bytes returns the big endian 32 byte representation of this int256. First bit of LSB is used as the sign bit.
+func (i *Int256) Bytes() [32]byte {
+
+	var b [32]byte
+	// no of bits in i.v.Bytes() <= 255
+	// if its less pad the number in big endian order and copy to the 32 byte array
+	copy(b[:], math.PaddedBigBytes(&i.v, 32))
+
+	//  set the sign bit
+	b[signByteIdx] = byteutils.ClearBit(b[signByteIdx], 7)
```
vedhavyas: If the opertions on Int256 leads to result above max value, this function should error out instead of ignoring the excess bytes. Else, we might end up with unexpected results

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Int256 represents a signed 256 bit integer
+type Int256 struct {
+	v big.Int
+}
+
+// NewInt256 creates a new Int256 given a string
+func NewInt256(n string) (*Int256, error) {
+	if !validInt.MatchString(n) {
+		return nil, errors.NewTypedError(ErrInvalidInt256, errors.New("probably a decimal value: %s", n))
```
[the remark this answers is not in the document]
vimukthi-git: it is probably because it can fail for other strings as well.

documents/int256.go
vimukthi-git: yes, as the format needs to be bigendian with sign at LSB

documents/int256.go
vimukthi-git: sure

documents/int256.go
vimukthi-git: fail fast

documents/int256.go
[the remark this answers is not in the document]
vedhavyas: Then, lets say invalid integer ?

documents/int256.go
vedhavyas: will not be a fail fast if the max int is already generated. It would still be one operation :)

documents/int256.go
vimukthi-git: As the comment on line 73 says it's an invariant for int256 type that is enforced, this function shouldn't care about the size. When we have more ops we can add the check to those as well as I do now for `NewInt256`.

documents/int256.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// negative
+	if neg {
+		nn = nn.Neg(nn)
+	}
+
+	return &Int256{*nn}, nil
+}
+
+// Bytes returns the big endian 32 byte representation of this int256. First bit of LSB is used as the sign bit.
```
vedhavyas: Its not LSB but MSB :)

documents/int256.go
vimukthi-git: also it follows ethereum format

documents/int256.go
vedhavyas: Sure, lets add a todo here so that we track this. Better yet, change the function signature to return an error so that we dont have to refactor everything when that check comes in.

But this is check we must do though

vimukthi-git: As said the check is not related to this function, it will be on those `op` functions. Right now its impossible for you to create a value more than 256 bits, unless someone hacks the internal big int which we must not do.

documents/int256.go
vimukthi-git: it is LSB

Issue #721, opened 2019-02-04
Client APIs: New fields for Invoice


pstehlik:
Add the new fields for Invoice to client APIs and wire them up to models.
See also https://github.com/centrifuge/centrifuge-protobufs/issues/66

PR #1010, opened 2019-04-26
Custom Attr - Client protobufs

vimukthi-git:
Issue Link: #721


documents/attribute.go
```diff
@@ -16,6 +17,42 @@ func (a attributeType) String() string {
...
+	if err != nil {
+		return AttrKey{}, errors.NewTypedError(ErrCDAttribute, err)
+	}
+	var a [32]byte
+	copy(a[:], hashedKey)
+	return AttrKey(a), nil
+}
+
+// AttrKeyFromBytes converts bytes to AttrKey
+func AttrKeyFromBytes(b []byte) AttrKey {
```
manuelpolzhofer: For a byte array, we are assuming it is already hashed?

documents/coredocument.go
```diff
@@ -634,41 +634,41 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
...
+	nAttr, err := newAttribute(keyLabel, attributeType, value)
 	if err != nil {
 		return nil, err
 	}
 	if ncd.Attributes == nil {
-		ncd.Attributes = make(map[string]*attribute)
+		ncd.Attributes = make(map[AttrKey]Attribute)
 	}
-	ncd.Attributes[name] = nAttr
+	ncd.Attributes[nAttr.Key] = nAttr
```
manuelpolzhofer: We are not checking if the custom attribute already exists.

I guess it is the correct behaviour, needed for update etc.?

documents/coredocument.go
```diff
@@ -634,41 +634,41 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
...
 	ncd.Modified = true
 	return ncd, nil
```
manuelpolzhofer: I don't understand. 
Why are we returning the `coredocument` here?

documents/attribute.go
```diff
@@ -16,6 +17,42 @@ func (a attributeType) String() string {
 	return string(a)
 }
 
+// AttrKey represents a sha256 hash of a attribute label given by a user.
+type AttrKey [32]byte
```
manuelpolzhofer: Why is it public?

Is it not more an internal representation in coredocument.

Outside of the package the key should be seen as a normal string

documents/coredocument.go
```diff
@@ -634,41 +634,41 @@ func (cd *CoreDocument) Timestamp() (time.Time, error) {
...
-func (cd *CoreDocument) AddAttribute(name string, attributeType attributeType, value string) (*CoreDocument, error) {
+func (cd *CoreDocument) AddAttribute(keyLabel string, attributeType attributeType, value string) (*CoreDocument, error) {
 	ncd, err := cd.PrepareNewVersion(nil, CollaboratorsAccess{})
 	if err != nil {
 		return nil, errors.NewTypedError(ErrCDAttribute, errors.New("failed to prepare new version: %v", err))
 	}
 	// TODO convert value from string to correct type
 	// For now its only string that is supported
-	nAttr, err := newAttribute(name, attributeType, value)
+	nAttr, err := newAttribute(keyLabel, attributeType, value)
```
manuelpolzhofer: Currently, the `value` of the attribute is always of type `string`.

What would happen if I pass as attributeType of bytes. Where does the transformation happen?

manuelpolzhofer: I know `newAttribute` can handle all types depending on the passed typ of value.

But currently the variable `value` would be always of type string but the passed attributeType my differ.

PR #1015, opened 2019-04-30
funding: create + get

manuelpolzhofer:
First Funding PR.
testworld tests and register of new endpoint in the next PR


manuelpolzhofer:
@vedhavyas  I added a draft lets discuss it face to face

documents/invoice/model.go
```diff
@@ -615,6 +615,30 @@ func (*Invoice) DocumentType() string {
 	return documenttypes.InvoiceDataTypeUrl
 }
 
+// PrepareNewVersionWithExistingData prepares new version with existing current invoice data
+func (i *Invoice) PrepareNewVersionWithExistingData() (documents.Model, error) {
```
manuelpolzhofer: My idea is to add this method to the model interface.

documents/invoice/model.go
```diff
@@ -615,6 +615,30 @@ func (*Invoice) DocumentType() string {
...
+	newVersion := new(Invoice)
+
+	// copy current data to new version
+	err := newVersion.initInvoiceFromData(i.getClientData())
```
manuelpolzhofer: I didn't want to add another init method.

documents/extension/funding/service.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func defineFundingIdx(model documents.Model) (attr documents.Attribute, err error) {
+	key, err := documents.AttrKeyFromLabel(fundingLabel)
+	if err != nil {
+		return attr, err
+	}
+
+	if !model.AttributeExists(key) {
+		return documents.NewAttribute(fundingLabel, documents.AttrString, "0")
```
manuelpolzhofer: @vedhavyas first I wanted it to define as int256 but it is not so easy to compare `big.Int`

documents/extension/funding/service.go
```diff
@@ -0,0 +1,178 @@
...
+	DeriveFundingResponse(model documents.Model, payload *clientfundingpb.FundingCreatePayload) (*clientfundingpb.FundingResponse, error)
+}
+
+// service implements Service and handles all funding related persistence and validations
+type service struct {
+	documents.Service
+	tokenRegFinder func() documents.TokenRegistry
+}
+
+const fundingLabel = "centrifuge_funding"
```
vedhavyas: do we need to say `centrifuge` here?

documents/extension/funding/service.go
```diff
@@ -0,0 +1,178 @@
...
+		tokenRegFinder: tokenRegFinder,
+	}
+}
+
+func generateKey(idx, fieldName string) string {
+	return strings.Replace(fundingFieldKey, fundingIdx, idx, -1) + fieldName
+}
+
+func keyFromJSONTag(idx, jsonTag string) (key string, err error) {
+	correctJSONParts := 2
```
vedhavyas: we cannot restrict the tags to 2. IMO, we dont need to check for this. The first word in the tag is always name of the field

documents/extension/funding/service.go
vedhavyas: Not everything is String type right? There will be decimals, integers etc...

manuelpolzhofer: Agree I will change it to int256

documents/extension/funding/service.go
manuelpolzhofer: mhm. For the other types, we are using the protobuf github link. Like invoice etc.

documents/extension/funding/service.go
manuelpolzhofer: It was a workaround for removing the protobuf related fields.

`	XXX_NoUnkeyedLiteral struct{}                 `json:"-"`

Otherwise I need to check if the variable name is not equal to `-`

manuelpolzhofer: Anyway it changes if I use a internal struct

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"reflect"
+
+	clientfundingpb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
+)
+
+// Data is the default funding extension schema
+type Data struct {
+	FundingId             string `json:"funding_id,omitempty" attr:"string"`
+	Amount                string `json:"amount,omitempty" attr:"string"`
```
manuelpolzhofer: I will change amount to `Decimal` and other fields like `Days` in the next PR

documents/extension/funding/service.go
```diff
@@ -0,0 +1,178 @@
...
+	jsonKeyIdx := 0
+
+	// example `json:"days,omitempty"`
+	jsonParts := strings.Split(jsonTag, ",")
+	if len(jsonParts) == correctJSONParts {
```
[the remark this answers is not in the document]
manuelpolzhofer: Oh, I forgot that one. I will add it now

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	Fee                   string `json:"fee,omitempty" attr:"string"`
+	RepaymentDueDate      string `json:"repayment_due_date,omitempty" attr:"string"`
+	RepaymentOccurredDate string `json:"repayment_occurred_date,omitempty" attr:"string"`
+	RepaymentAmount       string `json:"repayment_amount,omitempty" attr:"string"`
+	Currency              string `json:"currency,omitempty" attr:"string"`
+	NftAddress            string `json:"nft_address,omitempty" attr:"string"`
+	PaymentDetailsId      string `json:"payment_details_id,omitempty" attr:"string"`
+}
+
+func (f *Data) initFundingFromData(data *clientfundingpb.FundingData) {
```
mikiquantum: Why use reflection here in a specific document extension? Are we planning on adding more extensions soon? in which case then we might want to have its own package with common init functions.

documents/extension/funding/funding_test.go
```diff
@@ -0,0 +1,18 @@
+// +build unit
+
+package funding
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+import clientfundingpb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
```
mikiquantum: we can merge imports here

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+	attr, err := model.GetAttribute(key)
+	if err != nil {
+		return idx, err
+	}
+
+	idx = attr.Value.Int256
+
+	z, err := documents.NewInt256("0")
+	if err != nil {
+		return idx, err
```
mikiquantum: can we return here `nil` instead as the idx ?

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// idx < 0
+	if idx.Cmp(z) == -1 {
+		return idx, ErrFundingIndex
```
mikiquantum: same here

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+		return documents.NewAttribute(fundingLabel, documents.AttrInt256, "0")
+	}
+
+	idx, err := getFundingsLatestIdx(model)
+	if err != nil {
+		return attr, err
+	}
+
+	// increment idx
+	newIdx := idx.Inc()
```
mikiquantum: 👍

documents/extension/funding/service.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	for ; i.Cmp(lastIdx) != 1; i.Inc() {
+		label := generateLabel(i.String(), fundingIDLabel)
+		k, err := documents.AttrKeyFromLabel(label)
+		if err != nil {
+			return idx, err
+		}
+
+		attr, err := model.GetAttribute(k)
```
mikiquantum: As I understand it, here there is an iteration over all the fundings stored to find the correct one. Wouldnt this GetAttribute return error when is not found? and therefore never keep looking for more until the fundingIdx is matched?

documents/extension/funding/service.go
manuelpolzhofer: We could if we use a string pointer

documents/extension/funding/service.go
manuelpolzhofer: Yes, it would return an error.

Do you mean it is not necessary to store the length of the array?
I think it could be useful to have it.

documents/extension/funding/service.go
mikiquantum: I meant that as idx is nil until we assign: `idx = attr.Value.Int256` then after any error will have the value of the attr. to be consistent is better to return `nil` here explicitely. Unless I am missing something else.

documents/extension/funding/service.go
mikiquantum: Why do we iterate then? if for more than 1 entry we will never look forward? That is my only question.

manuelpolzhofer: A document could have multiple funding agreements.

The `funding_id` is a value and not a key. We might have 10 fundings but we don't know which index has the funding with the `funding_id` we are looking for.

If we would store it like `centrifuge_funding[{funding_id}]` it would not be possible to return a list of all funding agreements.

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"reflect"
+
+	clientfundingpb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
+)
+
+// Data is the default funding extension schema
+type Data struct {
+	FundingId             string `json:"funding_id,omitempty" attr:"string"`
```
vedhavyas: this will be of type bytes

documents/extension/funding/funding.go
vedhavyas: this will be decimal

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	Amount                string `json:"amount,omitempty" attr:"string"`
+	Apr                   string `json:"apr,omitempty" attr:"string"`
+	Days                  string `json:"days,omitempty" attr:"integer"`
+	Fee                   string `json:"fee,omitempty" attr:"string"`
+	RepaymentDueDate      string `json:"repayment_due_date,omitempty" attr:"string"`
```
vedhavyas: timestamp

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	RepaymentOccurredDate string `json:"repayment_occurred_date,omitempty" attr:"string"`
```
vedhavyas: timestamp

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	RepaymentAmount       string `json:"repayment_amount,omitempty" attr:"string"`
```
vedhavyas: decimal

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	clientfundingpb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
+)
+
+// Data is the default funding extension schema
+type Data struct {
+	FundingId             string `json:"funding_id,omitempty" attr:"string"`
+	Amount                string `json:"amount,omitempty" attr:"string"`
+	Apr                   string `json:"apr,omitempty" attr:"string"`
+	Days                  string `json:"days,omitempty" attr:"integer"`
+	Fee                   string `json:"fee,omitempty" attr:"string"`
```
vedhavyas: apr and fee should also be timestamp

documents/extension/funding/funding.go
vedhavyas: FundingId -> FundingID

documents/extension/funding/funding.go
```diff
@@ -0,0 +1,33 @@
...
+	RepaymentDueDate      string `json:"repayment_due_date,omitempty" attr:"string"`
+	RepaymentOccurredDate string `json:"repayment_occurred_date,omitempty" attr:"string"`
+	RepaymentAmount       string `json:"repayment_amount,omitempty" attr:"string"`
+	Currency              string `json:"currency,omitempty" attr:"string"`
+	NftAddress            string `json:"nft_address,omitempty" attr:"string"`
+	PaymentDetailsId      string `json:"payment_details_id,omitempty" attr:"string"`
```
vedhavyas: NftAddress -> NFTAddress

vedhavyas: PaymentDetailsId -> PaymentDetailsID

documents/extension/funding/funding.go
vedhavyas: I think it is better to write the converter for this instead.

documents/extension/funding/handler.go
```diff
@@ -0,0 +1,100 @@
...
+	// returns model with added funding custom fields
+	model, err := h.service.DeriveFromPayload(ctxHeader, req, identifier)
+	if err != nil {
+		return nil, err
+	}
+
+	model, jobID, _, err := h.service.Update(ctx, model)
+	if err != nil {
+		apiLog.Error(err)
+		return nil, centerrors.Wrap(err, "could not update document")
```
vedhavyas: lets not use `centerrors`. they are deprecated. Use `errors` instead

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// service implements Service and handles all funding related persistence and validations
+type service struct {
+	documents.Service
+	tokenRegFinder func() documents.TokenRegistry
+}
+
+const (
+	fundingLabel    = "centrifuge_funding"
```
vedhavyas: `centrifuge` do we need to have this ?

vedhavyas: `funding_agreement` seems more apt IMO

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func labelFromJSONTag(idx, jsonTag string) string {
+	jsonKeyIdx := 0
+	// example `json:"days,omitempty"`
+	jsonParts := strings.Split(jsonTag, ",")
+	return generateLabel(idx, jsonParts[jsonKeyIdx])
+}
+
+func getFundingsLatestIdx(model documents.Model) (idx *documents.Int256, err error) {
```
vedhavyas:
```suggestion
func getFundingsLatestIDX(model documents.Model) (idx *documents.Int256, err error) {
```

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+	// idx < 0
+	if idx.Cmp(z) == -1 {
+		return idx, ErrFundingIndex
+	}
+
+	return idx, nil
+
+}
+
+func defineFundingAttrIdx(model documents.Model) (attr documents.Attribute, err error) {
```
vedhavyas:
```suggestion
func incrementFundingAttrIDX(model documents.Model) (attr documents.Attribute, err error) {
```

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+
+		attributes = append(attributes, attr)
+
+	}
+
+	return attributes, nil
+}
+
+func (s service) DeriveFromPayload(ctx context.Context, req *clientfundingpb.FundingCreatePayload, identifier []byte) (documents.Model, error) {
+	fd := Data{}
```
vedhavyas: var fd Data

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+			return i.String(), nil
+		}
+
+	}
+
+	return idx, ErrFundingNotFound
+}
+
+func (s service) deriveFundingData(model documents.Model, idx string) (*clientfundingpb.FundingData, error) {
+	data := &clientfundingpb.FundingData{}
```
vedhavyas:
```suggestion
	data := new(clientfundingpb.FundingData)
```

documents/extension/funding/service.go
```diff
@@ -0,0 +1,270 @@
...
+	fd := Data{}
```
vedhavyas:
```suggestion
	var fd Data
```

documents/extension/funding/funding.go
manuelpolzhofer: Yes. I will change all the field type in the next commit.

documents/extension/funding/funding.go
manuelpolzhofer: I think both should be `decimal` :)

apr ->  annual percentage rate
fee -> a value in dollar or ETH

vedhavyas: oops! my bad

documents/extension/funding/service.go
vedhavyas: we could just say `funding` ?

manuelpolzhofer:
@vedhavyas can we merge or you want to take a second look?

documents/extension/funding/funding_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"testing"
+
+	clientfundingpb "github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInitFundingFromData(t *testing.T) {
+	fdc := &clientfundingpb.FundingData{Currency: "eur"}
+	fd := &Data{}
```
vedhavyas:
```suggestion
	fd := new(Data)
```

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
+package funding
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+
```
vedhavyas: White space

documents/extension/funding/service.go
vedhavyas: Any comment on this ?

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil, err
+	}
+
+	// add updated idx
+	attributes = append(attributes, idx)
+
+	types := reflect.TypeOf(data)
+	values := reflect.ValueOf(data)
+	for i := 0; i < types.NumField(); i++ {
+
```
vedhavyas: white space

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+			// set field in client data
+			n := types.Field(i).Name
+
+			v, err := attr.Value.String()
+			if err != nil {
+				return nil, err
+			}
+
+			reflect.ValueOf(data).Elem().FieldByName(n).SetString(v)
+
```
vedhavyas: white space

documents/int256.go
```diff
@@ -129,3 +129,25 @@ func isValidInt256(n big.Int) bool {
 
 	return true
 }
+
+// Add sets i to the sum x+y and returns i
+func (i *Int256) Add(x *Int256, y *Int256) *Int256 {
```
vedhavyas: this should check for the overflow. I think you can re-use the function called `validInt256` and return an error if not

documents/int256.go
```diff
@@ -129,3 +129,25 @@ func isValidInt256(n big.Int) bool {
...
+//    0 if i == y
+//   +1 if i >  y
+//
+func (i *Int256) Cmp(y *Int256) int {
+	return i.v.Cmp(&y.v)
+}
+
+// Inc increments i by one
+func (i *Int256) Inc() *Int256 {
+	i.v.Add(&i.v, big.NewInt(1))
```
vedhavyas: This should also check for overflow and return an error if overflowed

documents/extension/funding/service_test.go
```diff
@@ -0,0 +1,198 @@
...
+		&config.Bootstrapper{},
+		&leveldb.Bootstrapper{},
+		&queue.Bootstrapper{},
+		&ideth.Bootstrapper{},
+		&configstore.Bootstrapper{},
+		anchors.Bootstrapper{},
+		documents.Bootstrapper{},
+		p2p.Bootstrapper{},
+		documents.PostBootstrapper{},
+		// &Bootstrapper{}, // todo add own bootstrapper
```
vedhavyas: Are we doing something with the TODO ?

documents/extension/funding/service_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+
```
vedhavyas: import space

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+
+		attrFundingID, err := attr.Value.String()
+		if err != nil {
+			return idx, err
+		}
+		if attrFundingID == fundingID {
+			return i.String(), nil
+		}
+
```
vedhavyas: White space

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+	data, err := s.deriveFundingData(model, idx)
+	if err != nil {
+		return nil, err
+	}
+
+	return &clientfundingpb.FundingResponse{
+		Header: h,
+		Data:   data,
+	}, nil
+
```
vedhavyas: White space

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+			v, err := attr.Value.String()
+			if err != nil {
+				return nil, err
+			}
+
+			reflect.ValueOf(data).Elem().FieldByName(n).SetString(v)
+
+		}
+
```
vedhavyas: White space

documents/extension/funding/service.go
```diff
@@ -0,0 +1,277 @@
...
+			attrType := types.Field(i).Tag.Get("attr")
+			attr, err := documents.NewAttribute(label, documents.AttributeType(attrType), value)
+			if err != nil {
+				return nil, err
+			}
+
+			attributes = append(attributes, attr)
+		}
+
```
vedhavyas: white space

documents/extension/funding/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	resp, err := h.service.DeriveFundingResponse(model, req.FundingId)
+	if err != nil {
+		apiLog.Error(err)
+		return nil, centerrors.Wrap(err, "could not derive response")
+	}
+
+	return resp, nil
+
```
vedhavyas: White space

documents/extension/funding/handler.go
```diff
@@ -0,0 +1,103 @@
...
+	resp, err := h.service.DeriveFundingResponse(model, req.Data.FundingId)
+	if err != nil {
+		apiLog.Error(err)
+		return nil, errors.NewTypedError(ErrFundingAttr, err)
+	}
+
+	resp.Header.JobId = jobID.String()
+	return resp, nil
+
```
vedhavyas: White space

documents/extension/funding/handler.go
```diff
@@ -0,0 +1,103 @@
+package funding
+
+import (
+	"context"
+
+	"github.com/centrifuge/go-centrifuge/documents"
+	"github.com/centrifuge/go-centrifuge/errors"
+
```
vedhavyas: import space

PR #1017, opened 2019-05-02
 Adjustable p2p delay

mikiquantum:
Based on environmental testing, decided to make it configurable as it response times fluctuate greatly due to the host resources (SSDvsHDD, CPU...)


vedhavyas:
Can we increase the code coverage for this?

p2p/receiver/handler.go
```diff
@@ -55,7 +49,11 @@ func New(
 
 // HandleInterceptor acts as main entry point for all message types, routes the request to the correct handler
 func (srv *Handler) HandleInterceptor(ctx context.Context, peer peer.ID, protoc protocol.ID, msg *pb.P2PEnvelope) (*pb.P2PEnvelope, error) {
-	defer timeutils.EnsureDelayOperation(time.Now(), responseDelay)
+	cfg, err := srv.config.GetConfig()
+	if err != nil {
+		return srv.convertToErrorEnvelop(err)
+	}
+	defer timeutils.EnsureDelayOperation(time.Now(), cfg.GetP2PResponseDelay())
```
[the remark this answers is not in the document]
mikiquantum: Ideally the handler should have its config interface with only the properties that are needed. 
I rather not refactor that now, as we will need to apply this to all the handlers, and potentially services.

mikiquantum:
@vedhavyas I will add some dummy test cases to increase coverage

build/configs/default_config.yaml
```diff
@@ -119,6 +119,10 @@ p2p:
   port: 38202
   # Timeout when opening connections to peers
   connectTimeout: "30s"
+  # minimum delay for every p2p response to avoid timing attacks
+  # adjust based on host resources (SSD, CPU, cores ...)
+  # Look in logs for: "Time consumed by operation" if x=(valueRead * 2) is less than value below, then change responseDelay to x
+  responseDelay: "500ms"
```
lucasvo: Did you measure 500ms? I think that's pretty short as a default. We should also explain why the delay is needed and the privacy implications of setting it too low.

mikiquantum: I did perform a benchmark with an SSD and with HDD, resulting in the above for HDD. For SSD:
```
- Document Exist on a ~1.2M DB:
  - MessageTypeRequestSignature: ~11ms
  - MessageTypeSendAnchoredDoc: ~13ms
  - MessageTypeGetDoc: ~5ms

- Document Does not exist on a ~1.2M DB:
  - MessageTypeRequestSignature: ~11ms
  - MessageTypeSendAnchoredDoc: ~13ms
  - MessageTypeGetDoc: ~1.5ms (*timing attack)
```
This by having 1.2 Million records in DB. You can see the significant jump between both host configurations, it is hard to set a static delay that work for all. That is why is configurable. I explain the delay in the comment, we can add more details to it in the docs.

PR #1023, opened 2019-05-06
Feat/signed attrs

vedhavyas:


documents/validator.go
```diff
@@ -400,6 +400,33 @@ func anchorRepoAddressValidator(anchoredRepoAddr common.Address) Validator {
...
+// TODO(ved): tests
+func attributeValidator(idSrv identity.Service) Validator {
+	return ValidatorFunc(func(_, model Model) (err error) {
+		attrs := model.GetAttributes()
+		for _, attr := range attrs {
+			if attr.Value.Type != AttrSigned {
+				continue
+			}
+
+			signed := attr.Value.Signed
```
lucasvo: I would reuse the signature payload generation code from above here.

documents/validator.go
```diff
@@ -400,6 +400,33 @@ func anchorRepoAddressValidator(anchoredRepoAddr common.Address) Validator {
 	})
 }
 
+// attributeValidator validates the signed attributes.
+// TODO(ved): tests
+func attributeValidator(idSrv identity.Service) Validator {
```
lucasvo: I assume you will still add this to the validator groups?

documents/attribute.go
```diff
@@ -164,20 +188,56 @@ type Attribute struct {
...
+// takes keyLabel, signer identity, signer account, model and value
+// signature payload: sign(identity + docID + docVersion + value + time)
+func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, model Model, value []byte) (attr Attribute, err error) {
 	attrKey, err := AttrKeyFromLabel(keyLabel)
 	if err != nil {
 		return attr, err
 	}
 
-	attrVal, err := AttrValFromString(attrType, value)
+	// TODO(ved): time should be part of the signature
```
lucasvo: I don't think time needs to be added.

documents/attribute.go
```diff
@@ -164,20 +188,56 @@ type Attribute struct {
...
+	var signPayload []byte
```
lucasvo: Can you create a helper method that takes a Signed field struct and constructs the payload from it and then reuse it below?

documents/attribute.go
vedhavyas: yeah. I just realised it

documents/validator.go
vedhavyas: not complete yet

documents/attribute.go
vedhavyas: I still want to think on this more.

documents/validator.go
```diff
@@ -400,6 +400,44 @@ func anchorRepoAddressValidator(anchoredRepoAddr common.Address) Validator {
...
 
+// attributeValidator validates the signed attributes.
+func attributeValidator(repo anchors.AnchorRepository, idSrv identity.Service) Validator {
+	return ValidatorFunc(func(_, model Model) (err error) {
+		ts, err := model.Timestamp()
+		if err != nil {
+			return err
+		}
+
+		attrs := model.GetAttributes()
```
mikiquantum: Would it be interesting to have a getter in the model for all SignedAttributes?

documents/validator.go
```diff
@@ -400,6 +400,44 @@ func anchorRepoAddressValidator(anchoredRepoAddr common.Address) Validator {
...
+			signed := attr.Value.Signed
+			var aid anchors.AnchorID
+			aid, err = anchors.ToAnchorID(signed.DocumentVersion)
+			if err != nil {
+				return err
+			}
+
+			_, ats, erro := repo.GetAnchorData(aid)
+			if erro != nil {
+				// maybe not anchored yet
```
mikiquantum:
```suggestion
				// maybe not anchored yet as we validate signatures a several times before anchoring
```

documents/attribute.go
```diff
@@ -96,14 +118,11 @@ type AttrVal struct {
 	Str       string
 	Bytes     []byte
 	Timestamp *timestamp.Timestamp
+	Signed    Signed
```
mikiquantum: Having a new type, implies that we only support []byte values and we will not need to perform any specific/typed operation over values of type `signed`?

vedhavyas: not sure I understand this

documents/validator.go
vedhavyas: The idea to have a single function to give all attributes and keep the model API simple

PR #1024, opened 2019-05-06
feat: funding GET list

manuelpolzhofer:
- added support for GET list endpoint


documents/extension/funding/handler.go
```diff
@@ -67,9 +66,9 @@ func (h *grpcHandler) Create(ctx context.Context, req *clientfundingpb.FundingCr
 
 	resp.Header.JobId = jobID.String()
 	return resp, nil
-
 }
 
+// Get returns a funding agreement from an existing document
 // Create handles a new funding document extension and adds it to an existing document
```
mikiquantum: remove comment

documents/extension/funding/service.go
```diff
@@ -5,11 +5,11 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/centrifuge/go-centrifuge/utils"
```
mikiquantum: order

documents/extension/funding/handler.go
```diff
@@ -3,11 +3,10 @@ package funding
 import (
 	"context"
 
-	"github.com/centrifuge/go-centrifuge/centerrors"
-	"github.com/centrifuge/go-centrifuge/config"
-	"github.com/centrifuge/go-centrifuge/contextutil"
 	"github.com/centrifuge/go-centrifuge/documents"
 	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/config"
```
mikiquantum: order

documents/extension/funding/service.go
```diff
@@ -264,4 +274,42 @@ func (s service) DeriveFundingResponse(model documents.Model, fundingID string)
...
+
+		i, err := documents.NewInt256("0")
+		if err != nil {
+			return nil, err
+		}
+
+		for ; i.Cmp(lastIdx) != 1; i.Inc() {
+			funding, err := s.deriveFundingData(model, i.String())
+			if err != nil {
+				continue
```
mikiquantum: Why would it fail derive here for a known index?

documents/extension/funding/service.go
```diff
@@ -264,4 +274,42 @@ func (s service) DeriveFundingResponse(model documents.Model, fundingID string)
...
+		return nil, errors.New("failed to derive response: %v", err)
+	}
+	response.Header = h
+
+	fl, err := documents.AttrKeyFromLabel(fundingLabel)
+	if err != nil {
+		return nil, err
+	}
+
+	if model.AttributeExists(fl) {
```
mikiquantum: I think in this case becomes easier to read when:
```
if !model.AttributeExists(fl) {
       return response, nil
}
lastIdx, err := getFundingsLatestIDX(model)
...
return response, nil
```

documents/extension/funding/service.go
manuelpolzhofer: Currently, there is no real case. 

I don't know if we implement `Delete` funding agreement in the future. 

If we do it, it might be simpler to just remove the index instead of reordering the arrays in the map.  In that case, it could be that an index is not existing and we `continue`

We could also throw an error for now.

protobufs/funding/funding.proto
```diff
@@ -30,6 +30,14 @@ service FundingService {
       description: "Get a funding agreement of a document"
     };
   }
+  rpc GetList(GetListRequest) returns (FundingListResponse) {
+    option (google.api.http) = {
+      get: "/documents/{identifier}/extension/funding"
```
vedhavyas: I understand the extension on the codebase. Dont think we need to put it under URL IMO

documents/extension/funding/service.go
```diff
@@ -264,4 +274,44 @@ func (s service) DeriveFundingResponse(model documents.Model, fundingID string)
...
+	if err != nil {
+		return nil, err
+	}
+
+	i, err := documents.NewInt256("0")
+	if err != nil {
+		return nil, err
+	}
+
+	for ; i.Cmp(lastIdx) != 1; i.Inc() {
```
vedhavyas: is this valid? I thought i.Inc() returns error

protobufs/funding/funding.proto
manuelpolzhofer: I don't have a strong opinion about it.  We could also change it.

documents/extension/funding/service.go
manuelpolzhofer: True.

PR #1025, opened 2019-05-06
[ON HOLD] Add transition validation check to signature

### LINKED ISSUE #795: Defining the meaning of a signature on a state (version)
When a node signs a document the meaning of that signature is not entirely clear. There are multiple scenarios where it can mean one of the two things and we currently do not have a way to differentiate the two.

1) I signed off on the transition from `d -> d'`
2) I have received `d'`

There are different scenarios where this can happen:

1) Node added as a new collaborator: if this is simply the first version a node should be able to read, the case *1* would be impossible so the meaning must be *2*. 
2) The node doesn't have the previous version and can therefore only attest to the receipt (e.g. if the data is missing). We can work around this by also sending the previous version and the node can verify that the transition is indeed correct.

However there is a third scenario that is not as simple to deal with:

Let's assume you have `d_1 -> d_2 -> d_3 -> d_4`. A collaborator signed `d_1` but then went offline. In the meantime two new versions (`d_2, d_3`) are published. After that the collaborator comes back online. And another node wants to publish `d_3 -> d_4` What should happen now? Technically it's not enough to verify `d_3 -> d_4` and sign saying the transition is valid. The node should also check `d_2 -> d_3`. If the node wasn't able to check `d_2 -> d_3` because it wasn't given `d_2`, what should it do? If it does, how does it specify what the signature means?

Possible solution: Extend the signature to sign the current signing root along with the last good version. If the node is able to verify it all the way back to `d_1`, that's ideal and if not, it could specify what it has seen.

Acceptance criteria:
- Make sure collaborators can be accessed at a later point in time
- Testworld coverage for different signature meanings

mikiquantum:
Closes #795


Gopkg.toml
```diff
@@ -28,7 +28,8 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 
 [[constraint]]
   name = "github.com/centrifuge/centrifuge-protobufs"
-  revision = "211e4297ee949705422f74820b023903341199fc"
+  source = "github.com/mikiquantum/centrifuge-protobufs"
```
mikiquantum: Will replace soon

documents/service.go
```diff
@@ -194,6 +194,10 @@ func (s service) RequestDocumentSignature(ctx context.Context, model Model, coll
 	if err != nil {
 		return nil, err
 	}
+	// If there is a previous version and we have successfully validated the transition then set the signature flag
+	if old != nil {
+		sig.TransitionValidated = true
```
vedhavyas: this bool should be part of the payload i.e bool + signing_root

documents/service.go
```diff
@@ -190,10 +191,25 @@ func (s service) RequestDocumentSignature(ctx context.Context, model Model, coll
 
 	srvLog.Infof("document received %x with signing root %x", model.ID(), sr)
 
-	sig, err := acc.SignMsg(sr)
+	var transitionValidated bool
+	var transitionFlag int8
+	// If there is a previous version and we have successfully validated the transition then set the signature flag
+	if old != nil {
+		transitionValidated = true
+		transitionFlag = 1
```
vimukthi-git: is there a case where we set the `transitionValidated=false` and `transitionFlag=1` and vice versa? If not, why define the same thing twice?

mikiquantum: You are right, it can be simplified.

mikiquantum:
@vedhavyas @lucasvo @pstehlik holding on to this until we have done the tree changes.

mikiquantum:
closing in favor of: https://github.com/centrifuge/go-centrifuge/pull/1107

PR #1032, opened 2019-05-09
feat: funding sign service + handler

manuelpolzhofer:
**Implemented**
- endpoint for Sign 
- create a signature array attribute in funding attributes (latestIdxField)
-create signature and add to attributes

**In the next PR**
- signature validation (valid, invalid)
- adding signatures to client Get and GetList


documents/extension/funding/service.go
```diff
@@ -403,3 +412,72 @@ func (s service) DeriveFundingListResponse(model documents.Model) (*clientfundin
...
 }
+
+func (s service) createSignAttrs(model documents.Model, idxFunding string, selfDID identity.DID, account config.Account) ([]documents.Attribute, error) {
+	var attributes []documents.Attribute
+	data, err := s.deriveFundingData(model, idxFunding)
+	if err != nil {
+		return nil, err
+	}
+
+	signMsg, err := json.Marshal(data)
```
manuelpolzhofer: We agreed we would like to sign all fields of the funding struct.
Currently, I sign it as a JSON.  Should we use another format?

manuelpolzhofer: in the ideal case, we generate a merkle root hash of the funding fields

documents/extension/funding/service.go
```diff
@@ -2,9 +2,14 @@ package funding
...
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 	"strings"
 
+	"github.com/centrifuge/go-centrifuge/config"
+	"github.com/centrifuge/go-centrifuge/contextutil"
+	"github.com/centrifuge/go-centrifuge/identity"
+
```
vedhavyas: import spacing

documents/extension/funding/service_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/centrifuge/centrifuge-protobufs/gen/go/coredocument"
+	"github.com/centrifuge/go-centrifuge/contextutil"
+
```
vedhavyas: import spacing

documents/extension/funding/service_test.go
```diff
@@ -183,6 +186,90 @@ func TestDeriveFundingListResponse(t *testing.T) {
...
+	srv, model, lastFundingId := setupFundingsForTesting(t, fundingAmount)
+
+	// add signature
+	acc := &mockAccount{}
+	acc.On("GetIdentityID").Return(utils.RandomSlice(20), nil)
+	// success
+	signature := utils.RandomSlice(32)
+	acc.On("SignMsg", mock.Anything).Return(&coredocumentpb.Signature{Signature: signature}, nil)
+	ctx, err := contextutil.New(context.Background(), acc)
+	assert.Nil(t, err)
```
vedhavyas:
```suggestion
	assert.NoError(t, err)
```

protobufs/funding/funding.proto
```diff
@@ -24,14 +24,23 @@ service FundingService {
...
     };
   }
-  rpc Get(GetRequest) returns (FundingResponse) {
+  rpc Sign(Request) returns (FundingResponse) {
+    option (google.api.http) = {
+      post: "/document/{identifier}/funding/{funding_id}/sign"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      description: "Updates a funding agreement in a document"
```
vedhavyas:
```suggestion
      description: "Signs a funding aggrement"
```

documents/extension/funding/service.go
```diff
@@ -95,17 +105,17 @@ func getFundingsLatestIDX(model documents.Model) (idx *documents.Int256, err err
 
 }
 
-func incrementFundingAttrIDX(model documents.Model) (attr documents.Attribute, err error) {
-	key, err := documents.AttrKeyFromLabel(fundingLabel)
+func incrementArrayAttrIDX(model documents.Model, arrayLabel string) (attr documents.Attribute, err error) {
```
vedhavyas: I think the previous function name makes more sense IMO instead making this array specific

documents/extension/funding/service.go
vedhavyas: I would love it if we don't have to rely on internal representations for signature payload generations even though its json.

documents/extension/funding/service.go
manuelpolzhofer: I am re-using the same method for the signatureArray

documents/extension/funding/service.go
manuelpolzhofer: What alternative do you have in your mind?
Adding the definition to the protobuf repo or sorting the fields?

vedhavyas: yeah, something like that. That would be encoder agnostic IMO

documents/extension/funding/funding.go
```diff
@@ -31,3 +31,17 @@ func (f *Data) initFundingFromData(data *clientfundingpb.FundingData) {
...
+
+func (f *Data) getClientData() *clientfundingpb.FundingData {
+	clientData := new(clientfundingpb.FundingData)
+
+	types := reflect.TypeOf(*f)
+	values := reflect.ValueOf(*f)
+	for i := 0; i < types.NumField(); i++ {
+		n := types.Field(i).Name
+		v := values.FieldByName(n).Interface().(string)
+		reflect.ValueOf(clientData).Elem().FieldByName(n).SetString(v)
```
mikiquantum: I am not sure how safe is to assume that all our API fields are strings. They could be wrapped as well as complex structures of strings. Let's make sure we add a comment about the limitations of this converter.

documents/extension/funding/service.go
mikiquantum: I was going to ask, can we generate proofs about of a custom attribute? We should be able right? JSON will not be easily verifiable onchain. One thing is verifying signature (which you need the `concat(fundingField1+fundingField2...)` or `fundingRoot`) and another verifying one single field of the fundingrequest as a proof.
I agree that we should have this document as a protobuf (otherwise we cannot create proofs), but then I am confused if it should be defined in `centrifuge-protobufs` repo or in `go-centrifuge` as it is an extension/property of another document. Thoughts?

mikiquantum:
@manuelpolzhofer in addition can you upgrade the swagger version to `0.0.5`?

documents/extension/funding/service.go
manuelpolzhofer: If we are adding it to protobuf I think we should use the protobuf repo

manuelpolzhofer: @mikiquantum  I discussed with Ved, We will keep it for now as JSON but I will schedule a meeting where we discuss a long term solution for the next iteration. Maybe the end of this week or next week.

PR #1035, opened 2019-05-13
feat: funding GET with signatures

manuelpolzhofer:
please review #1032 first

- added to funding response
- added signature verification


documents/attribute.go
```diff
@@ -238,7 +238,8 @@ func NewSignedAttribute(keyLabel string, identity identity.DID, account config.A
 	}, nil
 }
 
-func attributeSignaturePayload(did, id, version, value []byte) []byte {
+// AttributeSignaturePayload creates the payload for signing an attribute
+func AttributeSignaturePayload(did, id, version, value []byte) []byte {
```
manuelpolzhofer: I made the func public but I actually would like to have it as a method of signAttr.

What do you think?

config/configstore/model.go
```diff
@@ -460,6 +460,11 @@ func (acc *Account) SignMsg(msg []byte) (*coredocumentpb.Signature, error) {
 	}, nil
 }
 
+// VerifyMsg verifies a message with a given public key
+func (acc *Account) VerifyMsg(msg []byte, signature []byte, pk []byte) bool {
```
manuelpolzhofer: Not needed. I will remove it.

documents/attribute.go
vedhavyas: Having a function would be preferable. any specific need for a method on AttrSign

documents/extension/funding/boostrapper_test.go
```diff
@@ -3,13 +3,17 @@
 package funding
 
 import (
+	"fmt"
 	"testing"
 
+	"github.com/centrifuge/go-centrifuge/identity"
+	"github.com/centrifuge/go-centrifuge/testingutils/commons"
+
```
vedhavyas: import space

documents/extension/funding/boostrapper_test.go
```diff
@@ -28,10 +32,17 @@ func TestBootstrapper_Bootstrap(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "document service not initialised")
 
-	// missing token registry
+	// missing identity service
 	ctx[documents.BootstrappedDocumentService] = new(mockService)
 	err = b.Bootstrap(ctx)
 	assert.Error(t, err)
+	fmt.Println(err)
```
vedhavyas: forgot to remove this?

documents/extension/funding/handler.go
```diff
@@ -188,7 +225,7 @@ func (h *grpcHandler) GetListVersion(ctx context.Context, req *clientfundingpb.G
 		return nil, documents.ErrDocumentNotFound
 	}
 
-	resp, err := h.service.DeriveFundingListResponse(model)
+	resp, err := h.service.DeriveFundingListResponse(ctxHeader, model)
```
vedhavyas: better rename ctxHeader to ctx no?
since both are of type context.Context

documents/extension/funding/service.go
```diff
@@ -5,6 +5,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/centrifuge/go-centrifuge/identity"
+
```
vedhavyas: import space

documents/extension/funding/service_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/centrifuge/go-centrifuge/testingutils/config"
```
vedhavyas: import spacing

documents/extension/funding/signatures.go
```diff
@@ -0,0 +1,189 @@
...
+	// example: "funding_agreement[2].signatures[4]"
+	sFieldLabel := generateLabel(generateLabel(fundingFieldKey, idxFunding, "")+fundingSignaturesFieldKey, attrIdx.Value.Int256.String(), "")
+
+	attrSign, err := documents.NewSignedAttribute(sFieldLabel, selfDID, account, model, signMsg)
+	if err != nil {
+		return nil, err
+	}
+
+	attributes = append(attributes, attrSign)
+
```
vedhavyas: white space

documents/extension/funding/signatures.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+func (s service) createSignAttrs(model documents.Model, idxFunding string, selfDID identity.DID, account config.Account) ([]documents.Attribute, error) {
+	var attributes []documents.Attribute
+	data, err := s.deriveFundingData(model, idxFunding)
+	if err != nil {
+		return nil, err
+	}
+
+	signMsg, err := json.Marshal(data)
```
vedhavyas: any update on using a merkle tree for this?

documents/extension/funding/signatures.go
```diff
@@ -0,0 +1,189 @@
...
+
+	err = model.AddAttributes(attributes...)
+	if err != nil {
+		return nil, err
+	}
+
+	return model, nil
+}
+
+func (s service) validateSignedFundingVersion(ctx context.Context, identifier []byte, fundingID string, signAttr documents.Attribute) (*clientfundingpb.FundingSignature, error) {
```
vedhavyas: Why are we validating this here?
These validations are already part of the document consensus. We do not need that here.

documents/extension/funding/signatures.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, ErrFundingNotFound
+	}
+
+	valueSignedVersion, err := json.Marshal(signedFunding)
+	if err != nil {
+		return nil, ErrJSON
+	}
+
+	payload := documents.AttributeSignaturePayload(did[:], identifier, signAttr.Value.Signed.DocumentVersion, valueSignedVersion)
+	err = s.idSrv.ValidateSignature(did, address32Bytes[:], signAttr.Value.Signed.Signature, payload, time.Now())
```
vedhavyas: we cannot use `time.now` here.

protobufs/funding/funding.proto
```diff
@@ -120,3 +129,15 @@ message FundingData {
...
 }
+
+message FundingResponseData {
+  FundingData funding = 1;
+  repeated FundingSignature signatures = 2;
+}
+
+message FundingSignature {
+  bool valid = 1;
+  bool outdated_signature = 2;
```
vedhavyas: what is an outdated signature here ?

protobufs/funding/funding.proto
```diff
@@ -120,3 +129,15 @@ message FundingData {
...
   string payment_details_id = 12;
 }
+
+message FundingResponseData {
+  FundingData funding = 1;
+  repeated FundingSignature signatures = 2;
+}
+
+message FundingSignature {
+  bool valid = 1;
```
vedhavyas: we should not add any invalid signatures, IMO

documents/extension/funding/signatures.go
manuelpolzhofer: If the value of the signed attribute would be `value = "i am something"` the document consensus validation would say everything is fine.

But we actually want the value to be a specific set of fields. (=funding agreement)

Here we are checking if you put the correct fields in the value.

protobufs/funding/funding.proto
manuelpolzhofer: let's say:

I have a funding with
 ```json
funding =  {"apr": 0.5}
```
If I would add a signature and afterwards change the apr value. 
`outdatedSignature=true`

File: documents/extension/funding/signatures.go (hunk 10 above)
Comment (manuelpolzhofer) [reply documents/extension/funding/signatures.go]: Instead of fully validating the signature I also could only check the value field. Relying on the assumption only correctly signed attributes are added.

File: protobufs/funding/funding.proto (hunk 12 above)
Comment (manuelpolzhofer) [reply protobufs/funding/funding.proto]: otherwise, it is `false`. It is basically tagging outdated signatures

File: protobufs/funding/funding.proto [hunk 14]
```diff
@@ -87,12 +87,12 @@ message FundingUpdatePayload {
...
 message FundingResponse {
   document.ResponseHeader header = 1;
-  FundingData data = 2;
+  FundingResponseData data = 2;
 }
 
 message FundingListResponse {
   document.ResponseHeader header = 1;
-  repeated FundingData list = 2;
+  repeated FundingResponseData list = 2;
```
mikiquantum: `FundingListResponse` that has a property `list` sounds redundant to me. Can we follow the same pattern and keep the name as `data`.

protobufs/funding/funding.proto
mikiquantum: Do we still need to keep this `signature_outdated` field? Vs validating that if a an attribute is modified we always regenerate the signature

documents/attribute.go
```diff
@@ -238,6 +238,7 @@ func NewSignedAttribute(keyLabel string, identity identity.DID, account config.A
 	}, nil
 }
 
+// AttributeSignaturePayload creates the payload for signing an attribute
```
mikiquantum: it is fine to leave the comment but better lower case it.

documents/extension/funding/boostrapper_test.go
```diff
@@ -28,7 +28,7 @@ func TestBootstrapper_Bootstrap(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "document service not initialised")
 
-	// missing token registry
+	// missing identity service
```
mikiquantum: looks like the comment renaming doesnt match the error below.

documents/extension/funding/signatures.go
```diff
@@ -0,0 +1,182 @@
...
+	value, err := json.Marshal(funding)
+	if err != nil {
+		return false, ErrJSON
+	}
+	return utils.IsSameByteSlice(value, signAttr.Value.Signed.Value), nil
+}
+
+func (s service) validateSignedFundingVersion(ctx context.Context, identifier []byte, fundingID string, signAttr documents.Attribute) (*clientfundingpb.FundingSignature, error) {
+	did := signAttr.Value.Signed.Identity
+	signedDocVersion, err := s.Service.GetVersion(ctx, identifier, signAttr.Value.Signed.DocumentVersion)
```
mikiquantum: I believe we are loading again the model, knowing that the parent invoker had it loaded?

manuelpolzhofer: Resolved.

testworld/funding_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	docIdentifier := getDocumentIdentifier(t, res)
+
+	params := map[string]interface{}{
+		"document_id": docIdentifier,
+		"currency":    "USD",
+	}
+	getDocumentAndCheck(t, alice.httpExpect, alice.id.String(), typeInvoice, params, true)
+	getDocumentAndCheck(t, bob.httpExpect, bob.id.String(), typeInvoice, params, true)
+	fmt.Println("Host test success")
```
vedhavyas: Forgot to remove ?

documents/extension/funding/service_test.go
```diff
@@ -292,12 +209,12 @@ func TestService_DeriveFromUpdatePayload(t *testing.T) {
 	model, err = srv.DeriveFromUpdatePayload(context.Background(), p2, utils.RandomSlice(32))
 	assert.NoError(t, err)
 
-	response, err := srv.DeriveFundingListResponse(model)
-	assert.Equal(t, 1, len(response.List))
-	assert.Equal(t, p2.Data.Fee, response.List[0].Fee)
+	response, err := srv.DeriveFundingListResponse(context.Background(), model)
```
vedhavyas: check error pls

manuelpolzhofer:
I will close this one. All changes combined in #1041

Issue #715, opened 2019-02-02
Remove access to txpool by relying on eth_getTransactionCount for pending txs


mikiquantum:
go-ethereum has fixed the issue here: https://github.com/ethereum/go-ethereum/issues/2880

We can go ahead and test that fix out, and if successful replace the txpool access logic from the code. In addition, the txpool config flag will not be needed.

This will allow us to integrate with our ethereum providers that implement standard rpc calls.
Parent topic: https://github.com/centrifuge/go-centrifuge/issues/403

#### Testing
- The change needs to be tested on all 3 centrifuge testnets russianhill, bernelheights, dogpatch using Testworld. 

#### Acceptance Criteria
Tests pass on all testnets

mikiquantum:
We can remove the access to the txpool completely so it is compatible with infura api set:
- This will work on a self hosted node (single node access)
- Need to verify how that would work for infura infra type. The reason for it is that behind the infura api, a read request can go to a different node than the one that the write request can go. In which case, I am not sure how infura can make the txpool (internal) available for all nodes within the cluster, so it is is possible that a retrial mechanism needs to remain (as is right now)

I believe that we can tackle this and remove access to txpool directly and remove config flag and leverage the fixed api transactionCount(pending) for it.

PR #1036, opened 2019-05-13
Remove txpool access

vimukthi-git:
Issue Link: #715

### DIFF
--- a/ethereum/geth_client.go
+++ b/ethereum/geth_client.go
@@ -5,11 +5,11 @@ import (
 	"math/big"
 	"net/url"
 	"reflect"
-	"sort"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/ethereum/go-ethereum"
 
 	"github.com/centrifuge/go-centrifuge/config"
 	"github.com/centrifuge/go-centrifuge/errors"
@@ -35,7 +35,6 @@ type Config interface {
 	GetEthereumAccount(accountName string) (account *config.AccountConfig, err error)
 	GetEthereumIntervalRetry() time.Duration
 	GetEthereumMaxRetries() int
-	GetTxPoolAccessEnabled() bool
 	GetEthereumContextReadWaitTimeout() time.Duration
 }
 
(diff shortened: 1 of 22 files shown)


ethereum/mock_ethclient.go
```diff
@@ -1,16 +1,14 @@
 // +build integration unit
 
-package testingcommons
```
vimukthi-git: had to change this because of dep cycles.

ethereum/geth_client.go
```diff
@@ -5,12 +5,12 @@ import (
...
 	"net/url"
 	"reflect"
-	"sort"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum"
+
```
vedhavyas: import spacing

mikiquantum:
@vedhavyas In systems like infura you will never have certainty of the transaction queue nor pending transactions (unless they build up a common pool per API account). This fix focuses on not relaying on the internal txpool as it is not exposed in these type of clients. We can talk later about how parity does it, but it needs to be compatible with all clients. If this is still not clear let's sync offline.

ethereum/geth_client.go
```diff
@@ -373,59 +382,17 @@ type callContexter interface {
...
-// else we increment the greater of current nonce or the nonce derived from txpool
-func (gc *gethClient) incrementNonce(opts *bind.TransactOpts, txpoolAccessEnabled bool, noncer noncer, cc callContexter) error {
-	// check if the txpool access is enabled
-	if !txpoolAccessEnabled {
-		log.Warningf("Ethereum Client doesn't support txpool API, may cause transactions failures")
-		opts.Nonce = nil
-		return nil
-	}
-
+func (gc *gethClient) incrementNonce(opts *bind.TransactOpts) error {
```
mikiquantum: I would rename it to:
```suggestion
func (gc *gethClient) setNonce(opts *bind.TransactOpts) error {
```

vimukthi-git:
@mikiquantum as agreed online, anchoring concurrency test is a good test of this and we need to use contract methods in eth package to add a test like you mention. Therefore I will not add a test here.

PR #1037, opened 2019-05-14
Ping API

vedhavyas:
This PR includes the skeleton for swagger and API. The swagger is generated from the comments. Health package will define the health-specific APIs and router will define the General Swagger information.

Once this is in, we can start to migrate remaining endpoints without any changes specific to the swagger. The new APIs will be added from the comments itself.

PR intends to avoid noise as much as possible
Changes:
- Added new package `httpapi`. This will replace `api` once we migrate from protobufs
- `gen-swagger` creates a new `api.swagger.json` and will include the new APIs to final swagger.json. With new endpoints coming, this command will automatically pick those and add them to the final swagger.
- removed health check protobufs and reverse gateways logic
- fully compatible with old APIs


Makefile
```diff
@@ -68,6 +71,7 @@ vendorinstall: ## Installs all protobuf dependencies with go-vendorinstall
 	go-vendorinstall github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
 	go-vendorinstall golang.org/x/tools/cmd/goimports
 	go get -u github.com/jteeuwen/go-bindata/...
+	go get -u github.com/swaggo/swag/cmd/swag
```
mikiquantum: I would lock the version in the toml, so we do not have undesired behaviors with upgrades.

api/server.go
```diff
@@ -78,7 +79,10 @@ func (c apiServer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr cha
 		mux.Handle("/debug/", http.DefaultServeMux)
 	}
 
-	mux.Handle("/", gwmux)
+	// if we dont find the route in the mux, we will redirect it to gmux
```
mikiquantum: How would an interceptor set up for authentication look like?

httpapi/health/health_test.go
```diff
@@ -0,0 +1,38 @@
...
+type mockConfig struct{}
+
+func (mockConfig) GetNetworkString() string {
+	return "test network"
+}
+
+func TestHandler_Ping(t *testing.T) {
+	h := handler{mockConfig{}}
+	r := httptest.NewRequest("GET", "/ping", nil)
+	res := httptest.NewRecorder()
```
mikiquantum: 👍

httpapi/router.go
```diff
@@ -0,0 +1,36 @@
...
+// @schemes http
+func Router(config Config) *chi.Mux {
+	r := chi.NewRouter()
+
+	// add middlewares
+	r.Use(middleware.Recoverer)
+	r.Use(middleware.DefaultLogger)
+
+	// health check
+	health.Register(r, config)
```
mikiquantum: Here we would centralize the registration of all services, right?

httpapi/health/health.go
```diff
@@ -0,0 +1,47 @@
...
+
+// @summary Health check for Node
+// @description returns node version and network
+// @id ping
+// @produce json
+// @success 200 {object} health.Pong
+// @router /ping [get]
+func (h handler) Ping(w http.ResponseWriter, r *http.Request) {
+	render.Status(r, http.StatusOK)
+	render.JSON(w, r, Pong{
```
mikiquantum: How much control we have over the JSON marshaller?

api/server.go
vedhavyas: For now the, since ping doesn't have any auth there is no interceptor and for the remaining, it will happen as part of the `gwmux`. Once we move to this framework, we will have a middleware just like the `middleware.Recovery` i have defined which exclude `ping`

httpapi/router.go
vedhavyas: yeap. Each one follows as we add more and this is locations where we update the swagger API version through comments. For now, we will fallback to the one we defined on the swagger.js

httpapi/health/health.go
vedhavyas: render.JSON uses std lib's json.Marshaller. If we need to provide a custom marshaller, we may choose to write our own func like render.JSON or give them a PR to allow custom marshaller

api/server.go
mikiquantum: Yep, I understand that in this particular endpoint we exclude the `auth`. My question is more on looking ahead of this endpoint into a `create` for example. That requires authentication, How would it look like in the code? Can you add a simple snippet here as a comment?

httpapi/router.go
mikiquantum: That part is still not clear enough, how swagger generates the json out of comments, can you add an example here?

vedhavyas: This is through declarative comments: https://github.com/swaggo/swag#declarative-comments-format

api/server.go
vedhavyas: Not sure I understand what you are asking here :-D

PR #1060, opened 2019-05-21
Fix multi account wrong local collaborator bug

vimukthi-git:
thanks @rdinicut for your QA effort :D

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -26,6 +26,11 @@ func (s *peer) SendAnchoredDocument(ctx context.Context, receiverID identity.DID
 		return nil, err
 	}
 
+	selfDID, err := contextutil.AccountDID(ctx)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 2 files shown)


p2p/client.go
```diff
@@ -38,7 +43,7 @@ func (s *peer) SendAnchoredDocument(ctx context.Context, receiverID identity.DID
 		if err != nil {
 			return nil, err
 		}
-		return h.SendAnchoredDocument(localCtx, in, receiverID)
+		return h.SendAnchoredDocument(localCtx, in, selfDID)
```
manuelpolzhofer: Let's say I have 5 collaborators on a document. We loop over the 5 collaborators on a higher level but this method is called to actually send it to each of the receivers?

Or, I am wrong?

manuelpolzhofer: A got it. It is the local account case.

Issue #1040, opened 2019-05-16
Active find peer by PeerID when connection request


mikiquantum:
Currently we cannot get more than 20 peers connected at a time and we are no leveraging full node discovery. We need to proactively look for peer when a client request is triggered.

#### Acceptance Criteria
Test discovery works for more than n peer result set by making use of distributed routing table.

PR #1066, opened 2019-05-23
Fix for DHT KValue=20 issue

vimukthi-git:
Issue Link: #1040

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -207,16 +207,6 @@ func (s *peer) getPeerID(ctx context.Context, id identity.DID) (libp2pPeer.ID, e
 		s.host.Peerstore().AddAddrs(peerID, pinfo.Addrs, pstore.PermanentAddrTTL)
 	}
 
-	//if !s.disablePeerStore {
-	//	// Decapsulate the /ipfs/<peerID> part from the target
-	//	// /ip4/<a.b.c.d>/ipfs/<peer> becomes /ip4/<a.b.c.d>
-	//	targetPeerAddr, _ := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s", pid))
-	//	targetAddr := ipfsAddr.Decapsulate(targetPeerAddr)
-	//	// We have a peer ID and a targetAddr so we add it to the peer store
-	//	// so LibP2P knows how to contact it
-	//	s.host.Peerstore().AddAddr(peerID, targetAddr, pstore.PermanentAddrTTL)
-	//}
-
 	return peerID, nil
 }
 
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -52,7 +52,7 @@ func (s *peer) SendAnchoredDocument(ctx context.Context, receiverID identity.DID
 	}
 
 	// this is a remote account
-	pid, err := s.getPeerID(receiverID)
+	pid, err := s.getPeerID(ctx, receiverID)
 	if err != nil {
 		return nil, err
 	}
@@ -126,7 +126,7 @@ func (s *peer) GetDocumentRequest(ctx context.Context, requesterID identity.DID,
 	}
 
 	// this is a remote account
-	pid, err := s.getPeerID(requesterID)
+	pid, err := s.getPeerID(ctx, requesterID)
 	if err != nil {
 		return nil, err
 	}
@@ -167,8 +167,8 @@ func (s *peer) GetDocumentRequest(ctx context.Context, requesterID identity.DID,
 	return r, nil
 }
 
-// OpenClient returns P2PServiceClient to contact the remote peer
-func (s *peer) getPeerID(id identity.DID) (libp2pPeer.ID, error) {
+// getPeerID returns peerID to contact the remote peer
+func (s *peer) getPeerID(ctx context.Context, id identity.DID) (libp2pPeer.ID, error) {
 	lastB58Key, err := s.idService.CurrentP2PKey(id)
 	if err != nil {
 		return "", errors.New("error fetching p2p key: %v", err)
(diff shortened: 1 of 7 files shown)


p2p/client.go
```diff
@@ -191,15 +191,32 @@ func (s *peer) getPeerID(id identity.DID) (libp2pPeer.ID, error) {
...
+		}
+
 		// We have a peer ID and a targetAddr so we add it to the peer store
-		// so LibP2P knows how to contact it
-		s.host.Peerstore().AddAddr(peerID, targetAddr, pstore.PermanentAddrTTL)
+		// so LibP2P knows how to contact it (this call might be redundant)
+		s.host.Peerstore().AddAddrs(peerID, pinfo.Addrs, pstore.PermanentAddrTTL)
 	}
 
+	//if !s.disablePeerStore {
```
vimukthi-git: old code commented out

p2p/server.go
```diff
@@ -124,8 +136,53 @@ func (s *peer) InitProtocolForDID(DID *identity.DID) {
...
+	log.Info("Announcing ourselves...")
+	tctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	if err := s.dht.Provide(tctx, cidPref, true); err != nil {
+		// Important to keep this as Non-Fatal error, otherwise it will fail for a node that behaves as well as bootstrap one
+		log.Infof("Error: %s\n", err.Error())
+	}
+	cancel()
+
+	// Now, look for others who have announced
+	log.Info("Searching for other peers ...")
```
vimukthi-git: since this block depends on KValue, its not really worth keeping.

p2p/server.go
```diff
@@ -124,8 +136,53 @@ func (s *peer) InitProtocolForDID(DID *identity.DID) {
 	s.mes.Init(p)
 }
 
+func (s *peer) runDHT(ctx context.Context, bootstrapPeers []string) error {
+	dht.KValue = 1
```
vimukthi-git: remove after testing

p2p/server.go
```diff
@@ -175,13 +229,14 @@ func makeBasicHost(priv crypto.PrivKey, pub crypto.PubKey, externalIP string, li
 	}
 
 	opts := []libp2p.Option{
+		libp2p.Peerstore(ps),
```
vimukthi-git: the peerstore created above weren't being used before :D

p2p/client.go
```diff
@@ -191,15 +191,32 @@ func (s *peer) getPeerID(id identity.DID) (libp2pPeer.ID, error) {
 	}
 
 	if !s.disablePeerStore {
-		// Decapsulate the /ipfs/<peerID> part from the target
-		// /ip4/<a.b.c.d>/ipfs/<peer> becomes /ip4/<a.b.c.d>
-		targetPeerAddr, _ := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s", pid))
-		targetAddr := ipfsAddr.Decapsulate(targetPeerAddr)
+		nc, err := s.config.GetConfig()
```
vimukthi-git: fix for the KValue=20 problem

vimukthi-git: commenting this and uncommenting the prev code below will fail the travis build because an address could not be found.

mikiquantum:
I was able to constantly reproduce the problem for both approaches, using Bootstrap and Provide:
```
Provide->
  - Consistently getting error when KValue set to 2 and 5 nodes up (1 Bootstrap + 4 nodes)
    - Eve (last one to start) trying to send msg to Bob -> Fails
  - Even adding FindPeer before client call, doesnt work, apparently the routing table is not populated enough due to the small kvalue set
Bootstrap->
  - Harder to reproduce for KValue=2|1
  - Routing table looks more complete
  - When some peer was not included at bootstrap in the routing table, similar problem as above.
    - For smaller values of KValue, the bucket size will be smaller, therefore is possible that will not be able to store the optimum nearest peers

In the near future we need to add a retrial for at least 10 mins on 1 minute increments. As the bootstrap job runs every 5 mins and can get new entries populated.
```

I would replace the current `provide` with bootstrap, as shown in the code comment.

p2p/server.go
```diff
@@ -124,8 +136,37 @@ func (s *peer) InitProtocolForDID(DID *identity.DID) {
...
+	for _, addr := range bootstrapPeers {
+		iaddr, _ := ipfsaddr.ParseString(addr)
+		pinfo, _ := pstore.InfoFromP2pAddr(iaddr.Multiaddr())
+		if err := s.host.Connect(ctx, *pinfo); err != nil {
+			log.Info("Bootstrapping to peer failed: ", err)
+		}
+	}
+
+	// Using the sha256 of our "topic" as our rendezvous value
+	cidPref, _ := cid.NewPrefixV1(cid.Raw, mh.SHA2_256).Sum([]byte("centrifuge-dht"))
```
mikiquantum: Replace this block with:
```
err := s.dht.Bootstrap(ctx)
	if err != nil {
		log.Errorf("Bootstrap Error: %s", err.Error())
		return err
	}
```

p2p/server.go
```diff
@@ -93,7 +94,18 @@ func (s *peer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr chan<-
 	}
 
 	// Start DHT and properly ignore errors :)
-	_ = runDHT(ctx, s.host, nc.GetBootstrapPeers())
+	_ = s.runDHT(ctx, nc.GetBootstrapPeers())
```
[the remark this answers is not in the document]
vimukthi-git: this error should be ignored as the comment says..

p2p/client.go
```diff
@@ -191,15 +191,32 @@ func (s *peer) getPeerID(id identity.DID) (libp2pPeer.ID, error) {
...
+	//if !s.disablePeerStore {
+	//	// Decapsulate the /ipfs/<peerID> part from the target
+	//	// /ip4/<a.b.c.d>/ipfs/<peer> becomes /ip4/<a.b.c.d>
+	//	targetPeerAddr, _ := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s", pid))
+	//	targetAddr := ipfsAddr.Decapsulate(targetPeerAddr)
+	//	// We have a peer ID and a targetAddr so we add it to the peer store
+	//	// so LibP2P knows how to contact it
+	//	s.host.Peerstore().AddAddr(peerID, targetAddr, pstore.PermanentAddrTTL)
+	//}
+
```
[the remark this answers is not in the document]
vimukthi-git: Yes, I was going to remove before merge. this pr was created only as a test at first.

p2p/server.go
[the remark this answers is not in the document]
vedhavyas: If we are not going to use it, then lets log it like we are doing in other cases. Ignoring errors would cause unexpected issues at a later time :)

Issue #1043, opened 2019-05-16
Define migration file item to update job keys to hex


mikiquantum:
We will have it as part of the GO code base with static go migration files that are compiled.
- Takes DB file paths
- Run migration file code and returns with status code back to wrapper.

PR #1067, opened 2019-05-23
Migrate/keys

mikiquantum:
Closes: #1043


migration/files/00Initial_test.go
```diff
@@ -2,36 +2,22 @@ package files
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 	"testing"
 
-	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/centrifuge/go-centrifuge/migration/utils"
```
vedhavyas: why a new package name, can we keep them as part of migrations package if no other package is using these ?

migration/utils/utils.go
```diff
@@ -0,0 +1,38 @@
+package migrationutils
+
+import (
+	"crypto/rand"
+	"os"
+	"path/filepath"
+)
+
+// RandomSlice returns a randomly filled byte array with length of given size
+func RandomSlice(size int) (out []byte) {
```
vedhavyas: we do have another function in utils right ?

migration/files/00Initial_test.go
mikiquantum: I wanted to have a folder division for the migration files vs the migration wrapper. I am less interested on the package naming though, I am open to move it to migration or migrationfiles.

migration/utils/utils.go
mikiquantum: Yes we do, the main reason to duplicate is that we do not want to add any dependencies if possible with the code base, as we could extend this package and make it more configurable and its own repository for go-(leveldb-otherdb) migrations

Issue #876, opened 2019-03-21
Byzantine fault behaviour tests 1


vimukthi-git:
This ticket is to implement 3 testcases with our [`testworld` p2p simulation frame work](https://github.com/centrifuge/go-centrifuge/tree/develop/testworld). 

Refer `testworld/README.md` for env setup and a small guide on how to run. Define your tests in `testworld/signature_test.go`. This test involves some research in how to simulate abnormal behaviour by the new host Mallory. A possible way is to define Mallory in https://github.com/centrifuge/go-centrifuge/blob/develop/testworld/park.go#L30 but after initialisation replace her `p2p/receiver/handler.go` with a different implementation (inject with key `bootstrap.BootstrappedPeer` in https://github.com/centrifuge/go-centrifuge/blob/develop/p2p/bootstrapper.go#L44).

* Alice creates a document with Mallory and Mallory signs it with a key that is not in collaborators

* Alice creates a document with Mallory and Mallory returns a signature for a different signing root

* Eve creates a document with Alice but calculates the signing root incorrectly and sends it to Alice.
 

This issue has a gitcoin bounty attached to it. If you'd like to start working on it, feel free to reach out to me (lucas@centrifuge.io) directly, join our slack channel #bounties on our [slack](
https://centrifuge.io/slack/).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 180.0 DAI (180.0 USD @ $1.0/DAI)  attached to it as part of the Centrifuge fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/centrifuge/go-centrifuge/876/2690).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/centrifuge/go-centrifuge/876/2690).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $53,555.09 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 6 days from now.
Please review their action plans below:


[...]

PR #1077, opened 2019-06-03
Byzantine fault behaviour tests 1 #876

zyfrank:
Issue Link: <Github link>


testworld/park.go
```diff
@@ -293,6 +297,14 @@ func (h *host) init() error {
 	if err != nil {
 		return err
 	}
+
+	if h.name == "Mallory" {
+		malloryDocMockSrv := new(mockdoc.MockService)
```
vimukthi-git: Are you sure this is the actual doc service that gets called for the host? 
Seems though the underlying idea is correct, this would not inject the mock doc service correctly because the p2p bootstrapper you create below is not added to the host anywhere. 
My idea is that you can take the `h.bootstrappedCtx[bootstrap.BootstrappedPeer]` struct(type `p2p.peer`) here and set its `handlerCreator` field to something you control during the test.  It gets set for the host originally at https://github.com/centrifuge/go-centrifuge/blob/develop/p2p/bootstrapper.go#L44

testworld/park.go
```diff
@@ -37,6 +39,8 @@ var hostConfig = []struct {
 	{"Charlie", 8086, 38206, true},
 	{"Kenny", 8087, 38207, false},
 	{"Eve", 8088, 38208, false},
+	// Mallory has a mock document.Serivce to facilitate some Byzantine test
```
vimukthi-git: 👍

testworld/park.go
zyfrank: Hi @vimukthi-git 
you can try comment line 54 ,57 59 of signature_test.go, following is my result
image

zyfrank: init() function is called before live()

testworld/signature_test.go
```diff
@@ -17,12 +17,55 @@ import (
...
+	malloryDocMockSrv := mallory.host.bootstrappedCtx[documents.BootstrappedDocumentService].(*mockdoc.MockService)
+
+	// so when got request on signature of document, mocking documents.Service of Mallory return a random signature
+	malloryDocMockSrv.On("RequestDocumentSignature", mock.Anything, mock.Anything, mock.Anything).Return(sig, nil).Once()
+
+	malloryDocMockSrv.On("DeriveFromCoreDocument", mock.Anything).Return(dm, nil).Once()
+
+	//TODO
+	signatures, signatureErrors, err := alice.host.p2pClient.GetSignaturesForDocument(actxh, dm)
+	//seems to me, following should get signature verification errors but it is not.  Currenly p2p/client.go just do validateSignatureResp verification (very simple DID verification?), is this the right behavior?
```
mikiquantum: @zyfrank I believe you are right, at some point we removed the actual signature check (not sure with what purpose).
The block that was removed was:
```
err = identityService.ValidateSignature(resp.Signature, signingRoot)
if err != nil {
	return centerrors.New(code.AuthenticationFailed, fmt.Sprintf("signature invalid with err: %s", err.Error()))
}
```
I believe the test itself is fine, as long as the mocked service is invoked (keep in mind that modifying bootstrapped svcs after the bootstrapping step might not affect to all the other svcs that may had a copy of the original one). 
My suggestion here is to re-add the block above just after this line: https://github.com/centrifuge/go-centrifuge/blob/develop/p2p/client.go#L362
Might need to mock some other unit tests... But at least we will make this test world test fail.

zyfrank: @mikiquantum about the mocking service, you are right. Injecting mocking service to test should be done carefully case by case to make sure injecting really happen

testworld/signature_test.go
```diff
@@ -17,12 +18,82 @@ import (
...
+	}
+
+	// when got request on signature of document, mocking documents.Service of Mallory return a random signature
+	malloryDocMockSrv.On("RequestDocumentSignature", mock.Anything, mock.Anything, mock.Anything).Return(sig, nil).Once()
+
+	malloryDocMockSrv.On("DeriveFromCoreDocument", mock.Anything).Return(dm, nil).Once()
+
+	//TODO
+	signatures, signatureErrors, err = alice.host.p2pClient.GetSignaturesForDocument(actxh, dm)
+	//seems to me, following should get signature verification errors but it is not.  Currenly p2p/client.go just do validateSignatureResp verification (very simple DID verification?), is this the right behavior?
```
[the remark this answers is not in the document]
zyfrank: still failure after merging latest change, will do some investigation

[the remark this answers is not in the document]
zyfrank: @mikiquantum  I think the root cause is this loop: https://github.com/centrifuge/go-centrifuge/blob/31427c571d89b68a9941a18fb478502e83830f7f/p2p/client.go#L336
In our updated verification logic test, due to verification failure, a null signature will be returned, but this loop ignore this error and keep doing document Anchor (which in this case will not fail because of null signature compared against failure(due to revoked key signing) of doing document Anchor before change) 
Should we keep behavior of this loop (so change the test) or  change this loop and make it return error if any returned signature verification happen error (this may lead to DOS ?)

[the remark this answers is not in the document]
mikiquantum: At the moment, it is intended to continue anchoring even if gathering signatures fail. This might change in the near future, though. The test above in travis pass and as you are asserting that there is a signature error for that collaborator, it is enough to cover the test case. 
For the `TestHost_RevokedSigningKey` test, as the signature validation now fails earlier, for now it is enough that we catch the error (as it is) but let's remove the `assert.Contains(t, message, "failed to validate signatures")` for now.

testworld/incorrect_proto_test.go
```diff
@@ -88,6 +89,7 @@ func TestIncorrectProto_InvalidBody(t *testing.T) {
 	signatures, signatureErrors, err := p.GetSignaturesForDocumentIncorrectMessage(ctxh, dm, "invalidBody")
 	assert.NoError(t, err)
 	assert.Error(t, signatureErrors[0], "Message failed error")
+	fmt.Println(signatureErrors[0].Error())
```
mikiquantum: I think we can remove this `Println`

testworld/signature_test.go
```diff
@@ -96,11 +219,13 @@ func TestHost_RevokedSigningKey(t *testing.T) {
 
 	res := createDocument(bob.httpExpect, bob.id.String(), typeInvoice, http.StatusOK, defaultInvoicePayload([]string{eve.id.String()}))
 	txID := getTransactionID(t, res)
-	status, message := getTransactionStatusAndMessage(bob.httpExpect, bob.id.String(), txID)
+	status, _ := getTransactionStatusAndMessage(bob.httpExpect, bob.id.String(), txID)
+	//New Client Signature veirifation cause following assert not make sense any more
+	//status is now ok and message is null
```
mikiquantum: Let's remove this block for now, keep the comment, and assert that status is `success`:
```
// Even though there was a signature validation error, as of now, we keep anchoring document
assert.Equal(t, status, "success")
```

testworld/incorrect_proto_test.go
zyfrank: of courese

testworld/signature_test.go
zyfrank: ok

p2p/client.go
```diff
@@ -358,7 +357,22 @@ func validateSignatureResp(
...
+	if err != nil {
+		return centerrors.New(code.AuthenticationFailed, fmt.Sprintf("cannot get model timestamp : %s", err.Error()))
+	}
+	signingRoot, err := model.CalculateSigningRoot()
+	if err != nil {
+		return centerrors.New(code.AuthenticationFailed, fmt.Sprintf("failed to calculate signing root: %s", err.Error()))
+	}
+
+	err = s.idService.ValidateSignature(receiver, resp.Signature.PublicKey, resp.Signature.Signature, signingRoot, tm)
+
```
vedhavyas: white space

p2p/client.go
```diff
@@ -358,7 +357,22 @@ func validateSignatureResp(
...
+	if err != nil {
+		return centerrors.New(code.AuthenticationFailed, fmt.Sprintf("signature invalid with err: %s", err.Error()))
```
vedhavyas: `centerrors` are deprecated, please use `errors` instead

zyfrank: what about  package code?

vedhavyas: Ignore this

mikiquantum:
@zyfrank Can you increase the `go timeout` here? https://github.com/centrifuge/go-centrifuge/blob/0a28657575b925d343b08afae1b0391e91a4bf96/build/scripts/tests/run_testworld.sh#L11 
So travis is happy again.

PR #1078, opened 2019-06-03
Integrate proto and pp changes

mikiquantum:
Issue Link: <Github link>


mikiquantum:
@lucasvo expect NFT tests to fail.

PR #1080, opened 2019-06-03
Feat/coreapi handler

vedhavyas:
Issue Link: <Github link>


utils/httputils/httputils.go
```diff
@@ -0,0 +1,6 @@
+package httputils
```
mikiquantum: what do you think about renaming the file hierarchy to: `utils/http/utils.go` keeping the package name as `httputils`?

httpapi/coreapi/documents_api.go
```diff
@@ -0,0 +1,89 @@
...
+// @produce json
+// @Failure 500 {object} httputils.HTTPError
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 403 {object} httputils.HTTPError
+// @success 200 {object} coreapi.DocumentResponse
+// @router /documents [post]
+func (h handler) CreateDocument(w http.ResponseWriter, r *http.Request) {
+	var err error
+	var code int
+	defer func() {
```
mikiquantum: This looks like a common block for all handlers, do you want to extract it so it can be reused?

httpapi/coreapi/documents_api.go
```diff
@@ -0,0 +1,89 @@
...
+			return
+		}
+
+		render.Status(r, code)
+		render.JSON(w, r, httputils.HTTPError{Message: err.Error()})
+	}()
+
+	ctx := r.Context()
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
```
mikiquantum: I would log the errors as well in the server side. Unless you think they are logged downstream.

httpapi/coreapi/types.go
```diff
@@ -0,0 +1,156 @@
...
+	NFTs        []NFT            `json:"nfts"`
+}
+
+// DocumentResponse is the common response for Document APIs.
+type DocumentResponse struct {
+	Header ResponseHeader `json:"header"`
+	Data   interface{}    `json:"data"`
+}
+
+func convertAttributes(cattrs map[string]Attribute) (map[documents.AttrKey]documents.Attribute, error) {
```
mikiquantum:
```suggestion
func toDocumentAttributes(cattrs map[string]Attribute) (map[documents.AttrKey]documents.Attribute, error) {
```

httpapi/coreapi/documents_api.go
vedhavyas: not unless we want to deal with pointers. ATM, i'll keep it as is.

Issue #837, opened 2019-03-11
Incorrect Protocol Messages Testcases


lucasvo:
This ticket is to implement four testcases with our [`testworld` p2p simulation frame work](https://github.com/centrifuge/go-centrifuge/tree/develop/testworld). 

Refer `testworld/README.md` for env setup and a small guide on how to run. `testworld/park_test.go` is a good starting point for any other tests you define. Copy it and define a file called `p2p_connection_test.go` for your tests. Refer to #438 and the related PR to see how this can be achieved.

* Alice's node sends bytes to a Bob that is not a valid protobuf message but has a valid header
* Alice sends bytes to Bob that has invalid headers
* Alice calls Bob but Alice is on a different node/protocol version than Bob
* Eve's node sends a protobuf that is larger than the max allowed size

This issue has a gitcoin bounty attached to it. If you'd like to start working on it, feel free to reach out to me (lucas@centrifuge.io) directly, join our slack channel #bounties on our [slack](
https://centrifuge.io/slack/).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 140.0 DAI (140.0 USD @ $1.0/DAI)  attached to it as part of the Centrifuge fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/centrifuge/go-centrifuge/837/2595).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/centrifuge/go-centrifuge/837/2595).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $47,548.53 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 6 days from now.
Please review their action plans below:


[...]

PR #1083, opened 2019-06-04
 Incorrect Protocol Messages Testcases #837

haritha-j:
issue link : https://github.com/centrifuge/go-centrifuge/issues/837

Implemented tests for invalid body and invalid node version using the document signing feature. Also attempted to implement invalid header by setting random values for the header fields, but not sure if this is the implementation requested in the original issue.


p2p/common/messenger_testworld.go
```diff
@@ -0,0 +1,149 @@
+// +build testworld
```
vimukthi-git: can we rename this file to `protocol_testworld.go`?

p2p/common/messenger_testworld.go
```diff
@@ -0,0 +1,149 @@
...
+
+package p2pcommon
+
+import (
+	"context"
```
vimukthi-git: like mentioned on slack we can run `goimports -w .` to fix this import order everywhere.

testworld/incorrect_proto_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+//send a valid signature request message
+func TestIncorrectProto_ValidMessage(t *testing.T){
+	t.Parallel()
+	// Hosts
+	bob := doctorFord.getHostTestSuite(t, "Bob")
+	oskar := doctorFord.getHostTestSuite(t, "Oskar")
```
vimukthi-git: Lets just use `Eve` in these tests as mentioned in the comment below.

testworld/park.go
```diff
@@ -37,6 +37,7 @@ var hostConfig = []struct {
 	{"Charlie", 8086, 38206, true},
 	{"Kenny", 8087, 38207, false},
 	{"Eve", 8088, 38208, false},
+	{"Oskar", 8089, 38209, false},
```
vimukthi-git: Because of your nice implementation, I see that we don't need `Oskar` at all. Since every new host consumes a lot of resources, we can remove Oskar and just use Eve for your tests right?

testworld/incorrect_proto_test.go
```diff
@@ -0,0 +1,113 @@
+// +build testworld
+
+package testworld
+
+import (
+	"github.com/centrifuge/go-centrifuge/config"
+	"github.com/centrifuge/go-centrifuge/testingutils/config"
+
+	"testing"
```
vimukthi-git: could you organise imports according to https://github.com/golang/go/wiki/CodeReviewComments#imports ? Its the convention we follow. Group standard lib imports first and then the others (like in the other files above ).
Unfortunately `goimports` does not do it sometimes.

haritha-j: oh right, of course.

PR #1085, opened 2019-06-04
feat: nft tokenTransfer and ownerOf support

manuelpolzhofer:
- added nft transferFrom support


nft/invoice_unpaid.go
```diff
@@ -82,13 +82,17 @@ type MintNFTRequest struct {
...
 	GetRequiredInvoiceUnpaidProofFields(ctx context.Context) ([]string, error)
+	// TransferFrom transfers an NFT to another address
+	TransferFrom(ctx context.Context, registry common.Address, to common.Address, tokenID TokenID) (*Response, chan bool, error)
+	// OwnerOf returns the owner of an NFT
+	OwnerOf(registry common.Address, tokenID []byte) (owner common.Address, err error)
 }
 
-// MintNFTResponse holds tokenID and transaction ID.
-type MintNFTResponse struct {
+// Response holds tokenID and transaction ID.
```
charleenfei: what was the reasoning behind making this naming more generic?

manuelpolzhofer: Because I could reuse the same struct for `mint` and `transfer`.

Both should return the `jobID` and the `tokenID`

build/configs/default_config.yaml
```diff
@@ -150,6 +150,7 @@ ethereum:
     anchorCommit: 85000
     anchorPreCommit: 100000
     nftMint: 900000
+    nftTransferFrom: 900000
```
vimukthi-git: :O

manuelpolzhofer: Need to replace it with the correct amount + buffer :)

nft/handler.go
```diff
@@ -97,6 +98,44 @@ func (g grpcHandler) MintInvoiceUnpaidNFT(ctx context.Context, request *nftpb.NF
...
 
+// TokenTransfer will be called to transfer a token owned by the identity contract
+func (g grpcHandler) TokenTransfer(ctx context.Context, request *nftpb.TokenTransferRequest) (*nftpb.TokenTransferResponse, error) {
+	ctxHeader, err := contextutil.Context(ctx, g.config)
+	if err != nil {
+		apiLog.Error(err)
+		return nil, err
+	}
+
+	tokenID, err := TokenIDFromString(request.TokenId)
```
vedhavyas: Why export this function ?

nft/invoice_unpaid.go
vedhavyas: IMO, TokenResponse might be specific instead of Response which seems too generic. Also, do we need to export this structure?

nft/service_integration_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/centrifuge/go-centrifuge/testingutils"
+
```
vedhavyas: import spacing

nft/handler.go
manuelpolzhofer: We are currently using it in the integration tests

protobufs/nft/service.proto
```diff
@@ -23,19 +23,57 @@ service NFTService {
   }
   rpc MintInvoiceUnpaidNFT(NFTMintInvoiceUnpaidRequest) returns (NFTMintResponse) {
     option (google.api.http) = {
-      post: "/token/mint/invoice/unpaid/{identifier}"
+      post: "/nfts/{identifier}/invoice/unpaid/mint"
```
manuelpolzhofer: @rdinicut new mint endpoint rest api path

rdinicut: we do not care about the paths. That will get updated when we regenerate the api. If the way we call the api changes it will break stuff

manuelpolzhofer: cool.

PR #1104, opened 2019-06-07
set range for webhook http status code job log

### LINKED ISSUE #1100: Job webhook send misleading log
Currently we have this check:
https://github.com/centrifuge/go-centrifuge/blob/2cacbe13be88c64e818ac59994e27550e8e887e5/notification/notification.go#L69

This will not work for all the 2xx range responses. Let's make sure that we do not error out on some of the 201 for example.

charleenfei:
Closes #1100

### DIFF
--- a/notification/notification.go
+++ b/notification/notification.go
@@ -3,8 +3,6 @@ package notification
 import (
 	"context"
 	"encoding/json"
-	"net/http"
-
 	"github.com/centrifuge/go-centrifuge/contextutil"
 
 	"github.com/centrifuge/centrifuge-protobufs/gen/go/notification"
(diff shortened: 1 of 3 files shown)


notification/notification.go
```diff
@@ -66,7 +64,7 @@ func (wh webhookSender) Send(ctx context.Context, notification *notificationpb.N
 		return Failure, err
 	}
 
-	if statusCode != http.StatusOK {
+	if !utils.InRange(statusCode, 200, 205){
```
vedhavyas: lets accept full 2xx codes. Any 2xx can be considered okay

PR #1112, opened 2019-06-12
Pedersen Hash function

mikiquantum:
Issue Link: <Github link>


crypto/pedersen/pedersen.go
```diff
@@ -0,0 +1,56 @@
...
+	hashed      []byte
+}
+
+func (pd *pedersen) Reset() {
+	pd.hashed = []byte{}
+}
+
+func (pd *pedersen) Write(p []byte) (n int, err error) {
+	args := []string{pd.binLocation, "hash", hex.EncodeToString(p)}
+	o, err := exec.Command("python3", args...).Output()
```
mikiquantum: Temporary until we have a go implementation of this hashing function.

vedhavyas: we have to update the docker to have the python 3 executable then

crypto/pedersen/pedersen.go
```diff
@@ -0,0 +1,56 @@
...
+		return 0, err
+	}
+	pd.hashed, err = hex.DecodeString(string(o[:len(o)-1])) // removing newline
+	if err != nil {
+		return 0, err
+	}
+	return len(p), nil
+}
+
+func (pd *pedersen) Sum(b []byte) []byte {
```
vedhavyas: `Sum` should append the `b` to pd.hashed and return the result

crypto/pedersen/pedersen.go
```diff
@@ -0,0 +1,56 @@
...
+	return pd.hashed
+}
+
+func (pd *pedersen) Size() int {
+	return len(pd.hashed)
+}
+
+func (pd *pedersen) BlockSize() int {
+	return 0
```
vedhavyas: I think the block size is 32 but need to check the python code. Its better we return the block size instead of 0 here

crypto/pedersen/pedersen_test.go
```diff
@@ -0,0 +1,25 @@
...
+package pedersen
+
+import (
+	"testing"
+
+	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHashPayload(t *testing.T) {
```
vedhavyas: Let us also test for predefined payload and hash and compare the hashes instead?

crypto/pedersen/pedersen.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Requires to follow instructions to install pycrypto: https://github.com/stefandeml/zokrates-pycrypto
+// And either set CENT_PED_BIN with cli.py location or store the repo in $HOME/zksnarks/pycrypto
+
+// NewPedersenHash returns a pedersen hash function
+func NewPedersenHash() hash.Hash {
+	// Resolve python binary location
+	pedBin := os.Getenv("CENT_PED_BIN")
+	if pedBin == "" {
+		pedBin = fmt.Sprintf("%s/zksnarks/pycrypto/cli.py", os.Getenv("HOME"))
```
vedhavyas: IMO, I think we should ship the python libs into the docker so that we can properly version them. That way we can always know and not expect anything from the env and wont break anything when we move to the native go implementation except for the docker which will be self contained

crypto/pedersen/pedersen_test.go
```diff
@@ -0,0 +1,25 @@
+// +build pedersenunit
```
vedhavyas: Why the new build tag ?

PR #1119, opened 2019-06-17
TransferDetails Services

### LINKED ISSUE #1110: TransferDetails User API

This new extension should be implemented with a new set of custom attributes and allow for transfer details to be added to a Document.

charleenfei:
Closes #1110 

- nolint on transferdetails/funding.go/userapi types.go files because of special needs for parsing due to custom attributes (temporary solution)
- separates out generic attribute utils into
- services and testing
- some work on migrating funding extension to userAPI format, but this will be finished in a separate PR

opening another PR for the userapi portion due to size of this PR.


extensions/transferdetails/converters.go
```diff
@@ -0,0 +1,27 @@
+package transferdetails
```
charleenfei: this file will be removed.

extensions/funding/converters.go
```diff
@@ -0,0 +1,35 @@
+package funding
+
+import (
+	"reflect"
+
+	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/funding"
+)
+
+// TODO: generic? Create something like ServiceRegistry for Custom Attributes...
+func (f *Data) initFundingFromData(data *funpb.FundingData) {
```
manuelpolzhofer: I think this method could be generic

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+			c = append(c, did)
+		}
+	}
+
+	return c, nil
+}
+
+func (s service) DeriveFromPayload(ctx context.Context, req CreateTransferDetailRequest, identifier []byte) (documents.Model, error) {
+	model, err := s.GetCurrentVersion(ctx, identifier)
```
manuelpolzhofer: This method could be more generic.

Where you pass in `transfersFieldKey, transfersLabel`  and the `dids`

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+	idx, err := extensions.FindAttributeSetIDX(model, transferID, transfersLabel, transferIDLabel, transfersFieldKey)
+	if err != nil {
+		return nil, err
+	}
+	return s.deriveTransferData(model, idx)
+}
+
+// TODO: generic?
+func (s service) deriveTransferData(model documents.Model, idx string) (*TransferDetailData, error) {
+	data := new(TransferDetailData)
```
manuelpolzhofer: Same here:

The only individual part is:`data := new(TransferDetailData)`

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+	return &TransferDetailResponse{
+		Header: h,
+		Data:   data,
+	}, nil
+
+}
+
+// DeriveTransfersListResponse returns a transfers list
+func (s service) DeriveTransferListResponse(ctx context.Context, model documents.Model) (*TransferDetailListResponse, error) {
+	response := new(TransferDetailListResponse)
```
manuelpolzhofer: Same here:

We only need an additional parameter for the struct to be generic. like ` TransferDetailData` or `Funding`

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+
+		}
+
+	}
+	return data, nil
+}
+
+// DeriveTransferResponse returns create response from the added transfer detail
+func (s service) DeriveTransferResponse(ctx context.Context, model documents.Model, transferID string) (*TransferDetailResponse, error) {
+	idx, err := extensions.FindAttributeSetIDX(model, transferID, transfersLabel, transferIDLabel, transfersFieldKey)
```
manuelpolzhofer: same here:

the method could be moved to the extension package together with an extra parameter for the struct.

this method would only call the generic method and afterwards return the `&TransferDetailResponse{...}`

extensions/funding/converters.go
charleenfei: this whole file should be removed when we get rid of the protobuf :)

charleenfei:
totally agree @manuelpolzhofer -- I think it should be part of the rethinking when we migrate the fundingAPI also over to the new userAPI structure.

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/centrifuge/go-centrifuge/identity"
+	logging "github.com/ipfs/go-log"
+)
+
+// Service defines specific functions for extension funding
+type Service interface {
+	documents.Service
+
+	// DeriveFromUpdatePayload derives Funding from clientUpdatePayload
+	DeriveFromUpdatePayload(ctx context.Context, req UpdateTransferDetailRequest, identifier []byte) (documents.Model, error)
```
vedhavyas: Why not make the identifier part of the `req` ?

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/centrifuge/go-centrifuge/documents"
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/extensions"
+	"github.com/centrifuge/go-centrifuge/identity"
+	logging "github.com/ipfs/go-log"
+)
+
+// Service defines specific functions for extension funding
+type Service interface {
+	documents.Service
```
vedhavyas: Lets not embed this, let keep it as a field instead

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	return model, nil
+}
+
+// DeriveFromUpdatePayload derives Funding from clientUpdatePayload
+func (s service) DeriveFromUpdatePayload(ctx context.Context, req UpdateTransferDetailRequest, identifier []byte) (documents.Model, error) {
+	model, err := s.GetCurrentVersion(ctx, identifier)
+	if err != nil {
+		transfersLog.Error(err)
```
vedhavyas:
```suggestion
		log.Error(err)
```

vedhavyas: We will have only one log instance for this entire package. So calling logger should be enough IMO

httpapi/userapi/transfer_details_api.go
```diff
@@ -0,0 +1,14 @@
+package userapi
+
+import (
+	logging "github.com/ipfs/go-log"
+)
+
+var transfersLog = logging.Logger("tranfers-api")
```
vedhavyas:
```suggestion
var log = logging.Logger("user-api")
```

vedhavyas: TransferAPIService in `transferdetails` package will have its own logger and UserAPI logger is used for all USerAPI logs

httpapi/userapi/types.go
```diff
@@ -0,0 +1,53 @@
...
+
+// TransferDetailResponse
+type TransferDetailResponse struct {
+	Header *documentpb.ResponseHeader          `json:"header"`
+	Data   *transferdetails.TransferDetailData `json:"data"`
+}
+
+// TransferDetailListResponse
+type TransferDetailListResponse struct {
+	Header *documentpb.ResponseHeader            `json:"header"`
```
vedhavyas: We want to get rid of protobufs. Use the CoreAPI ResponseHeader instead

httpapi/userapi/types.go
```diff
@@ -0,0 +1,53 @@
...
+//CreateTransferDetailRequest
+type CreateTransferDetailRequest struct {
+	Identifier string                              `json:"identifier" swaggertype:"primitive,string"`
+	Data       *transferdetails.TransferDetailData `json:"data"`
+}
+
+// UpdateTransferDetailRequest
+type UpdateTransferDetailRequest struct {
+	Identifier string                              `json:"identifier" swaggertype:"primitive,string"`
+	TransferId string                              `json:"transfer_id" swaggertype:"primitive,string"`
```
vedhavyas:
```suggestion
	TransferID string                              `json:"transfer_id" swaggertype:"primitive,string"`
```

httpapi/userapi/types.go
```diff
@@ -0,0 +1,53 @@
...
+package userapi
+
+import (
+	"github.com/centrifuge/go-centrifuge/extensions/transferdetails"
+	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/document"
+)
+
+// TransferDetailData is the default transfer details extension schema
+type TransferDetailData struct {
+	TransferId          string `json:"transfer_id,omitempty" attr:"bytes"`
```
vedhavyas:
```suggestion
	TransferID          string `json:"transfer_id,omitempty" attr:"bytes"`
```

charleenfei:
closing this to open smaller PRs

PR #1122, opened 2019-06-17
Feat/generic

### LINKED ISSUE #1121: FlexDocs V1
A new document type with Empty Data. Used only for Attributes

vedhavyas:
Closes #1121


documents/generic/model.go
```diff
@@ -0,0 +1,303 @@
...
+// tree prefixes for specific to documents use the second byte of a 4 byte slice by convention
+func compactPrefix() []byte { return []byte{0, 5, 0, 0} }
+
+// Data is a empty  structure.
+type Data struct{}
+
+// Generic implements the documents.Model for Generic documents
+type Generic struct {
+	*documents.CoreDocument
+	Data Data
```
vimukthi-git: why do we need this at all?

documents/generic/service_test.go
```diff
@@ -0,0 +1,162 @@
...
+	testingcommons "github.com/centrifuge/go-centrifuge/testingutils/commons"
+	testingconfig "github.com/centrifuge/go-centrifuge/testingutils/config"
+	"github.com/centrifuge/go-centrifuge/testingutils/testingjobs"
+	"github.com/centrifuge/go-centrifuge/utils"
+	"github.com/centrifuge/gocelery"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestService_CreateModel(t *testing.T) {
```
vimukthi-git: IMO this test doesn't really test a generic doc, I would like to see it actually use custom attrs to store the fields.

vedhavyas: That is part of the testworld and documents test. This is just storing the generic doc

documents/generic/model.go
vedhavyas: At the moment, we dont but we would need this soon.

documents/generic/service_test.go
vedhavyas: TestCoreAPI_DocumentGenericCreateAndUpdate in the testworld and attributes does that since attributes are common to all the documents and the tests above are intended for specific document types alone

documents/generic/model.go
vedhavyas: Hence, keeping it empty would

PR #1135, opened 2019-06-19
TransferDetails Pt1: Generic attribute util functions, funding related changes

charleenfei:

- generic attr utils
- generic errors
- edited update/create payload methods on funding
- fixed breaking tests


extensions/funding/service.go
```diff
@@ -186,20 +73,31 @@ func deriveDIDs(data *clientfunpb.FundingData) ([]identity.DID, error) {
...
 }
 
-func (s service) DeriveFromPayload(ctx context.Context, req *clientfunpb.FundingCreatePayload, identifier []byte) (documents.Model, error) {
+func (s service) DeriveFromPayload(ctx context.Context, req *clientfunpb.FundingCreatePayload) (model documents.Model, err error) {
 	var fd Data
 	fd.initFundingFromData(req.Data)
 
-	model, err := s.GetCurrentVersion(ctx, identifier)
+	var docID []byte
+	if req.DocumentId != "" {
```
manuelpolzhofer: maybe to avoid nested if statements:
```golang
if req.DocumentId == ""  {
return nil, documents.ErrDocumentIdentifier
}
docID, err = hexutil.Decode(req.DocumentId)
if err != nil {
   return nil, err
}
````

File: extensions/funding/service.go [hunk 2]
```diff
@@ -225,57 +123,47 @@ func (s service) DeriveFromPayload(ctx context.Context, req *clientfunpb.Funding
...
-
 // DeriveFromUpdatePayload derives Funding from clientUpdatePayload
-func (s service) DeriveFromUpdatePayload(ctx context.Context, req *clientfunpb.FundingUpdatePayload, identifier []byte) (documents.Model, error) {
+func (s service) DeriveFromUpdatePayload(ctx context.Context, req *clientfunpb.FundingUpdatePayload) (model documents.Model, err error) {
 	var fd Data
 	fd.initFundingFromData(req.Data)
 
-	model, err := s.GetCurrentVersion(ctx, identifier)
+	var docID []byte
+	if req.DocumentId != "" {
```
manuelpolzhofer: maybe a small helper method if we need it twice

extensions/funding/service.go
```diff
@@ -300,63 +188,24 @@ func (s service) DeriveFromUpdatePayload(ctx context.Context, req *clientfunpb.F
...
-		}
-
-	}
-
-	return idx, ErrFundingNotFound
-}
-
+// TODO: Move to attribute utils
 func (s service) deriveFundingData(model documents.Model, idx string) (*Data, error) {
 	data := new(Data)
```
manuelpolzhofer: I think we can move the entire deriveFundingData to the extension package. With a data parameter `interface{}`

After that the `extensions.LabelFromJSONTag` can be private instead of public

extensions/funding/service.go
manuelpolzhofer: We should rename the error to `ErrDocumentID`

extensions/funding/service.go
manuelpolzhofer: It is not a requirement for this PR

extensions/attribute_utils_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+package extensions
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO: more testing for attribute utils functions
```
vedhavyas: todo ?

vedhavyas:
@charleenfei let try and bring up the test coverage to atleast 80%

PR #1136, opened 2019-06-19
TransferDetails Pt2: TD service, Bootstrapper

charleenfei:
- service and tests
- expose Service through Bootstrapper, tests


extensions/transferdetails/transferdetails.go
```diff
@@ -0,0 +1,42 @@
...
+
+// UpdateTransferDetailRequest holds the required fields to update a transfer agreement
+type UpdateTransferDetailRequest struct {
+	DocumentID string
+	TransferID string
+	Data       *TransferDetailData
+}
+
+// TransferDetailResponse holds a TransferDetail response
+type TransferDetailResponse struct {
```
charleenfei: this will be renamed TransferDetail in the next PR to avoid confusion with userapi types

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/centrifuge/go-centrifuge/errors"
+	"github.com/centrifuge/go-centrifuge/extensions"
+	"github.com/centrifuge/go-centrifuge/identity"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	logging "github.com/ipfs/go-log"
+)
+
+// Service defines specific functions for extension funding
+type Service interface {
+	documents.Service
```
manuelpolzhofer: We should change it do:
```suggestion
	srv documents.Service
```

We don't want to expose the general service method to the new handlers.
We need to change it for funding agreement later too.

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,284 @@
...
+		return nil, err
+	}
+	return s.deriveTransferData(model, idx)
+}
+
+// TODO: move to generic function in attribute utils
+func (s service) deriveTransferData(model documents.Model, idx string) (*TransferDetailData, error) {
+	data := new(TransferDetailData)
+
+	types := reflect.TypeOf(*data)
```
manuelpolzhofer: can be moved to extension package starting from this line

extensions/transferdetails/service.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	logging "github.com/ipfs/go-log"
+)
+
+// Service defines specific functions for extension funding
+type Service interface {
+	documents.Service
+
+	// DeriveFromUpdatePayload derives Funding from clientUpdatePayload
+	DeriveFromUpdatePayload(ctx context.Context, req UpdateTransferDetailRequest) (documents.Model, error)
```
manuelpolzhofer: The new api handler should be really simple, so we should move the existing logic from the handler. (calling derive and calling upate) to the service.


The service should only have public methods for the main interaction called by the new handler.

Like `Create` or `Update`. DeriveFromPayload methods should be private.

extensions/transferdetails/service.go
charleenfei: will do it when funding is migrated and we can make things more generic

extensions/transferdetails/service.go
charleenfei: I understand what you mean, but it is not possible here as this is an interface, not a struct

extensions/transferdetails/service.go
charleenfei: following conversation with @mikiquantum , will leave as is.

PR #1140, opened 2019-06-20
TransferDetails Pt3: userAPI/handler

### LINKED ISSUE #1110: TransferDetails User API

charleenfei:
Closes #1110


extensions/transferdetails/service.go
```diff
@@ -0,0 +1,350 @@
...
+		DocumentID: model.ID(),
+		CreatePayload: documents.CreatePayload{
+			Scheme:        model.Scheme(),
+			Collaborators: cs,
+			Attributes:    attr,
+			Data:          d,
+		},
+	}
+
+	updated, jobID, err := s.srv.UpdateModel(ctx, payload)
```
mikiquantum: Why we can't use here the `coreapi.Service` instead of the `documents.Service`? From my point of view any userapi logic, should always interact with the `coreapi.Service` and not directly to the underlaying documents service.
Does it trigger a cyclic dep?

httpapi/userapi/transfer_details_api.go
```diff
@@ -0,0 +1,268 @@
...
+// @accept json
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param body body userapi.CreateTransferDetailRequest true "Transfer Detail Create Request"
+// @param document_id path string true "Document Identifier"
+// @produce json
+// @Failure 500 {object} httputils.HTTPError
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 403 {object} httputils.HTTPError
+// @success 201 {object} userapi.TransferDetailResponse
+// @router /v1/documents/extensions/transfer_details [post]
```
mikiquantum: dont we need here the `document_id` as part of the path?

httpapi/router.go
```diff
@@ -49,6 +53,8 @@ func Router(
 	r.Route("/v1", func(r chi.Router) {
 		// core apis
 		coreapi.Register(r, nftSrv, configSrv, docsSrv, jobsSrv)
+		// user apis
+		userapi.Register(r, nftSrv, transferSrv)
```
mikiquantum: I am not really sure how this would look like when we want to have different versions, from core and userapi. Since the Router prefix `v1` for all services underneath. Lets keep it like this, but we might need to create either an explicit core route or duplicate the chi service.

extensions/transferdetails/service.go
charleenfei: yes, it does :/

mikiquantum: ok, this would involve a nice refactoring... lets add a comment there and move on with this implementation.

httpapi/router.go
charleenfei: okay, I will make a note to rethink this.

extensions/transferdetails/service.go
charleenfei: "nice refactoring" ^-^

extensions/transferdetails/service.go
```diff
@@ -2,30 +2,32 @@ package transferdetails
...
 	logging "github.com/ipfs/go-log"
 )
 
 // Service defines specific functions for extension funding
 type Service interface {
 
-	// DeriveFromUpdatePayload derives TransferDetail from clientUpdatePayload
-	DeriveFromUpdatePayload(ctx context.Context, req UpdateTransferDetailRequest) (documents.Model, error)
+	// UpdateTransferDetail updates a TransferDetail
+	UpdateTransferDetail(ctx context.Context, req UpdateTransferDetailRequest) (documents.Model, jobs.JobID, error)
```
manuelpolzhofer: I think in the go coding standard the package name should not be part of a method name.

We could rename it to `Update` and `Create`.

httpapi/userapi/transfer_details_api_test.go
```diff
@@ -0,0 +1,87 @@
+// +build unit
+
+package userapi
+
+//func TestHandler_CreateDocument(t *testing.T) {
```
vedhavyas: Why commented out ?

httpapi/userapi/types.go
```diff
@@ -0,0 +1,72 @@
...
+type CreateTransferDetailRequest struct {
+	DocumentID string                              `json:"document_id" swaggertype:"primitive,string"`
+	Data       *transferdetails.TransferDetailData `json:"data"`
+}
+
+// UpdateTransferDetailRequest is the request body for updating a Transfer Detail
+type UpdateTransferDetailRequest struct {
+	DocumentID string                              `json:"document_id" swaggertype:"primitive,string"`
+	TransferID string                              `json:"transfer_id" swaggertype:"primitive,string"`
+	Data       *transferdetails.TransferDetailData `json:"data"`
```
vedhavyas: `Data` should not be pointer. Value should be good

httpapi/userapi/types.go
```diff
@@ -0,0 +1,72 @@
...
+func toTransferDetailCreatePayload(request CreateTransferDetailRequest) (*transferdetails.CreateTransferDetailRequest, error) {
+	payload := new(transferdetails.CreateTransferDetailRequest)
+	payload.DocumentID = request.DocumentID
+	payload.Data = request.Data
+
+	return payload, nil
+}
+
+func toTransferDetailUpdatePayload(request UpdateTransferDetailRequest) (*transferdetails.UpdateTransferDetailRequest, error) {
+	payload := new(transferdetails.UpdateTransferDetailRequest)
```
vedhavyas: Use initialisation  of struct instead of assiging after initialisation

```
payload := transferdetails.UpdateTransferDetailRequest {
	...
}

```

httpapi/userapi/userapi.go
```diff
@@ -0,0 +1,31 @@
...
+// Register registers the core apis to the router.
+func Register(r chi.Router,
+	docSrv documents.Service,
+	nftSrv nft.Service,
+	transferSrv transferdetails.Service) {
+	h := handler{
+		tokenRegistry: nftSrv.(documents.TokenRegistry),
+		srv:           Service{docSrv: docSrv, transferDetailsService: transferSrv},
+	}
+	r.Post("/documents/{"+documentIDParam+"}/extensions/transfer_details", h.CreateTransferDetail)
```
vedhavyas: any specific reason for adding `extensions` to the url path ?
IMO, we dont need this to make it consistent with the funding API

httpapi/userapi/userapi.go
```diff
@@ -0,0 +1,31 @@
...
+	r.Put("/documents/{"+documentIDParam+"}/extensions/transfer_details/{"+transferIDParam+"}", h.UpdateTransferDetail)
+	r.Get("/documents/{"+documentIDParam+"}/extensions/transfer_details", h.GetTransferDetailList)
+	r.Get("/documents/{"+documentIDParam+"}/extensions/transfer_details/{"+transferIDParam+"}", h.GetTransferDetail)
+	// TODO: future GET methods for specific versions
+	//r.Get("/documents/{"+documentIDParam+"}/versions/{"+versionIDParam+"}/extensions/transfer_details", h.GetVersionTransferDetailList)
```
vedhavyas: Lets remove the commented code. This is still future work in DIfferent PR :)

p2p/receiver/handler_integration_test.go
```diff
@@ -138,7 +137,6 @@ func TestHandler_RequestDocumentSignature(t *testing.T) {
 
 	// valid transition for collaborator id
 	resp, err := handler.RequestDocumentSignature(ctxh, &p2ppb.SignatureRequest{Document: &ncd}, defaultDID)
-	fmt.Println(ncd.PreviousVersion, ncd.CurrentVersion)
```
vedhavyas: 👍

httpapi/userapi/userapi.go
charleenfei: if all userapis will be extensions, then you are right and this is redundant

httpapi/userapi/transfer_details_api_test.go
charleenfei: testing is a WIP, just finished the functionality first for PR review and for rasvan to start experimenting with the API calls

PR #1153, opened 2019-07-01
use core api service

### LINKED ISSUE #1142: Inject and use of core api service in any userapi service
Start here: https://github.com/centrifuge/go-centrifuge/blob/develop/httpapi/router.go#L54

Replace: https://github.com/centrifuge/go-centrifuge/blob/f9872c4af9e96500993d8b797c125a25a55f73c7/extensions/transferdetails/service.go#L102

vedhavyas:
Fixes #1142

### DIFF
--- a/api/service.go
+++ b/api/service.go
@@ -4,15 +4,12 @@ import (
 	"github.com/centrifuge/go-centrifuge/bootstrap"
 	"github.com/centrifuge/go-centrifuge/config"
 	"github.com/centrifuge/go-centrifuge/config/configstore"
-	"github.com/centrifuge/go-centrifuge/documents"
 	"github.com/centrifuge/go-centrifuge/documents/entity"
 	"github.com/centrifuge/go-centrifuge/documents/invoice"
 	"github.com/centrifuge/go-centrifuge/documents/purchaseorder"
 	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-centrifuge/extensions/funding"
-	"github.com/centrifuge/go-centrifuge/extensions/transferdetails"
 	"github.com/centrifuge/go-centrifuge/httpapi"
-	"github.com/centrifuge/go-centrifuge/jobs"
 	"github.com/centrifuge/go-centrifuge/nft"
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/account"
 	"github.com/centrifuge/go-centrifuge/protobufs/gen/go/entity"
@@ -44,16 +41,6 @@ func registerServices(ctx context.Context, grpcServer *grpc.Server, gwmux *runti
 		return nil, errors.New("failed to get %s", bootstrap.BootstrappedInvoiceUnpaid)
 	}
 
-	docService, ok := nodeObjReg[documents.BootstrappedDocumentService].(documents.Service)
-	if !ok {
-		return nil, errors.New("failed to get %s", documents.BootstrappedDocumentService)
-	}
-
-	transferService, ok := nodeObjReg[transferdetails.BootstrappedTransferDetailService].(transferdetails.Service)
-	if !ok {
-		return nil, errors.New("failed to get %s", transferdetails.BootstrappedTransferDetailService)
-	}
-
 	// register document types
 	err = registerDocumentTypes(ctx, nodeObjReg, grpcServer, gwmux, addr, dopts)
 	if err != nil {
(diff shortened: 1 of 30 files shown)


vimukthi-git:
@vedhavyas @charleenfei closing this as I don't have an idea about your design.
Please make sure to document or make code comments about your expected designs, since you never know when other team members would have to take over for you when you are not here.

cc @mikiquantum

vedhavyas:
@vimukthi-git yes, we will document the file structure next time in the spec. Apologies for the confusion :)

httpapi/router.go
```diff
@@ -33,12 +29,9 @@ import (
...
 // @schemes http
 func Router(
+	ctx map[string]interface{},
 	config Config,
-	configSrv config.Service,
-	nftSrv nft.Service,
-	docsSrv documents.Service,
-	transferSrv transferdetails.Service,
-	jobsSrv jobs.Manager) *chi.Mux {
+	configSrv config.Service) *chi.Mux {
```
vimukthi-git: config service is available within the context.

httpapi/userapi/bootstrapper.go
```diff
@@ -0,0 +1,29 @@
...
+	if !ok {
+		return errors.New("failed to get %s", documents.BootstrappedDocumentService)
+	}
+
+	tdSrv, ok := ctx[transferdetails.BootstrappedTransferDetailService].(transferdetails.Service)
+	if !ok {
+		return errors.New("failed to get %s", transferdetails.BootstrappedTransferDetailService)
+	}
+
+	ctx[BootstrappedUserAPIService] = Service{docSrv, tdSrv}
```
vimukthi-git: why is `UserAPI` allowed to use `documents.Service` ? for me this defeats the purpose of core vs user separation. It's one of the main problems I solved in my pr.

httpapi/router.go
```diff
@@ -33,12 +29,9 @@ import (
 // @host localhost:8082
 // @schemes http
 func Router(
+	ctx map[string]interface{},
```
vimukthi-git: I don't like the use of `ctx map[string]interface{}` outside of bootstrappers, lets just use the `context.Context` passed to `apiServer.Start` method when referring to the context after server process has started, since its not thread safe afterwards. I know its not being written to but this increases the possibility.

api/service.go
```diff
@@ -67,8 +54,7 @@ func registerServices(ctx context.Context, grpcServer *grpc.Server, gwmux *runti
 	}
 
 	cfg := nodeObjReg[bootstrap.BootstrappedConfig].(config.Configuration)
-	jobsMan := nodeObjReg[jobs.BootstrappedService].(jobs.Manager)
-	return httpapi.Router(cfg, configService, invoiceUnpaidService, docService, transferService, jobsMan), nil
+	return httpapi.Router(nodeObjReg, cfg, configService), nil
```
vimukthi-git: this call is not necessary to be done here, we can move it to `api/server.go` (like in my pr), this is mixup of new code and old code unnecessarily.

httpapi/userapi/bootstrapper.go
charleenfei: i think the question for me here is why not use the docSrv as wrapped in the coreAPISrv instead?

vedhavyas: Ahh. missed this one. Thanks!

httpapi/router.go
vedhavyas: We are already using it in api package. It is called synchronously. ATM, I would like to keep it that way

vedhavyas: I'll change it, seems like we follow this in API already

httpapi/userapi/transfer_details_api_test.go
```diff
@@ -28,7 +28,7 @@ import (
 func TestHandler_CreateTransferDetail(t *testing.T) {
 	docSrv := new(testingdocuments.MockService)
 	transferSrv := new(MockTransferService)
-	srv := Service{docSrv: docSrv, transferDetailsService: transferSrv}
+	srv := Service{coreAPISrv: newCoreAPIService(docSrv), transferDetailsService: transferSrv}
```
vimukthi-git: this test needs a mock of `coreAPISrv`, i believe we need an interface. Otherwise we are also testing the coreservice here.

vedhavyas: @vimukthi-git CoreAPI service will just be a wrapper for services underneath. Making this a service just adds another extra layer of interface with no usage IMO.

vimukthi-git: usage is testing as I said above, here we can't test user service without testing the core service as well (This is only for now, there would be more uses in the future if we are creating core service for any real purpose).

vedhavyas: We do not need to create any interface for CoreAPIservice since its consists the interfaces and any logic would just be testable since the underneath services will mocks. I'm thinking we can cross the brodge when we really need to make it an interface. Let me know if you wanna discuss more offline :)

PR #1157, opened 2019-07-03
Invoice create API migration

vimukthi-git:
Issue Link: <Github link>


httpapi/userapi/invoice_api.go
```diff
@@ -0,0 +1,100 @@
...
+		log.Error(err)
+		return
+	}
+
+	render.Status(r, http.StatusCreated)
+	render.JSON(w, r, resp)
+
+}
+
+func toDocumentsCreatePayload(request CreateInvoiceRequest) (documents.CreatePayload, error) {
```
vimukthi-git: @vedhavyas I had to keep this function as is since I'm creating `CreateInvoiceRequest ` for swagger purpose, now it doesn't make sense to convert it back to `CreateDocumentRequest` just to pass it to the converter on `coreapi`

httpapi/userapi/purchase_order_api.go
```diff
@@ -170,7 +170,7 @@ func (h handler) UpdatePurchaseOrder(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	render.Status(r, http.StatusCreated)
+	render.Status(r, http.StatusOK)
```
vimukthi-git: fix

vedhavyas: This should still be 201. Could you please let me know what is the spec you are referring to ?

vedhavyas:
```

The request has been fulfilled and resulted in a new resource being created. The newly created resource can be referenced by the URI(s) returned in the entity of the response, with the most specific URI for the resource given by a Location header field. The response SHOULD include an entity containing a list of resource characteristics and location(s) from which the user or user agent can choose the one most appropriate. The entity format is specified by the media type given in the Content-Type header field. The origin server MUST create the resource before returning the 201 status code. If the action cannot be carried out immediately, the server SHOULD respond with 202 (Accepted) response instead.
```
This the text from w3.com. Since we updating the anchor, whuich is not confirmed yet, the status should be 201 created but not ok 200

vimukthi-git: this is an update endpoint, where the resource `/documents/<id>` is updated. 
```
If an existing resource is modified, either the 200 (OK) or 204 (No Content) response codes > SHOULD be sent to indicate successful completion of the request.
```
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html
If your argument is that new version is created, then the resource name for the `PUT` should be `/documents/versions`.

vedhavyas: But the request is not complete until the anchor is created which is inturn a new resource for us and is the process of completion but not complete yet. `/documents/<id>` you mentioned is not a resource here. It is the new version that is derived from the doc that is the new resource which is not created yet and in the process of being created. Hence, as per my understanding, this would require us to return 201 so that users can poll the job status for the completion

vedhavyas: `/documents/<id>` has changed in the local node but not propagated yet. User would need to wait until the completion for it be propagated to other collaborators for it to be successful

vimukthi-git: you are right, I think in that case it makes sense to return http 202 accepted, since the its still being processed. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202

vimukthi-git:
@vedhavyas as agreed offline, merging this to change statuses later.

PR #1168, opened 2019-07-05
add key to attr response

### LINKED ISSUE #1165: Attribute Key is not returns in the New API
In the document response, attribute key, hash of the label, is not returned.

vedhavyas:
Closes #1165


httpapi/coreapi/types.go
```diff
@@ -58,13 +67,13 @@ type ResponseHeader struct {
 
 // DocumentResponse is the common response for Document APIs.
 type DocumentResponse struct {
-	Header     ResponseHeader `json:"header"`
-	Scheme     string         `json:"scheme" enums:"generic,invoice,purchase_order,entity"`
-	Data       interface{}    `json:"data"`
-	Attributes AttributeMap   `json:"attributes"`
+	Header     ResponseHeader       `json:"header"`
+	Scheme     string               `json:"scheme" enums:"generic,invoice,purchase_order,entity"`
```
manuelpolzhofer: Do you think it is possible to get tag from a config file. `json:"scheme" enums:"generic,invoice,purchase_order,entity"`

Let's say a new dev is trying to add a new doc type.
I guess it is a bit hard to figure out you have to change some tags.

PR #1190, opened 2019-07-12
Cleanuup/entity

vedhavyas:


documents/entity/bootstrapper.go
```diff
@@ -77,7 +77,7 @@ func (Bootstrapper) Bootstrap(ctx map[string]interface{}) error {
 	srv := DefaultService(
 		docSrv,
 		repo,
-		queueSrv, jobManager, factory, erService, didService, anchorRepo, processor, func() documents.ValidatorGroup {
+		queueSrv, jobManager, factory, erService, anchorRepo, processor, func() documents.ValidatorGroup {
```
charleenfei: why is didService gone here?

vedhavyas: we dont need it anymore

Issue #1098, opened 2019-06-07
Do not use `done` channel as error checking anywhere, check the job status


mikiquantum:
As shown here: https://github.com/centrifuge/go-centrifuge/blob/develop/identity/ideth/service.go#L143

This avoids propagating an error upstream as done will be always `true` no matter the underlaying status.

PR #1216, opened 2019-07-24
use error chan

mikiquantum:
closes: #1098

### DIFF
--- a/identity/did.go
+++ b/identity/did.go
@@ -211,10 +211,10 @@ type Service interface {
 	GetKey(did DID, key [32]byte) (*KeyResponse, error)
 
 	// RawExecute calls the execute method on the identity contract
-	RawExecute(ctx context.Context, to common.Address, data []byte, gasLimit uint64) (txID IDTX, done chan bool, err error)
+	RawExecute(ctx context.Context, to common.Address, data []byte, gasLimit uint64) (txID IDTX, done chan error, err error)
 
 	// Execute creates the abi encoding and calls the execute method on the identity contract
-	Execute(ctx context.Context, to common.Address, contractAbi, methodName string, args ...interface{}) (txID IDTX, done chan bool, err error)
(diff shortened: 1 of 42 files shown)


identity/did.go
```diff
@@ -211,10 +211,10 @@ type Service interface {
 	GetKey(did DID, key [32]byte) (*KeyResponse, error)
 
 	// RawExecute calls the execute method on the identity contract
-	RawExecute(ctx context.Context, to common.Address, data []byte, gasLimit uint64) (txID IDTX, done chan bool, err error)
+	RawExecute(ctx context.Context, to common.Address, data []byte, gasLimit uint64) (txID IDTX, done chan error, err error)
```
vimukthi-git: please review variable names for the changed type

jobs/job.go
```diff
@@ -174,7 +174,7 @@ type Config interface {
 // Manager is a manager for centrifuge Jobs.
 type Manager interface {
 	// ExecuteWithinJob executes the given unit of work within a Job
-	ExecuteWithinJob(ctx context.Context, accountID identity.DID, existingJobID JobID, desc string, work func(accountID identity.DID, jobID JobID, jobManager Manager, err chan<- error)) (jobID JobID, done chan bool, err error)
+	ExecuteWithinJob(ctx context.Context, accountID identity.DID, existingJobID JobID, desc string, work func(accountID identity.DID, jobID JobID, jobManager Manager, err chan<- error)) (jobID JobID, done chan error, err error)
```
vimukthi-git: could you check the if the `err chan<- error` is needed here anymore? I think we only need one.

mikiquantum: I believe we still need the inner job error channel. since it tells asynchronously when the actual work is done. Afterwards the wrapper job logic needs a different channel to publish the result of it, since the `ExecuteWithinJob` adds more logic on top of the `go work()` step

identity/did.go
mikiquantum: Can you suggest a name that fits the meaning of job completion and holds an error if triggered?

PR #1232, opened 2019-07-26
add status to the document

### LINKED ISSUE #1220: Add status to documents.CoreDocument
- Add Staus field to Core Document. Model should expose Setter and Getter.
- Define Status type and should not be exposed. Created new status types - Pending, Committing, Committed
- Add a migration to fill the documents with missing status

vedhavyas:
Closes #1220

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -53,7 +53,19 @@ const (
 
 	// SignaturesTreePrefix is the human readable prefix for signature props
 	SignaturesTreePrefix = "signatures_tree"
+
+	// Pending status represents document is in pending state
+	Pending status = "pending"
+
+	// Committing status represents document is being committed.
+	Committing status = "committing"
+
+	// Committed status represents document is committed/anchored.
+	Committed status = "committed"
 )
+
+// status represents the document status.
+type status string
 
 // CompactProperties returns the compact property for a given prefix
 func CompactProperties(key string) []byte {
@@ -89,6 +101,9 @@ type CoreDocument struct {
 	// Attributes are the custom attributes added to the document
 	Attributes map[AttrKey]Attribute
 
+	// Status represents document status.
+	Status status
+
 	Document coredocumentpb.CoreDocument
 }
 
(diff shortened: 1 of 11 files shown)


documents/coredocument.go
```diff
@@ -195,6 +215,22 @@ func (cd *CoreDocument) NextVersion() []byte {
...
+func (cd *CoreDocument) GetStatus() status {
+	return cd.Status
+}
+
+// SetStatus set the status of the document.
+// if the document is already committed, returns an error if set status is called.
+func (cd *CoreDocument) SetStatus(st status) error {
+	if cd.Status == Committed && st != Committed {
+		return ErrCDStatus
+	}
```
mikiquantum: Should we add as well a transition check on `committing`? So a state lower than that one, like `pending` cannot be set either? 
On those lines, if the committing fails, should we update the model to `pending` again? or to errored?

documents/service.go
```diff
@@ -238,6 +243,11 @@ func (s service) ReceiveAnchoredDocument(ctx context.Context, model Model, colla
 		return errors.NewTypedError(ErrDocumentInvalid, err)
 	}
 
+	// set the status to committed since the document is anchored already.
```
mikiquantum: I think we need as well to change the state to `Committed` on the sender node?

documents/coredocument.go
vedhavyas: > On those lines, if the committing fails, should we update the model to pending again? or to errored?
This should be handled by the caller. We cannot assume it to be Pending or Committing

documents/service.go
vedhavyas: We cant, sender node can be a different implementation altogether. Status should be within the local storage IMO else protocol will know what is a status :-D

documents/coredocument.go
vedhavyas: > Should we add as well a transition check on committing? So a state lower than that one, like pending cannot be set either?

I dont think this is needed ATM since the committing state will be seet by the `Commit` function of the documents Service. and Pending state is handled by the pending document service

PR #1233, opened 2019-07-29
add invoice derive

vedhavyas:


documents/invoice/model.go
```diff
@@ -601,6 +601,29 @@ func (i *Invoice) unpackFromUpdatePayload(old *Invoice, payload documents.Update
 	return nil
 }
 
+// unpackFromUpdatePayloadOld unpacks the update payload and prepares a new version.
+func (i *Invoice) unpackFromUpdatePayload(payload documents.UpdatePayload) (*Invoice, error) {
+	var d Data
+	err := copier.Copy(&d, &i.Data)
```
mikiquantum: deep copy for complex structs?

Gopkg.toml
```diff
@@ -309,3 +309,7 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 [[constraint]]
   name = "github.com/urfave/cli"
   version = "1.20.0"
+
+[[constraint]]
+  branch = "master"
```
mikiquantum: Can we lock here a version or revision? In case that we need to regenerate the lock file.

documents/invoice/model.go
vedhavyas: yes, this way we dont have to old invoice data which on memory doesn't share the same data with the new one. One use case for this is validators

PR #1234, opened 2019-07-30
Add commit to pending service

### LINKED ISSUE #1222: Add Commit method to documents.Service
Add an achor method to service that does, validaiions, make document status to `committing`, store the document, initiate the anchoring and return jobID.

Once the document is anchored, we should also set the status to Committed.

mikiquantum:
closes #1222

### DIFF
--- a/resources/data.go
+++ b/resources/data.go
@@ -204,6 +204,7 @@ type bintree struct {
 	Func     func() (*asset, error)
 	Children map[string]*bintree
 }
+
 var _bintree = &bintree{nil, map[string]*bintree{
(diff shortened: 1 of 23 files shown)


resources/data.go
```diff
@@ -204,6 +204,7 @@ type bintree struct {
 	Func     func() (*asset, error)
 	Children map[string]*bintree
 }
+
```
vedhavyas: did we regenerate this ?

mikiquantum: just ran fmt on the whole project

PR #1238, opened 2019-08-01
add create

### LINKED ISSUE #1224: PendingDocumentService - Create
- Inject documetns.Service to pending doc service
- Add Create method which delegates the model derivation to dcuments.Service. Once model is derived, we store the model in PendingRepo and return the model

vedhavyas:
Closes #1224

### DIFF
--- a/pending/service.go
+++ b/pending/service.go
@@ -5,14 +5,22 @@ import (
 
 	"github.com/centrifuge/go-centrifuge/contextutil"
 	"github.com/centrifuge/go-centrifuge/documents"
+	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-centrifuge/jobs"
 	logging "github.com/ipfs/go-log"
 )
 
 var srvLog = logging.Logger("pending-service")
 
+// ErrPendingDocumentExists is a sentinel error used when document was created and tried to create a new one.
+const ErrPendingDocumentExists = errors.Error("Pending document already created")
+
 // Service provides an interface for functions common to all document types
 type Service interface {
+
+	// Create creates a pending document from the payload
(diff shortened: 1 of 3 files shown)


pending/service.go
```diff
@@ -31,6 +39,32 @@ func DefaultService(docSrv documents.Service, repo Repository) Service {
...
+// errors out if there an pending document created already
+func (s service) Create(ctx context.Context, payload documents.UpdatePayload) (documents.Model, error) {
+	accID, err := contextutil.AccountDID(ctx)
+	if err != nil {
+		return nil, contextutil.ErrDIDMissingFromContext
+	}
+
+	if len(payload.DocumentID) > 0 {
+		_, err := s.pendingRepo.Get(accID[:], payload.DocumentID)
+		if err == nil {
```
mikiquantum: Should we start being a more explicit on doc not found vs other DB errors? WDYT

pending/service.go
```diff
@@ -31,6 +39,32 @@ func DefaultService(docSrv documents.Service, repo Repository) Service {
 	}
 }
 
+// Create creates either a new document or next version of an anchored document and stores the document.
+// errors out if there an pending document created already
+func (s service) Create(ctx context.Context, payload documents.UpdatePayload) (documents.Model, error) {
```
mikiquantum: Do you think we could rename `documents.UpdatePayload` to something like `documents.UpsertPayload` or simply `documents.Payload` ?

vedhavyas: yes. Lets do it during the cleanup since this is being used by V1 as well?

pending/service.go
vedhavyas: yes, but in this scenario, we will fail if the doc is found in the DB since we cant have multiple

PR #1242, opened 2019-08-01
Pending service update

### LINKED ISSUE #1225: PendingDocServic - Update
- Add Update to Service. Delegate the model Derivation to documents.Service Derive
- Store the model and return the model

mikiquantum:
closes #1225

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -241,8 +241,8 @@ func (cd *CoreDocument) AppendSignatures(signs ...*coredocumentpb.Signature) {
 	cd.Modified = true
 }
 
-// PrepareNewData overrides only core document data without provisioning new versions since for document updates
-func (cd *CoreDocument) PrepareNewData(documentPrefix []byte, collaborators CollaboratorsAccess, attrs map[AttrKey]Attribute) (*CoreDocument, error) {
+// Patch overrides only core document data without provisioning new versions since for document updates
+func (cd *CoreDocument) Patch(documentPrefix []byte, collaborators CollaboratorsAccess, attrs map[AttrKey]Attribute) (*CoreDocument, error) {
 	cdp := coredocumentpb.CoreDocument{
 		DocumentIdentifier: cd.Document.DocumentIdentifier,
 		CurrentVersion:     cd.Document.CurrentVersion,
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -243,6 +243,10 @@ func (cd *CoreDocument) AppendSignatures(signs ...*coredocumentpb.Signature) {
 
 // Patch overrides only core document data without provisioning new versions since for document updates
 func (cd *CoreDocument) Patch(documentPrefix []byte, collaborators CollaboratorsAccess, attrs map[AttrKey]Attribute) (*CoreDocument, error) {
+	if cd.Status == Committing || cd.Status == Committed {
+		return nil, ErrDocumentNotInAllowedState
+	}
+
 	cdp := coredocumentpb.CoreDocument{
 		DocumentIdentifier: cd.Document.DocumentIdentifier,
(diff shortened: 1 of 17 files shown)


documents/invoice/model.go
```diff
@@ -624,6 +624,15 @@ func (i *Invoice) unpackFromUpdatePayload(payload documents.UpdatePayload) (*Inv
 	}, nil
 }
 
+// Patch merges payload data into model
+func (i *Invoice) Patch(payload documents.UpdatePayload) (documents.Model, error) {
+	inv, err := i.unpackFromUpdatePayload(payload)
```
vedhavyas: I see a problem here. `unpackFromUpdatePayload` internally calls `PreparesNewVersion` making it skip the current version without anchoring it.

I think we need to split the unpackFromUpdatePayload to patching and preparing new version. WDYT ?

documents/service.go
```diff
@@ -28,6 +28,12 @@ type DocumentProof struct {
 	FieldProofs []*proofspb.Proof
 }
 
+// Patcher interface defines a Patch method for inner Models
+type Patcher interface {
+	// Patch merges payload data into model
+	Patch(payload UpdatePayload) (Model, error)
```
vedhavyas: Since we anyway call this on the model, we dont need to return the model I think

pending/service.go
```diff
@@ -65,6 +71,35 @@ func (s service) Create(ctx context.Context, payload documents.UpdatePayload) (d
...
+	if err != nil {
+		return nil, contextutil.ErrDIDMissingFromContext
+	}
+
+	m, err := s.pendingRepo.Get(accID[:], payload.DocumentID)
+	if err != nil {
+		return nil, err
+	}
+
+	if m.GetStatus() != documents.Pending {
```
vedhavyas: this check is will always fail, no ?
since the pending repo will always store the pending documents, WDYT ?

mikiquantum: right, since the create will take care of the already anchored flow.

documents/invoice/model.go
mikiquantum: you are right, didn't catch that, I will separate those now.

documents/coredocument.go
```diff
@@ -241,6 +241,41 @@ func (cd *CoreDocument) AppendSignatures(signs ...*coredocumentpb.Signature) {
 	cd.Modified = true
 }
 
+// PrepareNewData overrides only core document data without provisioning new versions since for document updates
+func (cd *CoreDocument) PrepareNewData(documentPrefix []byte, collaborators CollaboratorsAccess, attrs map[AttrKey]Attribute) (*CoreDocument, error) {
```
vedhavyas: I think we should check for status==pending only. WDYT ?

documents/error.go
```diff
@@ -50,6 +50,9 @@ const (
 	// ErrDocumentProof must be used when document proof creation fails
 	ErrDocumentProof = errors.Error("document proof error")
 
+	// ErrNotPatcher must be used if an expected patcher model is not a patcher
+	ErrNotPatcher = errors.Error("model not patcher")
```
vedhavyas:
```suggestion
	ErrNotPatcher = errors.Error("document doesn't support patching")
```

pending/service.go
```diff
@@ -15,9 +15,15 @@ var srvLog = logging.Logger("pending-service")
 // ErrPendingDocumentExists is a sentinel error used when document was created and tried to create a new one.
 const ErrPendingDocumentExists = errors.Error("Pending document already created")
 
+// ErrInProcessDocument is a sentinel error used when a document is not in pending state
+const ErrInProcessDocument = errors.Error("document is not in pending state")
```
vedhavyas: Forgot to use this in the CoreDocument PrepareNewData ?

PR #1245, opened 2019-08-06
add update v2 API

### LINKED ISSUE #1229: Implement Update API
Add Update API to v2

vedhavyas:
Closes #1229

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -263,7 +263,7 @@ func (cd *CoreDocument) Patch(documentPrefix []byte, collaborators Collaborators
 	wcs := collaborators.ReadWriteCollaborators
 	rcs = append(rcs, wcs...)
 
-	ncd := &CoreDocument{Document: cdp}
+	ncd := &CoreDocument{Document: cdp, Status: Pending}
 	ncd.addCollaboratorsToReadSignRules(rcs)
 	ncd.addCollaboratorsToTransitionRules(documentPrefix, wcs)
 	// TODO convert it back to override when we have implemented add/delete for attributes in API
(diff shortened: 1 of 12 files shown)


httpapi/v2/documents_api.go
```diff
@@ -72,3 +79,66 @@ func (h handler) CreateDocument(w http.ResponseWriter, r *http.Request) {
...
+// @tags Documents
+// @accept json
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param body body v2.UpdateDocumentRequest true "Document Update request"
+// @param document_id path string true "Document Identifier"
+// @produce json
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 404 {object} httputils.HTTPError
+// @Failure 500 {object} httputils.HTTPError
+// @Failure 403 {object} httputils.HTTPError
```
mikiquantum: Do we need error code 403 here? Looks like this API doesnt produce this error?

documents/coredocument.go
```diff
@@ -263,7 +263,7 @@ func (cd *CoreDocument) Patch(documentPrefix []byte, collaborators Collaborators
 	wcs := collaborators.ReadWriteCollaborators
 	rcs = append(rcs, wcs...)
 
-	ncd := &CoreDocument{Document: cdp}
+	ncd := &CoreDocument{Document: cdp, Status: Pending}
```
mikiquantum: Since we are already setting this value in the `documents` package in other places, lets keep it like this, but lets review this approach soon.

httpapi/v2/documents_api.go
vedhavyas: it will if the auth is not passed

documents/coredocument.go
vedhavyas: this is a hack until we move the pending status from documents since patch is not setting status

PR #1246, opened 2019-08-06
commit api

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/httpapi/router.go
+++ b/httpapi/router.go
@@ -23,7 +23,7 @@ import (
 // Router returns the http mux for the server.
 // @title Centrifuge OS Node API
 // @description Centrifuge OS Node API
-// @version 0.0.5
+// @version 0.0.6
 // @contact.name Centrifuge
 // @contact.url https://github.com/centrifuge/go-centrifuge
 // @contact.email hello@centrifuge.io
--- a/httpapi/v2/documents_api.go
+++ b/httpapi/v2/documents_api.go
@@ -7,6 +7,8 @@ import (
 	"github.com/centrifuge/go-centrifuge/jobs"
 	"github.com/centrifuge/go-centrifuge/utils/byteutils"
 	"github.com/centrifuge/go-centrifuge/utils/httputils"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/go-chi/chi"
 	"github.com/go-chi/render"
 )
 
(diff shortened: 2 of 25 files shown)


httpapi/router.go
```diff
@@ -23,7 +23,7 @@ import (
 // Router returns the http mux for the server.
 // @title Centrifuge OS Node API
 // @description Centrifuge OS Node API
-// @version 0.0.5
+// @version 0.0.6
```
vedhavyas: why the change in version? Did we break anything from the old APIs

httpapi/v2/documents_api.go
```diff
@@ -72,3 +74,49 @@ func (h handler) CreateDocument(w http.ResponseWriter, r *http.Request) {
...
+// @id commit_document_v2
+// @tags Documents
+// @accept json
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param document_id path string true "Document Identifier"
+// @produce json
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 500 {object} httputils.HTTPError
+// @Failure 403 {object} httputils.HTTPError
+// @success 201 {object} coreapi.DocumentResponse
```
vedhavyas: the status should be accepted 202. WDYT ?

pending/test_pending.go
```diff
@@ -5,6 +5,8 @@ package pending
 import (
 	"context"
 
+	"github.com/centrifuge/go-centrifuge/jobs"
+
```
vedhavyas: import space

testworld/v2_documents_test.go
```diff
@@ -51,3 +51,28 @@ func TestV2DocumentCreate_next_version(t *testing.T) {
 	// alice should not have this version
 	nonExistingDocumentVersionCheck(alice.httpExpect, alice.id.String(), "documents", params)
```
vedhavyas: can you adapt commit to the functions NewDocument and Exisitng document?
We already have created a pending document at this point instead of re-creating a new one ?

mikiquantum: I rather keep it separate since it is a separate API call independent of the other tests. Easier to track.

httpapi/v2/documents_api.go
mikiquantum: yes!

httpapi/router.go
mikiquantum: 0.0.5 is what we released as part of Eunomia, and we have been overwritting all over, since the release (pushing to swaggerhub). We need to be more careful when cutting the release to change the version on the first commit after.

testworld/v2_documents_test.go
vedhavyas: the new test is making the same API calls as the new one. Adapting to old tests ensures we do not have a lot of tests doing the same. WDYT ?

PR #1247, opened 2019-08-07
Get and Get version API

### LINKED ISSUE #1231: Add Get and GetVersion v2

vedhavyas:
Closes #1231

### DIFF
--- a/httpapi/v2/documents_api.go
+++ b/httpapi/v2/documents_api.go
@@ -258,3 +258,56 @@ func (h handler) GetPendingDocument(w http.ResponseWriter, r *http.Request) {
 func (h handler) GetCommittedDocument(w http.ResponseWriter, r *http.Request) {
 	h.getDocumentWithStatus(w, r, documents.Committed)
 }
+
+// GetDocumentVersion returns the specific version of the document.
+// @summary Returns the specific version of the document.
+// @description Returns the specific version of the document.
+// @id get_document_version_v2
+// @tags Documents
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param document_id path string true "Document Identifier"
+// @param version_id path string true "Document Version Identifier"
+// @produce json
+// @Failure 403 {object} httputils.HTTPError
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 404 {object} httputils.HTTPError
+// @Failure 500 {object} httputils.HTTPError
+// @success 200 {object} coreapi.DocumentResponse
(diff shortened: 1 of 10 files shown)


httpapi/v2/documents_api.go
```diff
@@ -188,3 +189,125 @@ func (h handler) Commit(w http.ResponseWriter, r *http.Request) {
...
+// @tags Documents
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param document_id path string true "Document Identifier"
+// @produce json
+// @Failure 403 {object} httputils.HTTPError
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 404 {object} httputils.HTTPError
+// @Failure 500 {object} httputils.HTTPError
+// @success 200 {object} coreapi.DocumentResponse
+// @router /v2/documents/{document_id}/pending [get]
```
mikiquantum: Do you think we could define an abstract getdoc endpoint? Meaning:
`@router /v2/documents/{document_id}/{status} [get]`
and add a comment as:
`@param status path string true "status" oneof pending|committed`
Or will it create collisions on other paths?

httpapi/v2/handler.go
```diff
@@ -21,4 +21,7 @@ func Register(ctx map[string]interface{}, r chi.Router) {
 	r.Post("/documents", h.CreateDocument)
 	r.Patch("/documents/{"+coreapi.DocumentIDParam+"}", h.UpdateDocument)
 	r.Post("/documents/{"+coreapi.DocumentIDParam+"}/commit", h.Commit)
+	r.Get("/documents/{"+coreapi.DocumentIDParam+"}/pending", h.GetPendingDocument)
+	r.Get("/documents/{"+coreapi.DocumentIDParam+"}/committed", h.GetCommittedDocument)
```
mikiquantum: if a document fails to anchor and the `committed` state is not set in the document, can we still fetch the document that is stuck in committing? I believe the `committed` will still retrieve the committing document, right?

httpapi/v2/documents_api.go
vedhavyas: This was done intentional because we want to allow only two kinds of the status. Having `status` of the as url path param doesn't benefit the client and makes us having more checks for status.
And this will cause collisions if we introduce a different API with similar  path.

httpapi/v2/handler.go
vedhavyas: yes, that will still fetch the document even if its not committed.

PR #1248, opened 2019-08-08
fixed length timestamp and decimal

mikiquantum:
Issue Link: <Github link>

### DIFF
--- a/documents/converters.go
+++ b/documents/converters.go
@@ -1,6 +1,8 @@
 package documents
 
 import (
+	"bytes"
+	"encoding/binary"
 	"strings"
 	"time"
 
@@ -9,7 +11,10 @@ import (
 	"github.com/centrifuge/go-centrifuge/identity"
 	"github.com/centrifuge/go-centrifuge/utils/byteutils"
 	"github.com/centrifuge/go-centrifuge/utils/timeutils"
+	"github.com/golang/protobuf/ptypes/timestamp"
 )
+
+const maxTimeByteLength = 8
 
 // BinaryAttachment represent a single file attached to invoice.
 type BinaryAttachment struct {
@@ -195,7 +200,13 @@ func toProtocolAttributes(attrs map[AttrKey]Attribute) (pattrs []*coredocumentpb
 		case AttrBytes:
 			pattr.Value = &coredocumentpb.Attribute_ByteVal{ByteVal: attr.Value.Bytes}
 		case AttrTimestamp:
-			pattr.Value = &coredocumentpb.Attribute_TimeVal{TimeVal: attr.Value.Timestamp}
+			buf := new(bytes.Buffer)
+			err := binary.Write(buf, binary.BigEndian, attr.Value.Timestamp.Seconds)
+			if err != nil {
+				return nil, err
+			}
+			b := append(make([]byte, maxTimeByteLength-len(buf.Bytes())), buf.Bytes()...)
+			pattr.Value = &coredocumentpb.Attribute_ByteVal{ByteVal: b}
 		case AttrSigned:
 			signed := attr.Value.Signed
 			pattr.Value = &coredocumentpb.Attribute_SignedVal{
@@ -269,7 +280,12 @@ func attrValFromProtocolAttribute(attrType AttributeType, attribute *coredocumen
 	case AttrBytes:
 		attrVal.Bytes = attribute.GetByteVal()
 	case AttrTimestamp:
-		attrVal.Timestamp = attribute.GetTimeVal()
+		var n int64
+		err := binary.Read(bytes.NewBuffer(attribute.GetByteVal()), binary.BigEndian, &n)
+		if err != nil {
+			return attrVal, err
+		}
+		attrVal.Timestamp = &timestamp.Timestamp{Seconds: n}
 	case AttrSigned:
 		val := attribute.GetSignedVal()
 		did, err := identity.NewDIDFromBytes(val.Identity)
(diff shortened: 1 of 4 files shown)


documents/converters.go
```diff
@@ -195,7 +200,13 @@ func toProtocolAttributes(attrs map[AttrKey]Attribute) (pattrs []*coredocumentpb
 		case AttrBytes:
 			pattr.Value = &coredocumentpb.Attribute_ByteVal{ByteVal: attr.Value.Bytes}
 		case AttrTimestamp:
-			pattr.Value = &coredocumentpb.Attribute_TimeVal{TimeVal: attr.Value.Timestamp}
+			buf := new(bytes.Buffer)
+			err := binary.Write(buf, binary.BigEndian, attr.Value.Timestamp.Seconds)
```
vedhavyas: We are ignoring the nanoseconds here. Lets embed nanoseconds and increase the `maxTimeByteLength` to 12?

vedhavyas:
```suggestion
			binary.BigEndian.PutUint64(buf[:], attr.Value.Timestamp.Seconds)
```

documents/converters.go
```diff
@@ -195,7 +200,13 @@ func toProtocolAttributes(attrs map[AttrKey]Attribute) (pattrs []*coredocumentpb
 		case AttrBytes:
 			pattr.Value = &coredocumentpb.Attribute_ByteVal{ByteVal: attr.Value.Bytes}
 		case AttrTimestamp:
-			pattr.Value = &coredocumentpb.Attribute_TimeVal{TimeVal: attr.Value.Timestamp}
+			buf := new(bytes.Buffer)
```
vedhavyas:
```suggestion
			var buf [maxTimeByteLength]byte
```

documents/converters.go
```diff
@@ -195,7 +200,13 @@ func toProtocolAttributes(attrs map[AttrKey]Attribute) (pattrs []*coredocumentpb
...
+			err := binary.Write(buf, binary.BigEndian, attr.Value.Timestamp.Seconds)
+			if err != nil {
+				return nil, err
+			}
+			b := append(make([]byte, maxTimeByteLength-len(buf.Bytes())), buf.Bytes()...)
```
vedhavyas: this check wont be necessary once we have a fixed-length byte slice :-D

documents/converters.go
```diff
@@ -269,7 +280,12 @@ func attrValFromProtocolAttribute(attrType AttributeType, attribute *coredocumen
 	case AttrBytes:
 		attrVal.Bytes = attribute.GetByteVal()
 	case AttrTimestamp:
-		attrVal.Timestamp = attribute.GetTimeVal()
+		var n int64
+		err := binary.Read(bytes.NewBuffer(attribute.GetByteVal()), binary.BigEndian, &n)
```
vedhavyas:
```suggestion
		n := binary.BigEndian.Uint64(attribute.GetByteVal())
```

documents/converters.go
mikiquantum: seconds are int64

mikiquantum: We intentionally left the nanoseconds out, since blockchains are not that accurate anyways. If you see a big problem, I am open to change it to 12.

vedhavyas: > is that simple casting var y uint64 = uint64(x)
Yees, it is a direct conversion. 

> We intentionally left the nanoseconds out, since blockchains are not that accurate anyways.
The attribute is also given by the user. ignoring nano will change the client data. Hence, I feel we should add nano as well.

mikiquantum: Actually, I agree there. I will add nanoseconds as well. On the other hand I rather keep using `binary.Write` and `binary.Read` since it is more safe in case the user input exceeds the internal representation and error out instead of `PutUintx` panicking.

mikiquantum: but mostly it is easier to deal with buffers.

documents/converters.go
```diff
@@ -9,8 +11,11 @@ import (
 	"github.com/centrifuge/go-centrifuge/identity"
 	"github.com/centrifuge/go-centrifuge/utils/byteutils"
 	"github.com/centrifuge/go-centrifuge/utils/timeutils"
+	"github.com/golang/protobuf/ptypes/timestamp"
 )
 
+const maxTimeByteLength = 12
```
lucasvo: Could you add a comment what this constant is for?

documents/converters.go
lucasvo: @vedhavyas I disagree, I think rounding to 0 nanoseconds is fine. we can always have a second type that is more accurate later.

documents/converters.go
lucasvo: If you supply nanoseconds in json, you're also not guaranteed to have the parser persist them.

documents/converters.go
vedhavyas: @lucasvo Why is losing precision on nano when user provided the timestamp in nano? I feel we shouldn't change the user-provided data here

vedhavyas: > but mostly it is easier to deal with buffers, since in this case we already know the size of the buffer as we only deal with int64/int32
But this wont panic if we have a fixed byte array instead of slice/buffer. But upto you though

documents/converters.go
vedhavyas: If you mean through timestamp nano, then go time parser will parse the nano as well

documents/converters.go
mikiquantum: @lucasvo if nano is not provided by the user when defining the data parameter `2006-01-02T15:04:05Z07:00` then nanos will be zero. If they decide to add the nano precision: `2006-01-02T15:04:05.999999999Z07:00` then it will be part of the byte array serialized. I think is a free option for us to keep the nanos as well.

PR #1249, opened 2019-08-08
optimize proofs for generic nft

mikiquantum:


nft/service.go
```diff
@@ -117,12 +119,20 @@ func (s *service) prepareMintRequest(ctx context.Context, tokenID TokenID, cid i
 		return mreq, err
 	}
 
+	optProofs := docProofs.FieldProofs
```
mikiquantum: main change

PR #1264, opened 2019-09-06
Feat/signed attr api

vedhavyas:
Issue Link: <Github link>

### DIFF
--- a/documents/attribute.go
+++ b/documents/attribute.go
@@ -275,7 +275,7 @@ func NewMonetaryAttribute(keyLabel string, value *Decimal, chainID []byte, id st
 // doc version is next version of the document since that is the document version in which the attribute is added.
 // signature payload: sign(identity + docID + docNextVersion + value)
 // Note: versionID should always be the next version that is going to be anchored.
-func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, docID, versionID []byte, value []byte) (attr Attribute, err error) {
+func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, docID, versionID, value []byte) (attr Attribute, err error) {
 	attrKey, err := AttrKeyFromLabel(keyLabel)
 	if err != nil {
 		return attr, err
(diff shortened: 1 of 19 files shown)


documents/attribute.go
```diff
@@ -273,14 +273,15 @@ func NewMonetaryAttribute(keyLabel string, value *Decimal, chainID []byte, id st
 // NewSignedAttribute returns a new signed attribute
 // takes keyLabel, signer identity, signer account, model and value
 // doc version is next version of the document since that is the document version in which the attribute is added.
-// signature payload: sign(identity + docID + docVersion + value)
-func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, model Model, value []byte) (attr Attribute, err error) {
+// signature payload: sign(identity + docID + docNextVersion + value)
+// Note: versionID should always be the next version that is going to be anchored.
+func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, docID, versionID []byte, value []byte) (attr Attribute, err error) {
```
mikiquantum:
```suggestion
func NewSignedAttribute(keyLabel string, identity identity.DID, account config.Account, docID, versionID, value []byte) (attr Attribute, err error) {
```

httpapi/v2/attribute_api.go
```diff
@@ -0,0 +1,72 @@
+package v2
```
mikiquantum: Since the root of the operation is over a `document` can we add this logic to the `documents_api.go` file, or do you have any other plans around this new organization?

vedhavyas: I didn't want to add attribute related stuff to documents_api to keep it seperate and may potentially have other Attribute apis later

vedhavyas:
> About the user flow here. As a user I would need to first create a draft of an existing committed document and then sign an attribute right? Wouldn't be interesting doing both in one shot?

Yes, we can do it in a single shot but it would require us to keep the same logic for other similar APIs as well. I would prefer if we have a single API to create a pending doc and add other little actions like signed attribute individually through individual APIs

PR #1266, opened 2019-09-10
Remove Collaborator API

vedhavyas:
Issue Link: <Github link>

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -917,7 +917,6 @@ func (cd *CoreDocument) RemoveCollaborators(dids []identity.DID) error {
 	}
 
 	// remove each collaborator from the roles
-	var found bool
 	for _, did := range dids {
 		for _, role := range cd.Document.Roles {
 			i, f := isDIDInRole(role, did)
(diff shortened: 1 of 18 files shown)


documents/coredocument.go
```diff
@@ -154,6 +155,8 @@ func NewCoreDocument(documentPrefix []byte, collaborators CollaboratorsAccess, a
 
 	collaborators.ReadCollaborators = identity.RemoveDuplicateDIDs(collaborators.ReadCollaborators)
 	collaborators.ReadWriteCollaborators = identity.RemoveDuplicateDIDs(collaborators.ReadWriteCollaborators)
+	// remove any dids that are present in both read and read write from read.
+	collaborators.ReadCollaborators = filterCollaborators(collaborators.ReadCollaborators, collaborators.ReadWriteCollaborators...)
```
mikiquantum: was this a bug?

documents/coredocument.go
```diff
@@ -905,3 +908,32 @@ func (cd *CoreDocument) UnmarshalJSON(data []byte, m Model) error {
...
+			}
+
+			cd.Modified = true
+			found = true
+			role.Collaborators = byteutils.CutFromSlice(role.Collaborators, i)
+		}
+	}
+
+	if !found {
+		return errors.New("DIDs not found in the Document")
```
mikiquantum: What do you think about returning a `noop` instead of erroring out? Just a thought.

documents/coredocument.go
vedhavyas: yeap

documents/coredocument.go
vedhavyas: I was thinking that this is a mutated call. If we give noop, we might end up updating the next version with no changes. Hence, I thought we should let user know. Not that I'm thinking, it is a pending doc. Should be okay to return a noop as well

PR #1287, opened 2019-10-17
Add Document Transition Validation Flag to signature

### LINKED ISSUE #1279: Add transition Validation flag logic
### Transition Validation Flag
As part of the data to be signed, we will add a trailing byte to the payload.

| dataRoot | Val. Flag |
| -------- | --------  |

Flag status:
**0** -> Couldn't find or validate the transition validations based on a previous version of the document
**1** -> Successfully validated transition validations from a previous version of the document

##### go-centrifuge
Protobuf changes:
```javascript=
message Signature {
    ...
    // `transition_validated` defines if node was able to validate transition rules from document version A -> B
  bool transition_validated = 5;
}
```

Node
```go=
// Receiver side
func RequestDocumentSignature(req) (Signature, error) {
    new := req.Document
    old := repo.GetModel(model.PreviousVersion)
    err := validate(new, old)
    if err != nil {
        return err
    }
    
    transitionFlag := byte(0)
    if old != nil {
        transitionFlag := byte(1)
    }
    
    signature := acc.SignMsg(dataRoot+transitionFlag)
    ...
    return signature
}

// Sender side
func ValidateSignature(did, docRoot, sig) error {
    flag := sig.TransitionValidated
    payload := DocConsensusSignature(docRoot, flag)
    
    return idService.ValidateSignature(did, sig.Pk, payload)
}
```

mikiquantum:
closes #1279

### DIFF
--- a/documents/processor.go
+++ b/documents/processor.go
@@ -107,7 +107,7 @@ func (dp defaultProcessor) PrepareForSignatureRequests(ctx context.Context, mode
 		return errors.New("failed to calculate signing root: %v", err)
 	}
 
-	sig, err := self.SignMsg(ConsensusSignaturePayload(sr, byte(0)))
+	sig, err := self.SignMsg(ConsensusSignaturePayload(sr, false))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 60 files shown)


documents/processor.go
```diff
@@ -275,3 +275,12 @@ func (dp defaultProcessor) SendDocument(ctx context.Context, model Model) error
 
 	return err
 }
+
+// ConsensusSignaturePayload forms the payload needed to be signed during the document consensus flow
+func ConsensusSignaturePayload(dataRoot []byte, validated bool) []byte {
+	tFlag := byte(0)
```
vedhavyas: do we need this exposed ?

mikiquantum: Various other packages need it.

PR #1290, opened 2019-10-22
add simple nft

### LINKED ISSUE #1286: Mint Ethereum NFT using generic/simple registry
- Use new abi (generic) to integrate and pass correspondent params
- Node calculates bundled hash as asset to pass to mint (instead of getting it from events)
- Deploy generic smart contract below as part of tests setup
- Compliant with new protocol changes

vedhavyas:
Closes #1286


build/scripts/migrateDApp.sh
```diff
@@ -34,8 +27,8 @@ export ETH_RPC_URL=$CENT_ETHEREUM_NODEURL
 export ETH_PASSWORD="/dev/null"
 export ETH_FROM="0x89b0a86583c4444acfd71b463e0d3c55ae1412a5"
 
-regAddr=$(dapp create "TestNFT" "$ANCHOR_ADDR")
+regAddr=$(dapp create "NFT" "CentNFT" "CentNFT")
```
mikiquantum: Once we have the bridge assets populated we will have to pass here the ASSET_ADDR, for now it is fine.

nft/service.go
```diff
@@ -30,7 +32,7 @@ const (
...
..."name\":\"to\",\"type\":\"address\"},{\"indexed\":false,\"internalType\":\"uint256\",\"name\":\"tokenId\",\"type\":\"uint256\"}],\"name\":\"Minted\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"internalType\":\"address\",\"name\":\"from\",\"type\":\"address\"},{\"indexed\":true,\"internalType\":\"address\",\"name\":\"to\",\"type\":\"address\"},{\"indexed\":true,\"internalType\":\"uint256\",\"name\":\"tokenId\",\"type\":\"uint256\"}],\"name\":\"Transfer\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"internalType\":\"address\",\"name\":\"owner\",\"type\":\"address\"},{\"indexed\":true,\"internalType\":\"address\",\"name\":\"approved\",\"type\":\"address\"},{\"indexed\":true,\"internalType\":\"uint256\",\"name\":\"tokenId\",\"type\":\"uint256\"}],\"name\":\"Approval\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"internalType\":\"address\",\"name\":\"owner\",\"type\":\"address\"},{\"indexed\":true,\"internalType\":\"address\",\"name\":\"operator\",\"type\":\"address\"},{\"indexed\":false,\"internalType\":\"bool\",\"name\":\"approved\",\"type\":\"bool\"}],\"name\":\"ApprovalForAll\",\"type\":\"event\"}]"
```
mikiquantum: Can you extract only the mint method abi? since it is the only abi that the mint call needs. We achieve more succinct definition and less dependencies in case of other function changes.

nft/service.go
```diff
@@ -470,6 +478,18 @@ func bindContract(address common.Address, client ethereum.Client) (*InvoiceUnpai
 	return NewInvoiceUnpaidContract(address, client.GetEthClient())
 }
 
+// getBundledHash returns the sha3 of the concat of to + (props+values+salts)
+func getBundledHash(to common.Address, props, values [][]byte, salts [][32]byte) []byte {
+	res := to.Bytes()
+	for i := 0; i < len(props); i++ {
+		h := append(props[i], values[i]...)
```
mikiquantum: One liner is a bit too confusing right?

build/scripts/migrateDApp.sh
vedhavyas: yeap!

nft/service.go
vedhavyas: added comments. should be clear :-D

Issue #1280, opened 2019-10-15
New Tree structure


mikiquantum:
Tree Hashing:
- Leaf Node: sha3
- Internal Node: blake2b
- Modify `tree.NewDefaultTree()` with correspondent options.

Account Sign:
```go=
func (acc *Account) SignMsg(key KeyType, msg []byte) (*coredocumentpb.Signature, error) {
switch(key):
case "secp":
    return signSecp(msg)
case "jub": // Not implemented yet
    return signJub(msg)    
}
```
Request Signatures:
```go=
func RequestDocumentSignature() {
    ...
    ethDR := model.GetEthDataRoot()
    zkDR := model.GetZKDataRoot()
    
    sigs := make([]*coredocumentpb.Signature, 2)
    sigs[0] := acc.SignMsg(key.Secp, ethDR)
    sigs[1] := acc.SignMsg(key.Jub, zkDR)
    
    model.AppendSignatures(sigs...)
    ...
}
```
Validate Signatures:
```go=
func ValidateSignature(did, docRoot, sig) error {
    flag := sig.TransitionValidated
    payload := DocConsensusSignature(docRoot, flag)
    // Adding Type of signature
    return idService.ValidateSignature(did, sig.Type, sig.Pk, payload)
}
```
Extend the proto msg to include the curve type, excluding from tree, only as p2p message. 
Or have a try-error approach on the validator invoker that at least with one key needs to be validated.

mikiquantum:
By #1293

PR #1293, opened 2019-10-28
Tree Split

mikiquantum:
closes: https://github.com/centrifuge/go-centrifuge/issues/1280


mikiquantum:
@vedhavyas I didnt change anything on the submodule. Not sure why the commits have changed there, but the gitmodule branch is still the one that you specified and there are no diffs.

As I mentioned to you offline, I will add more tests to the PR soon.

PR #1301, opened 2019-11-05
Update/apis

### LINKED ISSUE #1297: Update API for account creation
Update APIs to take cent chain details.
Note: Generate account should accept cent chain account details.

vedhavyas:
Closes #1297


config/configstore/service.go
```diff
@@ -65,19 +65,23 @@ func (s service) CreateAccount(data config.Account) (config.Account, error) {
 	return data, s.repo.CreateAccount(id, data)
 }
 
-func (s service) GenerateAccount() (config.Account, error) {
+func (s service) GenerateAccount(cacc config.CentChainAccount) (config.Account, error) {
```
mikiquantum: Since this is invoked from the API, why not create the account with defaults (same as ethereum acc). The user would need to do an account update to change the default one (same as the eth one).
If what you want to do is omit this extra step, then I think it makes sense then to pass some AccountData instance containing as well other fields to override (from defaults)

vedhavyas: In cent chain, each account is not similar to an eth address. For us eth account pays for every account, but cent chain each account cannot pay for everyone since there is no concept of centrifuge identity. Generate account creates a new account and each account must have a unique cent chain account.

> If what you want to do is omit this extra step, then I think it makes sense then to pass some AccountData instance containing as well other fields to override (from defaults)
This is similar to Create API. If we go that route, then Generate API is just redundant

mikiquantum: So you cannot have just 1 cent-chain account per cent-node? Meaning all interactions of that cent-node use that account. My idea was to not enforce individual accounts per node but support it in the future (therefore my analogy with the eth account).

> This is similar to Create API. If we go that route, then Generate API is just redundant

Yes but not at the moment, since the account endpoints are just CRUD operations, and that would enforce the user to create identities/add keys ... outside of the scope of the createAccount. I believe that we will adapt that endpoint to not be CRUD, and be a hybrid between generateAccount and createAccount.
That is why I am referring now to just use one centchain account (the one in the config) and not touch the generateAccount externally, just internally.

vedhavyas: A single Cent chain account is equivalent to Centrifuge Identity on Eth. If we have a single cent account in a Node that is shared between multiple accounts A and B. If A does a pre-commit using the cent chain account, then B can B can commmit an anchor for that that pre commit which is not what we want. Hence, each account on node should have a cent chain accouont similar to an identity.

mikiquantum: Yes, that correct. Although that would mean that B needs to know the preimage, so it has to be a collaborator. In addition to be able to use the same centchain account needs to have access to the private key so the risk is reduced to the local node accounts only.
In any case, my main concern is with complexity on setting up a running node. Since the user now needs to know how to generate the centchain account in addition to the existing ethereum account creation.
Is there something we can do to abstract that from the user invoking the generateAccount?
We can jump on a call if that is easier/faster.

Issue #1281, opened 2019-10-15
GSRPC Integration -> SendAnchor Extrinsic - Task that listens to Extrinsic confirmation


mikiquantum:
- Define centchain package
- Define go-celery task that requests confirmation of tx/extrinsic
   - Use GSRPC library
- Unit testing logic

PR #1305, opened 2019-11-11
Extrinsic added to block listener

mikiquantum:
Closes: https://github.com/centrifuge/go-centrifuge/issues/1281

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -35,6 +35,10 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
   name = "github.com/centrifuge/centrifuge-ethereum-contracts"
   revision = "863c5efc0a49aff1e9c1cf2ef5f72f62c65425ed"
 
+[[override]]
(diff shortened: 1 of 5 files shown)


Gopkg.toml
```diff
@@ -245,7 +249,7 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
 
 [[constraint]]
   name = "golang.org/x/crypto"
-  revision = "0e37d006457bf46f9e6692014ba72ef82c33022c"
+  branch = "master"
```
mikiquantum: I rather lock a version, but since go-substrate-rpc-client uses master, I can't do in this PR. I will create another one agains the GSRPC repo to lock that crypto version

mikiquantum: https://github.com/centrifuge/go-substrate-rpc-client/pull/34

vedhavyas: You can replace `constraint` with `override`
That will lock the version we give here

PR #1306, opened 2019-11-12
submit extrisic

vedhavyas:
Issue Link: <Github link>

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -243,7 +243,7 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
   name = "github.com/whyrusleeping/go-smux-yamux"
   revision = "eac25f3e2d47aae211e457e7664b52634c95eea8"
 
-[[constraint]]
+[[override]]
   name = "golang.org/x/crypto"
   revision = "0e37d006457bf46f9e6692014ba72ef82c33022c"
 
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -36,6 +36,10 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
   revision = "863c5efc0a49aff1e9c1cf2ef5f72f62c65425ed"
 
 [[constraint]]
+  name = "github.com/centrifuge/go-substrate-rpc-client"
+  revision = "3a19d9c3f3899614dbbfd659e448ca859fc296fe"
+
+[[constraint]]
   name = "github.com/Masterminds/semver"
   version = "1.4.2"
 
@@ -243,9 +247,9 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
   name = "github.com/whyrusleeping/go-smux-yamux"
   revision = "eac25f3e2d47aae211e457e7664b52634c95eea8"
 
-[[override]]
+[[constraint]]
   name = "golang.org/x/crypto"
-  revision = "0e37d006457bf46f9e6692014ba72ef82c33022c"
+  revision = "4def268fd1a49955bfb3dda92fe3db4f924f2285"
 
 [[constraint]]
   name = "golang.org/x/net"
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ generate: ## autogenerate go files for config
 
 install-subkey: ## installs subkey
 	curl https://getsubstrate.io -sSf | bash -s -- --fast
+	source ~/.cargo/env
 	cargo install --force --git https://github.com/paritytech/substrate subkey
 
 vendorinstall: ## Installs all protobuf dependencies with go-vendorinstall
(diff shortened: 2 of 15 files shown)


Gopkg.toml
```diff
@@ -243,7 +243,7 @@ required = ["github.com/centrifuge/centrifuge-ethereum-contracts", "github.com/r
   name = "github.com/whyrusleeping/go-smux-yamux"
   revision = "eac25f3e2d47aae211e457e7664b52634c95eea8"
 
-[[constraint]]
+[[override]]
```
mikiquantum: I just merged the GSRPC PR so we can keep using constraint here.

Makefile
```diff
@@ -60,6 +60,11 @@ gen-swagger: ## generates the swagger documentation
 generate: ## autogenerate go files for config
 	go generate ./config/configuration.go
 
+install-subkey: ## installs subkey
+	curl https://getsubstrate.io -sSf | bash -s -- --fast
+	export PATH="$HOME/.cargo/bin:$PATH"
```
mikiquantum: I believe make will open a new terminal session, so export will not work. You might have to add this to the travis file.

anchors/chain_repository.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return txHash, bn, sig, err
+	}
+
+	meta, err := r.api.GetMetadataLatest()
+	if err != nil {
+		return txHash, bn, sig, err
+	}
+
+	c, err := types.NewCall(meta, "Anchor.pre_commit", types.NewHash(anchorID[:]), types.NewHash(signingRoot[:]))
```
mikiquantum: Maybe we can use constants for the Extrinsic methods, easier later for testing purposes, max gas usage references and so on.

centchain/api.go
```diff
@@ -0,0 +1,94 @@
...
+
+	auth := author.NewAuthor(a.getClient())
+	startBlock, err := a.getBlockLatest()
+	if err != nil {
+		return txHash, bn, sig, err
+	}
+
+	startBlockNumber := startBlock.Block.Header.Number
+	txHash, err = auth.SubmitExtrinsic(ext)
+	return txHash, startBlockNumber, ext.Signature.Signature, err
```
mikiquantum: Beautiful! 👍

centchain/test_centchain.go
```diff
@@ -0,0 +1,74 @@
...
+	txHash, _ = args.Get(0).(types.Hash)
+	bn, _ = args.Get(1).(types.BlockNumber)
+	sig, _ = args.Get(2).(types.Signature)
+	return txHash, bn, sig, args.Error(3)
+}
+
+func MetaDataWithCall(call string) *types.Metadata {
+	data := strings.Split(call, ".")
+	meta := types.NewMetadataV4()
+	meta.AsMetadataV4.Modules = []types.ModuleMetadataV4{
```
mikiquantum: I believe we are using MetadataV8, but since it is just mocking a flow should be fine right?

centchain/api.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return txHash, bn, sig, err
+	}
+
+	rv, err := a.getRuntimeVersionLatest()
+	if err != nil {
+		return txHash, bn, sig, err
+	}
+
+	key, err := types.CreateStorageKey(meta, "System", "AccountNonce", krp.PublicKey)
```
mikiquantum: I was wondering that we should probably add a lock to this method so we can ensure that the accountNonce is increased in the pool before sending it over. From my local testing sending multiple txs it is very possible that some of them will fail for this reason, WDYT?

philipstanislaus: @mikiquantum I did actually manually increase the nonce for the xts I sent in parallel, which also led to dropped xts.

mikiquantum: Yes, I did the same too. I was referring more in the future when we fix the problem we currently have in the centchain. 
At least from the same node (same accounts) we should be able to have a serial sending of txs so we can ensure the nonce increment, either by a lock (at the beginning) or something more complex in the future by keeping a state in the client node. Up to discussion.

philipstanislaus: Makes sense, so we lock in order to prevent 2 threads/workers using the same account with a nonce sequence?

mikiquantum: Exactly, it is not perfect, since it is possible that the go-centrifuge points to an infura like setup and the pending pool is not shared, it could still fail "silently". But it is a start.

philipstanislaus: Got it. On top of that, xts in pending pool might not increase the nonce - but we need to resend failed xts anyways.

mikiquantum: hmm, interesting one, I will research on that as soon as I get to the integration testing.

centchain/test_centchain.go
vedhavyas: yeap. Just for testing

centchain/api.go
vedhavyas: > Got it. On top of that, xts in pending pool might not increase the nonce - but we need to resend failed xts anyways.

This is big issue I feel. Wouldn't System.nonce take of this already?

Anyway, I added a lock for each extrinsic. This will ensure only one txn is sent at once.

Issue #1311, opened 2019-11-15
Fix tx nonce too low error so it can be retried on our geth client logic


mikiquantum:
Looks like the error when more than one concurrent transaction is sent , the error nonce is too low changed from:
`nonce too low` or `replacement transaction underpriced`
to image below.

![image.png](https://images.zenhubusercontent.com/5ba3f5945d222e04f2fae0ec/3144b215-96d0-4ea5-a43e-b99d35804c1b)

It is possible the is an infura like setup that is causing this error change. So we test out supporting both.

Issue #1312, opened 2019-11-18
Ensure go-centrifuge queues can retry ethereum task indefinitely (almost)


mikiquantum:
At the moment queues retrials are limited to 15, with a linear backoff retrial of 5 secs, which makes the total time until a tx is completed ~10mins. 
This is not enough for mainnet txs. 

We should add a higher configurable for ethereum tx status checks tasks. Ensure that go-celery supports any number.

PR #1314, opened 2019-11-19
Fix eth tx retrial message + gocelery max retrial

mikiquantum:
Closes: #1311 and #1312


ethereum/geth_client.go
```diff
@@ -347,10 +347,12 @@ func (gc *gethClient) SubmitTransactionWithRetries(contractMethod interface{}, o
 		}
 
 		if err == nil {
+			log.Infof("TX %s successfully sent with nonce %s", tx.Hash().String(), opts.Nonce.String())
 			return tx, nil
 		}
 
-		if (err.Error() == ErrTransactionUnderpriced.Error()) || (err.Error() == ErrNonceTooLow.Error()) {
+		if strings.Contains(err.Error(), ErrTransactionUnderpriced.Error()) || strings.Contains(err.Error(), ErrNonceTooLow.Error()) ||
```
mikiquantum: Fixing this will not ensure that all txs will always go through, since they can still fail silently in the processing eth node.

build/configs/default_config.yaml
```diff
@@ -152,7 +152,8 @@ ethereum:
     nftMint: 900000
     nftTransferFrom: 150000
   # Timeout to wait for an ethereum transaction to be added to a block and events triggered
-  contextWaitTimeout: "600s"
+  # Set to higher bound ~28 days
+  contextWaitTimeout: "672h"
```
mikiquantum: Since go-celery max limit is the same, makes sense to keep it here, since they go hand by hand

ethereum/geth_client.go
```diff
@@ -347,11 +347,12 @@ func (gc *gethClient) SubmitTransactionWithRetries(contractMethod interface{}, o
 		}
 
 		if err == nil {
+			log.Infof("TX %s successfully sent with nonce %s", tx.Hash().String(), opts.Nonce.String())
 			return tx, nil
 		}
 
-		if (err.Error() == ErrTransactionUnderpriced.Error()) || (err.Error() == ErrNonceTooLow.Error()) {
-			log.Warningf("Concurrent transaction identified, trying again [%d/%d]\n", current, maxTries)
+		if strings.Contains(err.Error(), ErrIncNonce.Error()) || strings.Contains(err.Error(), ErrIncPrice.Error()) {
```
mikiquantum: Fixing this will not ensure that all txs will always go through, since they can still fail silently in the processing eth node.

Issue #1283, opened 2019-10-15
GSRPC Integration -> SendAnchor Extrinsic - Integrate in protocol flow


mikiquantum:
- Update job functions that call new anchor tasks against centchain instead of ethereum

PR #1317, opened 2019-11-27
Integrate centchain for anchoring

mikiquantum:
closes: #1283


mikiquantum:
@vedhavyas / @philipstanislaus I expect some tests to fail since this PR is required to fix them: https://github.com/centrifuge/centrifuge-chain/pull/127
Please review, thank you!

anchors/chain_repository_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/centrifuge/go-centrifuge/jobs"
+	"github.com/centrifuge/go-centrifuge/testingutils/testingjobs"
+
```
vedhavyas: import space

anchors/chain_repository.go
```diff
@@ -99,8 +121,33 @@ func (r repository) Commit(
...
+
+// AnchorData holds data returned from previously anchored data against centchain
+type AnchorData struct {
+	AnchorID     types.Hash `json:"id"`
+	DocumentRoot types.Hash `json:"doc_root"`
+	BlockNumber  uint32     `json:"anchored_block"`
+}
+
+func (r repository) GetAnchorByID(id *big.Int) (*AnchorData, error) {
+	var ad *AnchorData
```
vedhavyas: is this a pointer to AnchorData?
We seem to pass the double pointer to the api.Call
Is that expected ?

.travis.yml
```diff
@@ -29,11 +29,12 @@ before_install:
   - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters https://cachix.cachix.org --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
   - cachix use dapp
   - git clone --recursive https://github.com/dapphub/dapptools $HOME/.dapp/dapptools
-  - nix-env -f $HOME/.dapp/dapptools -iA dapp solc
+  - nix-env -f $HOME/.dapp/dapptools -iA dapp seth solc
 
 install:
   - mkdir -p ~/bin
   - make install
+  - wget -P $HOME/.cargo/bin/ https://storage.googleapis.com/centrifuge-dev-public/subkey  && chmod +x $HOME/.cargo/bin/subkey
```
vedhavyas: nice 👍

anchors/service.go
```diff
@@ -2,132 +2,64 @@ package anchors
...
-	log.Infof("Add Anchor to Pre-commit %s from did:%s", anchorID.String(), did.ToAddress().String())
-	_, done, err := s.jobsMan.ExecuteWithinJob(contextutil.Copy(ctx), did, jobID, "Check Job for anchor commit",
-		s.ethereumTX(opts, s.anchorRepositoryContract.PreCommit, pc.AnchorID.BigInt(), pc.SigningRoot))
-	if err != nil {
-		return nil, err
+		return docRoot, anchoredTime, err
 	}
 
-	return done, nil
+	return dr, time.Unix(0, 0), nil
```
vedhavyas: are we not returning the actual anchor time here ?

centchain/api.go
```diff
@@ -1,74 +1,163 @@
...
 
-	key, err := types.CreateStorageKey(meta, "System", "AccountNonce", krp.PublicKey)
+	rv, err := a.sapi.GetRuntimeVersionLatest()
 	if err != nil {
 		return txHash, bn, sig, err
 	}
 
-	var nonce uint32
-	err = a.getStorageLatest(key, &nonce)
+	nonce, err := contextutil.Nonce(ctx)
```
vedhavyas: wouldn;t it be better to fetch the latest nonce here instead of fetching from CTX?

centchain/api.go
```diff
@@ -97,3 +186,160 @@ func (a api) SubmitExtrinsic(meta *types.Metadata, c types.Call, krp signature.K
...
+}
+
+func getSignature(msig types.MultiSignature) types.Signature {
+	if msig.IsEd25519 {
+		return msig.AsEd25519
+	}
+	if msig.IsSr25519 {
+		return msig.AsSr25519
+	}
+	return types.Signature{}
```
vedhavyas: maybe an error here ?

centchain/api.go
vedhavyas: or maybe a explicit argument to function maybe?

anchors/anchor_contract.go
```diff
@@ -172,7 +172,7 @@ func (_AnchorContract *AnchorContractTransactorRaw) Transact(opts *bind.Transact
 	return _AnchorContract.Contract.contract.Transact(opts, method, params...)
 }
 
-// GetAnchorById is a free data retrieval call binding the contract method 0x32bf361b.
```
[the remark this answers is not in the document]
mikiquantum: I guess I accidentally renamed it, but since it conforms to standards then, I believe is fine.

anchors/anchor_repository.go
```diff
@@ -21,7 +21,4 @@ type AnchorRepository interface {
 
 	// GetAnchorData takes an anchorID and returns the corresponding documentRoot from the chain.
 	GetAnchorData(anchorID AnchorID) (docRoot DocumentRoot, anchoredTime time.Time, err error)
-
-	// HasValidPreCommit checks if the given anchorID has a valid pre-commit
-	HasValidPreCommit(anchorID AnchorID) bool
```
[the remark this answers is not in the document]
mikiquantum: Nope, we never used it, only in tests

anchors/service.go
mikiquantum: As I mentioned above in the comment, it is a todo item. We need a way to get the block time out of the block events? I didnt want to block this PR just for this.

centchain/api.go
mikiquantum: I wanted to have the submitExtrinsic as simple as possible and delegate the nonce to the submit with retrial logic, since it is where the latest nonce per account is stored, since now we keep it in memory.

anchors/chain_repository.go
mikiquantum: Seems that it needs a pointer of course and looks like the currently defined seems redundant, but it doesnt work unless I pass with `&` I made a change to only define it by value and then pass by reference in my next commit. 
Thanks for catching it

centchain/api.go
mikiquantum: I think the CTX is an ok place to put data that is passed downstream.

vedhavyas: passing nonce explicitly will avoid unneeded type conversions with context and also error checking. At a later time, If we continue to pass nonce down the stream through context, nonce might not be the latest and we may have to deal with txn failure issues. 

For now, this is okay but we need to be careful on this IMO

PR #1324, opened 2019-12-13
Validate proofs against centchain

### LINKED ISSUE #1320: Validate proofs from centrifuge chain
Send extrinsic and wait for confirmation

vedhavyas:
Closes #1320

### DIFF
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -441,7 +441,7 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
 
 	return &DocumentProof{
 		FieldProofs:    rawProofs,
-		DataRoot:       targetTree.RootHash(),
+		BasicDataRoot:  targetTree.RootHash(),
 		SiblingRoot:    siblingTree.RootHash(),
 		SigningRoot:    sdr,
 		SignaturesRoot: signatureTree.RootHash(),
--- a/documents/coredocument.go
+++ b/documents/coredocument.go
@@ -423,10 +423,8 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
 	}
 
 	targetTree := basicDataTree
-	siblingTree := zkDataTree
 	if fromZKTree {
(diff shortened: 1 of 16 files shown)


documents/coredocument.go
```diff
@@ -441,6 +441,7 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
 
 	return &DocumentProof{
 		FieldProofs:    rawProofs,
+		DataRoot:       targetTree.RootHash(),
```
mikiquantum: Since the inner data trees are static, maybe we can rename this as well to:
- `DataRoot` to `BasicDataRoot` or `LeftDataRoot`
- `SiblingRoot` to `ZkDataRoot` or `RightDataRoot`
So any consumer of this can ultimately know what to do with this response.

nft/chain_api.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	c, err := types.NewCall(
+		meta,
+		ValidateMint,
+		types.NewHash(anchorID[:]),
+		depositAddress,
+		proofs,
```
mikiquantum: I am very curious on how the scale encoder/decoder can know how to encode the struct SubstrateProof. Do you know how since there is no enc/dec methods associated to the struct?

vedhavyas: Can explain more on this over a call :-D

documents/coredocument.go
vedhavyas: I would like to keep this as is because, target tree can be zk tree or basic treee. It would bee confusing to put either because we might give out wrong naming when the flag for zk is turned on. WDYT?

mikiquantum: The problem I see is that since `targetTree` and `siblingTree` will be swaped (in content) then when enabling zk flag, NFT logic will fail since it expects static_hashes={basicroot,zkroot,signaturesroot}, then the centchain nft module will consume the wrong order for dataRoot and siblingRoot.

PR #1325, opened 2019-12-18
Feat/nft mint

### LINKED ISSUE #1321: Send eth transaction to asset contract
Send eth txn to asset contract with asset hash and wait for asset deposited event

### LINKED ISSUE #1322: Make mint NFT txn to eth
Once asset deposited is received, make Mint NFT txn to eth

vedhavyas:
Closes #1321 Closes #1322


httpapi/userapi/types.go
```diff
@@ -351,9 +351,10 @@ func toFundingAgreementListResponse(ctx context.Context,
 
 // MintNFTRequest holds required fields for minting NFT
 type MintNFTRequest struct {
-	DocumentID     byteutils.HexBytes `json:"document_id" swaggertype:"primitive,string"`
-	DepositAddress common.Address     `json:"deposit_address" swaggertype:"primitive,string"`
-	ProofFields    []string           `json:"proof_fields"`
+	DocumentID          byteutils.HexBytes    `json:"document_id" swaggertype:"primitive,string"`
+	DepositAddress      common.Address        `json:"deposit_address" swaggertype:"primitive,string"`
+	AssetManagerAddress byteutils.OptionalHex `json:"asset_manager_address" swaggertype:"primitive,string"`
```
mikiquantum: We can even read the property `assetManager ` from the NFT Registry passed (if it is public), to avoid maintaining it in the code. WDYT?

nft/service.go
```diff
@@ -265,14 +268,26 @@ func (s *service) minterJob(ctx context.Context, tokenID TokenID, model document
 				return
 			}
 			log.Infof("Successfully validated Proofs on cent chain for anchorID: %s", requestData.AnchorID.String())
-			// TODO: send txn to asset manager and wait for success
-			// to common.Address, tokenId *big.Int, bundleHash [32]byte, properties [][]byte, values [][]byte, salts [][32]byte, proofs [][][32]byte
-			//args = []interface{}{requestData.To, requestData.TokenID, requestData.Props, requestData.Values, requestData.Salts}
-			//mintContractABI = GenericMintMethodABI
+			if !utils.IsEmptyAddress(req.AssetManagerAddress) {
```
mikiquantum: Maybe a comment to explain that it will be changed to listen to bridge event as soon as the bridge is operational.

nft/service.go
```diff
@@ -265,14 +268,26 @@ func (s *service) minterJob(ctx context.Context, tokenID TokenID, model document
...
+					return
+				}
+
+				log.Infof("Asset successfully deposited with TX hash: %v\n", txHash.String())
+			}
 
-			// TODO: remove the return once we have the generic NFT Mint function is working
-			errOut <- nil
-			return
+			// to common.Address, tokenId *big.Int, bundleHash [32]byte, properties [][]byte, values [][]byte, salts [][32]byte, proofs [][][32]byte
```
mikiquantum:
```suggestion
			// to common.Address, tokenId *big.Int, properties [][]byte, values [][]byte, salts [][32]byte
```

httpapi/userapi/types.go
vedhavyas: This piece of code will be removed as soon as the bridge integration is done. Then we will pull the asset manager address as part of default config. I have added this here since it easier to delegate the address to API caller, which will be functional tests. I would rather have us wait until bridge is in

PR #1336, opened 2020-01-14
Remove invoice

### LINKED ISSUE #1335: Clean up document specific logic

vedhavyas:
Closes #1335


bootstrap/bootstrapper.go
```diff
@@ -10,6 +10,7 @@ const (
 	BootstrappedQueueServer string = "BootstrappedQueueServer"
 	NodeObjRegistry         string = "NodeObjRegistry"
 	// BootstrappedInvoiceUnpaid is the key to InvoiceUnpaid NFT in bootstrap context.
+	// TODO(ved): remove this ?
```
mikiquantum: Since we already removed the invoice create/update APIs, and we need to update anyways gateway code, I would remove it as well.

vedhavyas: This is still being used and I don't know why we did that. We need to remove this during another cleanup IMO

PR #1337, opened 2020-01-17
Add bridge as test dep

mikiquantum:
Issue Link: <Github link>


build/scripts/test-dependencies/test-xbridge/config/config.toml
```diff
@@ -0,0 +1,15 @@
...
+endpoint = "ws://geth:9546"
+emitter = "0x1fA38b0EfccA4228EB9e15112D4d98B0CEe3c600"
+receiver = "0x290f41e61374c715C1127974bf08a3993afd0145"
+from = "021a7faa177621e1a6bb2e2794d77bcd1040221bcf5f0b83180f368eaab7848551"
+
+
+[[chains]]
+id = 1
+endpoint = "ws://cc:9944"
+emitter = "0x1fA38b0EfccA4228EB9e15112D4d98B0CEe3c600"
```
vedhavyas: are these deployed already?
Which testnet do they belong to ?

mikiquantum: No, these will change when ur run the test suite locally  or in travis, same as our other tests against local dev ethereum. We need to override them after the emitter/receiver contracts are migrated as part of a previous step.

PR #1338, opened 2020-01-20
Bridge integration

### LINKED ISSUE #1329: Integrate new bridge handling in go-centrifuge
NFT Bridge in the simplest (trusted) setup up and running to test the bridge module

#### Testing

#### Acceptance Criteria

### LINKED ISSUE #1331: Ethereum Contracts: NFT integration with bridge
integrate new bridge contracts with new NFT minting process.

move to ethereum contracts repo.

vedhavyas:
Closes #1329 
Closes #1331


ethereum/wait_for_event_task.go
```diff
@@ -0,0 +1,192 @@
...
+		},
+	}
+
+	return nil
+}
+
+// RunTask runs the task of fetching the logs.
+func (t *WaitForEventTask) RunTask() (res interface{}, err error) {
+	var jobValue *jobs.JobValue
+	ctx, cancelFunc := t.ethContextInitializer(5 * time.Second)
```
mikiquantum: Can we use `contextReadWaitTimeout` value from config instead or if it is still to high, create another constant with the appropriate value.

nft/service.go
```diff
@@ -257,6 +260,12 @@ func (s *service) minterJob(ctx context.Context, tokenID TokenID, model document
 		if req.UseGeneric { // TODO Remove once we have finalized the generic NFT work
```
mikiquantum: We can start thinking on dropping this branch, and default always to generic. I think we can do it after we are done with Amber release.

nft/service.go
```diff
@@ -269,17 +278,19 @@ func (s *service) minterJob(ctx context.Context, tokenID TokenID, model document
 			}
 			log.Infof("Successfully validated Proofs on cent chain for anchorID: %s", requestData.AnchorID.String())
 
-			// TODO(ved): remove as soon as bridge is integrated
 			if !utils.IsEmptyAddress(req.AssetManagerAddress) {
-				txHash, done, err := s.identityService.Execute(ctx, req.AssetManagerAddress, AssetManagerABI, "store", requestData.BundledHash)
+				// listen for event
```
mikiquantum: Nice

testingutils/setup.go
```diff
@@ -255,9 +233,6 @@ func BuildIntegrationTestingContext() map[string]interface{} {
 	StartPOAGeth()
 	StartCentChain()
 	StartBridge()
-	AddBalanceToBridgeAccount()
```
mikiquantum: cool, so since the startbridge script already bundles the add_balance, we do not need to add it here. What about the `RunSmartContractMigrations()`, when we only execute a single integration test? It will not be executed unless goes through the test_wrapper, right?

vedhavyas: nah, it will still go through if the bridge is not running

nft/service.go
vedhavyas: yeah

PR #1339, opened 2020-01-27
New envs + cleanup NFT

mikiquantum:
Issue Link: <Github link>


httpapi/coreapi/types.go
```diff
@@ -321,11 +319,12 @@ func toNFTMintRequest(req MintNFTRequest, registryAddress common.Address) nft.Mi
...
 		DepositAddress:           req.DepositAddress,
-		GrantNFTReadAccess:       req.GrantNFTReadAccess,
+		GrantNFTReadAccess:       false,
 		ProofFields:              req.ProofFields,
 		RegistryAddress:          registryAddress,
-		SubmitNFTReadAccessProof: req.SubmitNFTReadAccessProof,
-		SubmitTokenProof:         req.SubmitTokenProof,
+		AssetManagerAddress:      common.HexToAddress(req.AssetManagerAddress.String()),
+		SubmitNFTReadAccessProof: false,
+		SubmitTokenProof:         true,
```
vedhavyas: We can remove this as well IMO

mikiquantum: @vedhavyas I can do that. I will default to these values in the service code itself, and delete from here the flags. Is that what you meant?

vedhavyas: resolved

PR #1342, opened 2020-02-03
Run test world against any network

### LINKED ISSUE #1334: e2e integration with go-centrifuge testworld on flint and prepare for amber
Run tests even when cent chain is merged to develop

vedhavyas:
Closes #1334 

This will currently execute testworld on dev mode for every PR but will execute against fulvous on every merge to develop.

We could then add make the testworld run against flint on merge to master but adding envs to the TRAVIS.

Once we have proper config management in place, we can pull the configs from there instead. Right now, the envs in Travis are encrypted and are available to only develop and master.


testworld/configs/testing.json
```diff
@@ -0,0 +1,13 @@
+{
+  "run_network": true,
+  "run_migrations": true,
+  "network" : "testing",
+  "eth_node_url": "ws://127.0.0.1:9546",
+  "tx_pool_access": true,
```
mikiquantum: we can remove the `tx_pool_access` usage safely now

vedhavyas:
> If you can add an ignore rule for the DS_store coming from testworld
It was being tracked in the test world. Hence, I removed it. Are you proposing to add rule to .gitignore ?

> I noticed this in the env vars in travis TESTWORLD_NETWORK=fulvous as the var name, that might be malformed?
This is expected.

>The TRAVIS env vars are not injected in the build for PR stage right? Only for fulvous (dev) and flint (master) as you described.
Yes. They will not be injected in the PRs

PR #1353, opened 2020-03-18
2- Transition rules: Get API

### LINKED ISSUE #1349: Get Rule API

#### Testing

#### Acceptance Criteria

vedhavyas:
Closes #1349 

Review after #1352

### DIFF
--- a/build/scripts/docker/docker-compose-cc.yml
+++ b/build/scripts/docker/docker-compose-cc.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   cc:
-    image: "centrifugeio/centrifuge-chain:latest"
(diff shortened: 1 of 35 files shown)


vedhavyas:
Transition ruleID will be returned as part of the CreateRule in the response

mikiquantum:
@vedhavyas I believe a user will need to know the ruleIDs and it is unrealistic that they will store that on the client side so they know how to query them later on, so I believe we need to provide that response in either of the ways I refer in my previous comment. 
We can do this either in this PR or a different one.

vedhavyas:
Sure!
Another PR then :) I don't want to block this tinlake oracles for that

PR #1356, opened 2020-04-06
update to latest GSRPC

mikiquantum:
Issue Link: <Github link>


centchain/api.go
```diff
@@ -92,7 +92,8 @@ func (dsa *defaultSubstrateAPI) GetBlock(blockHash types.Hash) (*types.SignedBlo
 }
 
 func (dsa *defaultSubstrateAPI) GetStorage(key types.StorageKey, target interface{}, blockHash types.Hash) error {
-	return dsa.sapi.RPC.State.GetStorage(key, target, blockHash)
+	_, err := dsa.sapi.RPC.State.GetStorage(key, target, blockHash)
```
mikiquantum, line 95: @philipstanislaus  Since the new API returns bool, differentiating between empty or not, I am not sure if the downstream impact of subsequent calls if empty. Any pointers?

PR #1370, opened 2020-04-23
Add nft ownerof retrial

mikiquantum:


vedhavyas:
@mikiquantum why do we need the retrial here?

mikiquantum:
@vedhavyas every read operation against a fullnode (with load balancer behind) should have a retrial mechanism, since we do write operations and immediate read operations over data before, and if the request lands in a node that is not synced yet the whole initial action will mistakenly fail.
In this case, NFT is minted and immediately checks for owner and it fails bc the second read request lands in a node that is not synced yet.

PR #1381, opened 2020-05-14
update gocelery

vedhavyas:
All the celery tasks will use the default retrial mechanism defined in gocelerey with max backoff period between each retry on exponential basis leads to 5 Mins


mikiquantum:
@vedhavyas I think there are a couple other places that we need to adapt as well. Since waiting for the job to finish has a timeout as well driven by this config setting `ethereum.contextWaitTimeout`
We either remove this setting and default to the max go celery one or adapt it to match the max of go celery. I would do the former since any change in the config to be a lower value than the max can cause leaving open threads from the task side.

mikiquantum:
Like: https://github.com/centrifuge/go-centrifuge/blob/develop/ethereum/wait_for_event_task.go#L183

vedhavyas:
@mikiquantum ideally we shouldn't rely on ethcontextwaittime since task can be of any context. We already have a config param `MaxRetries`. I have repurposed it to hold a duration, which is used as both timeout and valid duration for a given task and renamed it accordingly. Thanks for pointing this out :-D

PR #1387, opened 2020-06-11
Go 14 and Modules

vedhavyas:


p2p/server.go
```diff
@@ -200,24 +183,29 @@ func makeBasicHost(ctx context.Context, priv crypto.PrivKey, pub crypto.PubKey,
...
 	opts := []libp2p.Option{
-		libp2p.Peerstore(ps),
-		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", listenPort)),
 		libp2p.Identity(priv),
+		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", listenPort)),
+		libp2p.Security(secio.ID, secio.New),
+		// support any other default transports (TCP)
+		libp2p.Transport(tcp.NewTCPTransport),
+		// Attempt to open ports using uPNP for NATed hosts.
+		libp2p.NATPortMap(),
```
manuelpolzhofer, line 194: I think it is good to enable the NATPortMap by default but we might should give the option to also deactivate it with a config variable. (If you know uPNP is deactivated on your router or the node is running on a public ip address)

p2p/server.go
```diff
@@ -200,24 +183,29 @@ func makeBasicHost(ctx context.Context, priv crypto.PrivKey, pub crypto.PubKey,
...
+		// Let this host use the DHT to find other hosts
+		libp2p.Routing(func(h host.Host) (routing.PeerRouting, error) {
+			idht, err = dht.New(ctx, h, dht.Mode(dht.ModeAutoServer))
```
manuelpolzhofer, line 197: 👍 
it's nicer to initialise the DHT with a closure

p2p/server.go
```diff
@@ -132,15 +132,24 @@ func (s *peer) InitProtocolForDID(DID *identity.DID) {
...
-		if err := s.host.Connect(ctx, *pinfo); err != nil {
+		multiaddr, _ := ma.NewMultiaddr(addr)
+		p, err := libp2pPeer.AddrInfoFromP2pAddr(multiaddr)
+		if err != nil {
+			log.Info(err)
+			continue
+		}
+
+		s.host.Peerstore().AddAddrs(p.ID, p.Addrs, peerstore.PermanentAddrTTL)
+		err = s.host.Connect(ctx, *p)
```
manuelpolzhofer, line 146: I think we could make all `Connect` calls in pararell with go-routines 
Similar to : https://github.com/ipfs/go-ipfs/blob/master/docs/examples/go-ipfs-as-a-library/main.go#L132

For improving performance.

Currently it doesn't matter that much, because every node uses our pre-defined bootstrap nodes.

p2p/server.go
manuelpolzhofer: It maybe would be nice to make the libp2p changes in a seperated PR?

p2p/server.go
mikiquantum: In either case it waits until all the go routines have filled the waitgroup. Internally already the routing table is refreshed every 5 mins using their own go routine.
I rather take the initial hit so we know that when the node is up, the routing table is filled with as much as relevant peers as possible.

documents/coredocument.go
```diff
@@ -19,8 +19,8 @@ import (
 	"github.com/centrifuge/precise-proofs/proofs/proto"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	"github.com/ethereum/go-ethereum/crypto/sha3"
 	"github.com/golang/protobuf/ptypes/any"
+	"golang.org/x/crypto/sha3"
```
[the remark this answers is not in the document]
vedhavyas: they switched to the same one

crypto/generate_test.go
```diff
@@ -19,6 +20,8 @@ const (
 	PrivateKeyED25519Len   = 64
 )
 
+var log = logging.Logger("keytools")
```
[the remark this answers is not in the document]
vedhavyas: moved it here since we werent using it. I think we should remove this if we are not using this logger

p2p/server.go
vedhavyas: We would need the node to be communicated by the external peers and we try to do that through config by explicitly mentioning an external. Having a config to disable this would defeat the purpose IMO. 

Moving libp2p changes to a different PR is more work than it is now since we  have to specifically downgrade some libs and hard to determine. Feels not worth that time IMO

manuelpolzhofer: Yes, it is in our interest to achieve communication with other peers if we are behind a NAT and uPNP would enable the port forwarding on the router with uPNP automatically. If I understood the `NATPortMap` option correctly.  

However, I thought if you know uPNP is anyway deactivated on you router it doesn't help you.
In IPFS you can deactivate the `NATPortMap` but it is enabled by default but i guess it is also fine to always enable it. 👍 
> Moving libp2p changes to a different PR is more work than it is now since we have to specifically downgrade some libs and hard to determine. Feels not worth that time IMO

makes sense 👍

p2p/server.go
manuelpolzhofer: Yes, but i think establishing the connection can take some time, especially if bootstrap peers are not reachable and it is a timeout. 

I think if you do it in parallel with a waitinggroup the max time is equal to the peer with the longest connection establishment/timout instead of the sum of all connection establishments.

But in our case it doesn't matter yet, because we don't provide 10 bootstrap peers with 4 peers maybe not reachable in it.

p2p/server.go
vedhavyas: > In IPFS you can deactivate the NATPortMap but it is enabled by default but i guess it is also fine to always enable it.
if lib unable to punch a hole through NAT, we have to ensure we pass the external IP through config, which we do that anyway. So leaving it enabled would be good in case the config has a missing external IP and NAT port mapping is supported

PR #1391, opened 2020-06-20
Add dev dependencies

vedhavyas:
`go.mod` is all we require to produce reproducible builds. By default `go build` will update mod file if there are any missing deps for any OS, architecture.

when we do `go get -u` this will update the dependency and its transitive dependency as well. Since we did `go get -u` earlier, this was an indicator for modules to update its transitive dependencies and there was a recent commit to the deps you have mentioned. Hence, the go.mod file changed.

always prefer to use just `go get` so that we don't implicitly update its transitive dependencies. More can be read here - https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies

`go.sum`:
go.sum is not a lock file. It contains cryptographic hashes of the dependencies and can contain deps that we no longer use as well. This is just another protection to cross-check the deps hash with the one in your local mod cache. This should be committed as well. We can verify that using `go mod verify`

This PR also locks the versions of the tools we need but will not be included in the final build since it is behind a build flag


vedhavyas:
@mikiquantum merging this. Let me know if there are any concerns here :-D

PR #1404, opened 2020-06-30
fix int256

### LINKED ISSUE #1394: Fix Int256. The encoding for negative values seems to be wrong.

vedhavyas:
Closes #1394 

This PR ensures we encode the Int256 negative as 2's complement

### DIFF
--- a/documents/int256.go
+++ b/documents/int256.go
@@ -1,12 +1,17 @@
 package documents
 
 import (
+	"fmt"
 	"math/big"
+	"regexp"
 	"strings"
 
 	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/ethereum/go-ethereum/common/math"
 )
+
+// validInt regex for only allowing integers
+var validInt = regexp.MustCompile(`^[-+]?\d+$`)
 
 // Int256 represents a signed 256 bit integer
 type Int256 struct {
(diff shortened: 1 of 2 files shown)


documents/int256.go
```diff
@@ -62,46 +49,24 @@ func NewInt256(n string) (*Int256, error) {
 	return &Int256{*nn}, nil
 }
 
-// Int256FromBytes converts the a big endian byte slice to an Int256
+// Int256FromBytes converts the a big endian 2's compliment byte slice to an Int256
```
mikiquantum, line 52:
```suggestion
// Int256FromBytes converts the a big endian 2's complement byte slice to an Int256
```

documents/int256.go
```diff
@@ -62,46 +49,24 @@ func NewInt256(n string) (*Int256, error) {
...
-		nn = nn.Neg(nn)
-	}
-
+	nn := math.S256(new(big.Int).SetBytes(b))
 	return &Int256{*nn}, nil
 }
 
-// Bytes returns the big endian 32 byte representation of this int256. First bit of LSB(least significant byte) is used as the sign bit.
-func (i *Int256) Bytes() [32]byte {
+// Bytes returns the big endian 2's compliment 32 byte representation of this int256.
```
mikiquantum, line 62:
```suggestion
// Bytes returns the big endian 2's complement 32 byte representation of this int256.
```

documents/int256.go
```diff
@@ -62,46 +49,24 @@ func NewInt256(n string) (*Int256, error) {
...
+func (i Int256) Bytes() [32]byte {
+	v := new(big.Int).Mul(&i.v, big.NewInt(1))
```
mikiquantum, line 64: Why do we multiply by 1, is it only for rounding?

vedhavyas: U256 is a destructive operation. We need to clone the value so that it is usable after calling bytes. Hence we clone it using a multiplier and shouldn't affect any rounding since we don't support any decimal here

vedhavyas:
> Looks good, I am curious to know, how did you find out that we were not begin consistent with the two complement representation.

https://github.com/centrifuge/go-centrifuge/pull/1404/files#diff-f5b9c3a1a3091d05221c4762d41890f1L72
this removed code doesn't do any 2 compliment complement but instead records the sign at MSB and ignores the remaining bit completely and encodes them as + integer. Noticed this while testing for WASM

PR #1405, opened 2020-07-01
add signed subtype

### LINKED ISSUE #1395: Add subtype to signed fields

vedhavyas:
Closes #1395 

This PR adds subtype to signed field instead of assuming it be bytes


documents/converters.go
```diff
@@ -364,3 +360,17 @@ func attrValFromProtocolAttribute(attrType AttributeType, attribute *coredocumen
 
 	return attrVal, err
 }
+
+func timestampToBytes(tm *timestamp.Timestamp) ([]byte, error) {
```
charleenfei, line 364: maybe this could go in the byteutils folder

vedhavyas: Agree!

Issue #1204, opened 2019-07-18
createconfig should not require http or https prefix


manuelpolzhofer:
###  Good First Issue
A good first issue for new developers.  Contributions are always very welcome.

##  Problem
The `centrifuge createconfig` command requires for the  parameter `-e, --ethnodeurl string`  a URL with an `http` or `https` prefix.

More information: https://developer.centrifuge.io/docs/getting-started/install-cent-node/

### Current behaviour
Example: `-e rinkeby.infura.io/v3/<CUSTOM PART>` fails
Example: `-e https://rinkeby.infura.io/v3/<CUSTOM PART>`works
##

### Correct behaviour
Example: `-e rinkeby.infura.io/v3/<CUSTOM PART>`  works
Example: `-e https://rinkeby.infura.io/v3/<CUSTOM PART>` works

### Solution
Check if passed url contains `http` or `https` prefix if not add by default `https`

###  Why?
Currently, Infura provides a default project link without an `http` or `https` prefix.

Copy/paste from Infura would result in an error which is bad onboarding experience for new users.

lucasvo:
We should also verify it is a correct url, if a prefix is defined, i.e. you can't specify `ftp://....`

PR #1407, opened 2020-07-02
Validate configuration ethereum.nodeURL

arberiii:
Hello 👋 , this is my first PR in go-centrifuge project. I picked [issue #1204](https://github.com/centrifuge/go-centrifuge/issues/1204) which is labeled as the good first issue and tried to work it out.

As described in the issue, etherNodeURL needs a prefix and by default should be set to https. The validateURL adds the prefix if necessary and checks if the schema is either http or https.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -17,7 +17,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/centrifuge/centrifuge-protobufs/gen/go/coredocument"
 	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-centrifuge/resources"
 	"github.com/centrifuge/go-centrifuge/storage"
@@ -30,6 +29,12 @@ import (
 )
 
 var log = logging.Logger("config")
+
+var allowedURLScheme = map[string]struct{}{
+	"http":  {},
+	"https": {},
+	"ws":    {},
+}
 
 // AccountHeaderKey is used as key for the account identity in the context.ContextWithValue.
 var AccountHeaderKey struct{}
@@ -695,7 +700,9 @@ func validateURL(u string) string {
 
 	if parsedURL.Scheme == "" {
 		parsedURL.Scheme = defaultURLScheme
-	} else if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+	}
+
+	if _, ok := allowedURLScheme[parsedURL.Scheme]; !ok {
 		log.Fatalf("error: url scheme %s is not allowed", parsedURL.Scheme)
 	}
 
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/big"
+	"net/url"
 	"os"
 	"reflect"
 	"strings"
@@ -40,6 +41,8 @@ type ContractName string
 type ContractOp string
 
 const (
+	defaultURLPrefix = "https"
+
 	// AnchorRepo is the contract name for AnchorRepo
 	AnchorRepo ContractName = "anchorRepository"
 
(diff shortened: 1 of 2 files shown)


config/configuration.go
```diff
@@ -651,7 +660,7 @@ func CreateConfigFile(args map[string]interface{}) (*viper.Viper, error) {
 	if p2pConnectTimeout != "" {
 		v.Set("p2p.connectTimeout", p2pConnectTimeout)
 	}
-	v.Set("ethereum.nodeURL", ethNodeURL)
+	v.Set("ethereum.nodeURL", validateURL(ethNodeURL))
 	v.Set("ethereum.accounts.main.key", "")
 	v.Set("ethereum.accounts.main.password", "")
 	v.Set("centChain.nodeURL", centChainURL)
```
vedhavyas, line 666: We would need to do the same check for the centrifugeChain URL as well

config/configuration.go
```diff
@@ -683,3 +692,20 @@ func CreateConfigFile(args map[string]interface{}) (*viper.Viper, error) {
 func (c *configuration) SetupSmartContractAddresses(network string, smartContractAddresses *SmartContractAddresses) {
 	c.v.Set("networks."+network+".contractAddresses.identityFactory", smartContractAddresses.IdentityFactoryAddr)
 }
+
+func validateURL(u string) string {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		log.Fatalf("error: %v", err)
```
vedhavyas, line 699: return an error and let the caller handle the error

config/configuration.go
```diff
@@ -683,3 +692,20 @@ func CreateConfigFile(args map[string]interface{}) (*viper.Viper, error) {
...
+	}
+
+	if parsedURL.Scheme == "" {
+		parsedURL.Scheme = defaultURLScheme
+	}
+
+	if _, ok := allowedURLScheme[parsedURL.Scheme]; !ok {
+		log.Fatalf("error: url scheme %s is not allowed", parsedURL.Scheme)
```
vedhavyas, line 707: same as above :)

config/configuration.go
```diff
@@ -30,6 +31,12 @@ import (
 
 var log = logging.Logger("config")
 
+var allowedURLScheme = map[string]struct{}{
+	"http":  {},
+	"https": {},
+	"ws":    {},
```
vedhavyas, line 37: we would need to include `wss` as well

arberiii:
> Thank you for your contribution!!
> 
> The approach LGTM but we are missing this check while loading the config on every run. I think we could move this check into the config package and validate the checks there.

@vedhavyas Thank you for the review! I didn't understand where to move the validation as for now it is placed at go-centrifuge/config package

config/configuration.go
```diff
@@ -683,3 +699,20 @@ func CreateConfigFile(args map[string]interface{}) (*viper.Viper, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	if parsedURL.Scheme == "" {
+		parsedURL.Scheme = defaultURLScheme
+	}
+
+	if _, ok := allowedURLScheme[parsedURL.Scheme]; !ok {
+		return "", errors.New("error: url scheme %s is not allowed", parsedURL.Scheme)
```
vedhavyas, line 714:
```suggestion
		return "", errors.New("url scheme %s is not allowed", parsedURL.Scheme)
```

vedhavyas:
@arberiii I revert the previous validations as well since those validations happen when config file was created and `initializeViper` validations happens when created config is loaded. The reason we are doing this because once the config is created, I could also change the config file with `ftp://` instead. Let me know if this makes sense to you

arberiii:
> @arberiii I would revert the previous validations as well since those validations happen when config file was created and `initializeViper` validations happens when created config is loaded. The reason we are doing this because once the config is created, I could also change the config file with `ftp://` instead. Let me know if this makes sense to you

Yes it makes sense. You actually explained well the first time, but my bad :). I revert it and fixed the assertion which was failing the tests

arberiii:
I added 3 test cases with expected behaviors.

vedhavyas:
@arberiii seems like `configuration_test.go` is not formatted properly
https://travis-ci.com/github/centrifuge/go-centrifuge/jobs/364988442#L718

vedhavyas:
Closing this in favour of #1472

PR #1415, opened 2020-07-14
createFingerprint method in CoreDocument

charleenfei:


documents/coredocument.go
```diff
@@ -134,6 +136,11 @@ type CoreDocument struct {
 	Document coredocumentpb.CoreDocument
 }
 
+// TransitionRulesFingerprint is a wrapper for the TransitionRulesFingerprint Protobuf
+type TransitionRulesFingerprint struct {
```
mikiquantum, line 140: Do we need a wrapper struct here?

documents/coredocument.go
```diff
@@ -158,6 +165,16 @@ func newCoreDocument() (*CoreDocument, error) {
 	}, nil
 }
 
+func newFingerprint() *TransitionRulesFingerprint {
+	f := coredocumentpb.TransitionRulesFingerprint{
+		Roles:           nil,
+		TransitionRules: nil,
+	}
+	return &TransitionRulesFingerprint{
```
mikiquantum, line 173: Following on the comment above, couldnt we just return the coredoc. TransitionRulesFingerprint object?

documents/tree.go
```diff
@@ -105,6 +105,14 @@ func (cd *CoreDocument) DocumentSaltsFunc() func(compact []byte) ([]byte, error)
 	}
 }
 
+// FingerprintSaltsFunc returns an empty function
+func (cd *CoreDocument) FingerprintSaltsFunc() func(compact []byte) ([]byte, error) {
```
mikiquantum, line 109: I would call this `ZeroSaltsFunc()`

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
...
+	p, err := cd.generateFingerprintHash(f.Fingerprint)
+	if err != nil {
+		return nil, err
+	}
+	return *p, nil
+}
+
+// generateFingerprintHash takes an assembled fingerprint message and generates the root hash from this message.
+// the return value can be used to verify if transition rules or roles have changed across documents
+func (cd *CoreDocument) generateFingerprintHash(fingerprint coredocumentpb.TransitionRulesFingerprint) (*[]byte, error) {
```
mikiquantum, line 488: Can we return just the []byte, without pointer.

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
...
+	b2bHash, err := blake2b.New256(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	t, err := proofs.NewDocumentTree(proofs.TreeOptions{
+		CompactProperties: false,
```
mikiquantum, line 495: Can we keep the CompactProperties generation?

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
 	}, nil
 }
 
+//CreateFingerprint generates a fingerprint for a Core Document
+func (cd *CoreDocument) CreateFingerprint() ([]byte, error) {
```
mikiquantum, line 469:
```suggestion
func (cd *CoreDocument) CalculateTransitionRulesFingerprint() ([]byte, error) {
```

documents/coredocument.go
mikiquantum, line 488:
```suggestion
func (cd *CoreDocument) generateTransitionRulesFingerprintHash(fingerprint coredocumentpb.TransitionRulesFingerprint) (*[]byte, error) {
```

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
...
+	f := newFingerprint()
+	if len(cd.Document.Roles) == 0 {
```
mikiquantum, line 471: Looks like both Roles and TransitionRules have to be set to proceed. Maybe combine these two checks in one line?

documents/entityrelationship/model.go
```diff
@@ -337,3 +337,12 @@ func (e *EntityRelationship) revokeRelationship(old *EntityRelationship, grantee
 	e.CoreDocument = cd
 	return nil
 }
+
+// CreateFingerprint creates a fingerprint from the transition rules and roles of a document
+func (e *EntityRelationship) CreateFingerprint() ([]byte, error) {
```
vedhavyas, line 342: we do not need this function if Coredocument already implements it

documents/generic/model.go
```diff
@@ -311,3 +311,12 @@ func (g *Generic) DeriveFromUpdatePayload(_ context.Context, payload documents.U
 func (g *Generic) Scheme() string {
 	return Scheme
 }
+
+// CreateFingerprint creates a fingerprint from the transition rules and roles of a document
+func (g *Generic) CreateFingerprint() ([]byte, error) {
```
vedhavyas, line 316: same as above

documents/tree.go
vedhavyas: this should not be method on Coredocument but rather a function.

documents/tree.go
```diff
@@ -105,6 +105,14 @@ func (cd *CoreDocument) DocumentSaltsFunc() func(compact []byte) ([]byte, error)
...
+	return func(compact []byte) ([]byte, error) {
+		randbytes := make([]byte, 32)
+		return randbytes, nil
```
vedhavyas, line 112: this function is should always return the same salt preferably nil bytes else it wouldn't be different from the CodeDocument salts function

resources/data.go
```diff
@@ -104,17 +105,16 @@ func AssetDir(name string) ([]string, error) {
 }
 
 type _bintree_t struct {
-	Func func() ([]byte, error)
```
vedhavyas, line 107: this file shouldn't have changed unless we changed the config files. Did we make any changes to default_config.yaml ?

vedhavyas: if not, could you revert it back ?

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
 	}, nil
 }
 
+//CreateFingerprint generates a fingerprint for a Core Document
+func (cd *CoreDocument) CreateFingerprint() ([]byte, error) {
+	f := newFingerprint()
```
vedhavyas, line 470: we dont need a new function IMO since this is the only place we are creating it

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
...
+	if len(cd.Document.Roles) == 0 {
+		return []byte{}, nil
+	}
+	if len(cd.Document.TransitionRules) == 0 {
+		return []byte{}, nil
+	}
+	f.Fingerprint.Roles = cd.Document.Roles
+	f.Fingerprint.TransitionRules = cd.Document.TransitionRules
```
vedhavyas, line 478: Roles shouldn't be copied since these roles might contain Roles belonging to ReadRules as well. First we should copy the transition rules and loop though each rule and collect the roles. These collected roles will be copied to fingerprint.roles

documents/coredocument.go
```diff
@@ -448,6 +465,56 @@ func (cd *CoreDocument) createProofs(fromZKTree bool, docType string, dataLeaves
...
+	p, err := cd.generateFingerprintHash(f.Fingerprint)
```
vedhavyas, line 479: `generateFingerprintHash` doesn't need to be a method to core document but instead can be a pure function

documents/entity/model.go
```diff
@@ -494,3 +494,12 @@ func (e *Entity) Patch(payload documents.UpdatePayload) error {
 func (e *Entity) Scheme() string {
 	return Scheme
 }
+
+// CreateFingerprint creates a fingerprint from the transition rules and roles of a document
+func (e *Entity) CreateFingerprint() ([]byte, error) {
```
vedhavyas, line 499: we do not need this function if Coredocument already implements it

httpapi/coreapi/types.go
```diff
@@ -83,15 +83,16 @@ type NFT struct {
...
+	DocumentID  string             `json:"document_id"`
+	VersionID   string             `json:"version_id"`
+	Author      string             `json:"author"`
+	CreatedAt   string             `json:"created_at"`
+	ReadAccess  []identity.DID     `json:"read_access" swaggertype:"array,string"`
+	WriteAccess []identity.DID     `json:"write_access" swaggertype:"array,string"`
+	JobID       string             `json:"job_id,omitempty"`
+	NFTs        []NFT              `json:"nfts"`
+	Status      string             `json:"status,omitempty"`
+	Fingerprint byteutils.HexBytes `json:"fingerprint,omitempty"`
```
vedhavyas, line 95:
```suggestion
	Fingerprint byteutils.HexBytes `json:"fingerprint,omitempty" swaggertype:"primitive,string"`
```

resources/data.go
charleenfei: for some reason running `make format-go` changes this file. I can revert it back.

charleenfei: i think it's just formatting changes from `make format-go`

documents/tree.go
```diff
@@ -105,6 +105,13 @@ func (cd *CoreDocument) DocumentSaltsFunc() func(compact []byte) ([]byte, error)
 	}
 }
 
+// ZeroSaltsFunc returns an empty function
+func ZeroSaltsFunc() func(compact []byte) ([]byte, error) {
+	return func(compact []byte) ([]byte, error) {
+		return []byte{}, nil
```
vedhavyas, line 111:
```suggestion
		return nil, nil
```

PR #1420, opened 2020-07-28
feat: clone endpoint

charleenfei:


documents/coredocument.go
```diff
@@ -173,6 +173,21 @@ type AccessTokenParams struct {
 	Grantee, DocumentIdentifier string
 }
 
+// NewClonedDocument generates new blank core document with a document type specified by the prefix: po or invoice.
```
vedhavyas, line 176: we do not have po or invoice anymore 👍 
maybe `Generic` ?

documents/entity/model.go
```diff
@@ -424,6 +424,23 @@ func (e *Entity) DeriveFromCreatePayload(_ context.Context, payload documents.Cr
 	return nil
 }
 
+// DeriveFromClonePayload unpacks the invoice data from the Payload
```
vedhavyas, line 427:
```suggestion
// DeriveFromClonePayload unpacks the entity data from the Payload
```

documents/generic/model.go
```diff
@@ -273,6 +273,23 @@ func (g *Generic) DeriveFromCreatePayload(_ context.Context, payload documents.C
 	return nil
 }
 
+// DeriveFromClonePayload unpacks the invoice data from the Payload
```
vedhavyas, line 276:
```suggestion
// DeriveFromClonePayload unpacks the generic document data from the Payload
```

documents/service.go
```diff
@@ -405,10 +405,26 @@ func (s service) Derive(ctx context.Context, payload UpdatePayload) (Model, erro
 		}
 
 		payload.Collaborators.ReadWriteCollaborators = append(payload.Collaborators.ReadWriteCollaborators, did)
-		if err := doc.(Deriver).DeriveFromCreatePayload(ctx, payload.CreatePayload); err != nil {
-			return nil, errors.NewTypedError(ErrDocumentInvalid, err)
+		for _, a := range payload.CreatePayload.Attributes {
```
vedhavyas, line 408: not sure I understand why the attribute has `template` that holds the template ID

httpapi/v2/documents_api.go
```diff
@@ -82,6 +89,59 @@ func (h handler) CreateDocument(w http.ResponseWriter, r *http.Request) {
...
+// @tags Documents
+// @accept json
+// @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
+// @param body body v2.CloneDocumentRequest true "Document Clone request"
+// @produce json
+// @Failure 400 {object} httputils.HTTPError
+// @Failure 500 {object} httputils.HTTPError
+// @Failure 403 {object} httputils.HTTPError
+// @success 201 {object} coreapi.DocumentResponse
+// @router /v2/documents/clone [post]
```
vedhavyas, line 105: IMO the url should be of the following form
`v2/documents/{doc_id}/clone`
Indicates we are cloning `docID`
WDYT ?

httpapi/v2/converters.go
```diff
@@ -21,6 +21,20 @@ func toDocumentsPayload(req DocumentRequest, docID []byte) (payload documents.Up
 	return documents.UpdatePayload{CreatePayload: cp, DocumentID: docID}, nil
 }
 
+func toCloneDocumentsPayload(req DocumentRequest, docID []byte) (payload documents.UpdatePayload, err error) {
+	cp, err := coreapi.ToDocumentsCreatePayload(req)
+	if err != nil {
+		return payload, err
+	}
+	for _, a := range cp.Attributes {
+		if a.KeyLabel == "template" {
```
mikiquantum, line 30: Following on Ved's comment above, adding the template as an attribute will make the payload susceptible of a name collision. Since we have an specific DeriveFromClonePayload I would rather pass the separate template id as an argument.

httpapi/v2/documents_api.go
charleenfei: yes, I think it makes sense

documents/service.go
```diff
@@ -88,6 +88,9 @@ type Service interface {
 	// If not provided, it is a fresh document.
 	Derive(ctx context.Context, payload UpdatePayload) (Model, error)
 
+	// DeriveClone derives the Model from the Payload, taking the provided template ID as the clone base
+	DeriveClone(ctx context.Context, templateID []byte, scheme string) (Model, error)
```
vedhavyas, line 92: Do you want to pass the `ClonePayload` instead to keep the consistency?

PR #1421, opened 2020-07-29
add Execute WASM function on Document

### LINKED ISSUE #1399: Add required WASM functions to compute the value give a set of inputs

### LINKED ISSUE #1400: Add a new Validator to check the compute field

vedhavyas:
closes #1399 
closes #1400 

I have made a fork of https://github.com/vedhavyas/life/commit/e96a0a4f69e36b6fdb8e490b8646d7cf50573fff that fixes the long running WASM. I have reached out to the perlin team to get this merged. The fix follows the standard Go context that makes the callers code much simpler.


documents/compute_fields.go
```diff
@@ -2,11 +2,14 @@ package documents
...
 	"bytes"
+	"context"
+	"time"
 
+	coredocumentpb "github.com/centrifuge/centrifuge-protobufs/gen/go/coredocument"
 	"github.com/centrifuge/go-centrifuge/errors"
 	"github.com/centrifuge/go-substrate-rpc-client/scale"
 	logging "github.com/ipfs/go-log"
-	"github.com/perlin-network/life/exec"
+	"github.com/wasmerio/go-ext-wasm/wasmer"
```
vedhavyas, line 12: had to change to wasmer since perlin doesn't have execution control to stop the function execution

documents/compute_fields.go
```diff
@@ -20,39 +23,45 @@ const (
...
 // returns a 32byte value. If the WASM exits with an error, returns a zero 32byte value
-func executeWASM(wasm []byte, attributes []Attribute) (result [32]byte) {
+// execution is allowed to run for upto timeout. Once the timeout is reached, VM is stopped and returned a zero value.
+func executeWASM(wasm []byte, attributes []Attribute, timeout time.Duration) (result [32]byte) {
 	i, allocate, compute, err := fetchComputeFunctions(wasm)
 	if err != nil {
 		computeLog.Error(err)
 		return result
 	}
+	defer i.Close()
```
vedhavyas, line 64: This is missing in perlin/life

documents/compute_fields.go
```diff
@@ -20,39 +23,45 @@ const (
 
 	// ErrComputeFieldsComputeNotFound is a sentinel error when WASM doesn't expose 'compute' function
 	ErrComputeFieldsComputeNotFound = errors.Error("'compute' function not exported")
+
+	// computeFieldsTimeout is the max time we let the WASM computation to be run.
```
charleenfei, line 27: why 30 secs? out of curiousity

documents/write_acls.go
```diff
@@ -341,6 +376,62 @@ func (cd *CoreDocument) AddTransitionRuleForAttribute(roleID []byte, key AttrKey
...
+		ComputeTargetField: []byte(targetField),
+		ComputeCode:        wasm,
+	}
+	cd.Document.TransitionRules = append(cd.Document.TransitionRules, rule)
+	cd.Modified = true
+	return rule, nil
+}
+
+// GetComputeFieldsRules returns all the compute fields rules from the document.
+func (cd CoreDocument) GetComputeFieldsRules() []*coredocumentpb.TransitionRule {
```
charleenfei, line 416: Should this be cd *CoreDocument?

documents/validator.go
```diff
@@ -465,6 +466,37 @@ func transitionValidator(collaborator identity.DID) Validator {
...
+			// execute compute fields
+			targetAttr, err := executeComputeField(computeField, attributes, timeout)
+			if err != nil {
+				return err
+			}
+
+			// verify the targetAttr is same as the one already stored
+			attr := attributes[targetAttr.Key]
+			if !reflect.DeepEqual(attr, targetAttr) {
+				return errors.New("compute fields[%s] validation failed", hexutil.Encode(computeField.RuleKey))
```
charleenfei, line 492: are we keeping errors in the `error.go` file in this directory?

documents/compute_fields.go
vedhavyas: should be good enough time to run a Complex WASM function as well

documents/write_acls.go
vedhavyas: Its a Get, pointer should't be used here

documents/validator.go
vedhavyas: We can't do that since this is dynamic error

documents/compute_fields.go
mikiquantum: Do you know our p2p timeout value? as long as this value is smaller than that one we are good. Do you think we should have it configurable? or no need for now.

documents/compute_fields.go
```diff
@@ -143,3 +172,63 @@ func toComputeFieldsAttribute(attr Attribute) (cattr computeAttribute, err error
 
 	return cattr, err
 }
+
+// ExecuteComputeFields executes all the compute fields and updates the document with target attributes
+// each WASM is executed at a max of timeout duration.
+func (cd *CoreDocument) ExecuteComputeFields(timeout time.Duration) error {
+	computeFields := cd.GetComputeFieldsRules()
```
mikiquantum, line 179:
```suggestion
	computeFieldsRules := cd.GetComputeFieldsRules()
```

documents/compute_fields.go
```diff
@@ -143,3 +172,63 @@ func toComputeFieldsAttribute(attr Attribute) (cattr computeAttribute, err error
...
+		fa, ok := attributes[key]
+		if !ok {
+			continue
+		}
+
+		attrs = append(attrs, fa)
+	}
+
+	// execute WASM
+	r := executeWASM(rule.ComputeCode, attrs, timeout)
```
mikiquantum, line 217: What would be the impact down the line of the result being 0..0 value due to an error in the wasm execution? Meaning that since we cannot distinguish between a succcessful run with 0..0 return value and an error. 
That means that we will be pushing this check for validation at the p2p consensus, then right?
What if all the participants run the code and it fails in all of them and then we store 0..0 in the oracle?

documents/compute_fields_test.go
```diff
@@ -120,6 +121,15 @@ func Test_executeWASM(t *testing.T) {
 			attrs: getInvalidComputeFieldAttrs(t),
 		},
 
+		// exceeded timeout
```
mikiquantum, line 124: So technically the timeout logic will not work in some (linux?) machines? And that is the environment that the code will be deployed into. If calling "Close" on long running code causes a panic in the main thread, maybe makes sense to do not implement the timeout logic yet.

documents/write_acls.go
```diff
@@ -204,11 +206,44 @@ func (cd *CoreDocument) CollaboratorCanUpdate(ncd *CoreDocument, collaborator id
...
+	if err != nil {
+		return err
+	}
+
+	cf := filterComputeFieldAttributes(GetChangedFields(oldTree, newTree), computeFieldsAttributes)
 	rules := cd.TransitionRulesFor(collaborator)
 	return ValidateTransitions(rules, cf)
 }
 
+func filterComputeFieldAttributes(changedFields []ChangedField, computeFieldsAttributes []string) (result []ChangedField) {
```
mikiquantum, line 219:
```suggestion
func filterOutComputeFieldAttributes(changedFields []ChangedField, computeFieldsAttributes []string) (result []ChangedField) {
```

documents/compute_fields.go
vedhavyas: This should not be configurable IMO. since this is core consensus, if configured differently on different nodes, might lead to timeouts and consensus being failed

documents/compute_fields.go
vedhavyas: We want the default value to be zero in case the WASm is failed to be executed or timeout. Oracle fetching the value zero is okay as well since the value above zero is what we expect for NFT feeds

documents/compute_fields_test.go
vedhavyas: Timeout logic works on all machines if the WASM has a forever loop or complex execution that goes beyond the 30 second timeout.

On some machines, the cgo is failing while stopping the instance when `timeout` exceeded. On a normal execution there should be no issues on any machines

vedhavyas: NVM, I have reverted back to perlin with a fix. We do not need Wasmer anymore :)

PR #1425, opened 2020-08-11
fix: doc id param

charleenfei:


httpapi/v2/documents_api.go
```diff
@@ -91,6 +91,7 @@ func (h handler) CreateDocument(w http.ResponseWriter, r *http.Request) {
 // @accept json
 // @param authorization header string true "Hex encoded centrifuge ID of the account for the intended API action"
 // @param body body v2.CloneDocumentRequest true "Document Clone request"
+// @param doc_id path string true "Document Identifier"
```
mikiquantum, line 94: Can we use the same naming convention as the rest of the document id params? `document_id`

mikiquantum: Below will need to change to:
```
@router /v2/documents/{document_id}/clone [post]
```

mikiquantum: In fact `coreapi.DocumentIDParam` is `document_id` so not sure if it would work fine with `doc_id` as the path param.

PR #1433, opened 2020-08-19
fix: rm write access

### LINKED ISSUE #1429: remove default write access for document creator
we need to do this to enable the templates use case for oracles, otherwise asset originators will retain master write access on the document

charleenfei:
Closes #1429


documents/service.go
```diff
@@ -397,7 +397,7 @@ func (s service) UpdateModel(ctx context.Context, payload UpdatePayload) (Model,
 // Derive looks for specific document type service based in the schema and delegates the Derivation to that service.˜
 func (s service) Derive(ctx context.Context, payload UpdatePayload) (Model, error) {
 	if len(payload.DocumentID) == 0 {
-		did, err := contextutil.AccountDID(ctx)
+		_, err := contextutil.AccountDID(ctx)
```
vedhavyas, line 400: if we are ignoring this, then we don't need to get the DID from the context

vedhavyas (approved):
Some code removal can be done.
LGTM after that

PR #1448, opened 2020-12-07
add attributes api

vedhavyas:
This API adds a new API to add attributes to the pending document.


mikiquantum (commented):
I am assuming that a DELETE attributes api would be added as well? If so then looks good to me.

vedhavyas:
yes @mikiquantum

PR #1450, opened 2020-12-10
Runtime Upgrade 4 

mikiquantum:


build/scripts/docker/docker-compose-bridge.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   bridge:
-    image: "chainsafe/chainbridge:v1.1.0"
+    image: "centrifugeio/chainbridge:20201209231213-ea7507c"
```
mikiquantum, line 4: There is a PR I would like to have in from the chainbridge repo: https://github.com/ChainSafe/ChainBridge/pull/564
It will give the relayer to provide a better QoS at the expense of gas cost, worth it in our usecases.

PR #1460, opened 2021-01-26
Accounts v2

vedhavyas:
This PR migrates account generate v1 to v2 using v2 of gocelery. There are still chunks of code that needs to be removed but will do that in a different PR so that this PR has the important changes I want to be looked at.


config/configstore/bootstrapper.go
```diff
@@ -25,15 +27,44 @@ func (*Bootstrapper) Bootstrap(context map[string]interface{}) error {
...
+		idFactory: idFactory,
+		idService: idService,
+		protocolSetterFinder: func() ProtocolSetter {
+			return context[bootstrap.BootstrappedPeer].(ProtocolSetter)
+		},
+		dispatcher:  dispatcher,
+		idFactoryV2: idFactoryV2,
+	}
+
+	go dispatcher.RegisterRunner(generateIdentityRunnerName, generateIdentityRunner{
```
mikiquantum, line 63: Is `RegisterRunner` a fast-executing function? should this be non-blocking? Since our health-check at startup will not know about when is this step done, might confuse the readiness.

config/configstore/jobs.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/centrifuge/go-centrifuge/ethereum"
+	"github.com/centrifuge/go-centrifuge/identity"
+	"github.com/centrifuge/go-centrifuge/jobs/jobsv2"
+	"github.com/centrifuge/gocelery/v2"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	generateIdentityRunnerName = "GenerateIdentity"
+	taskSendTxn                = "Create identity on Chain"
```
mikiquantum, line 18: I would add `Id` to the constant names `taskSendIdTxn` and `taskWaitForIdTxn` or alike.

ethereum/geth_client.go
```diff
@@ -19,15 +21,21 @@ import (
...
 	logging "github.com/ipfs/go-log"
 )
 
-var log = logging.Logger("geth-client")
+var log = logging.Logger("ethereum")
 var gc Client
 var gcMu sync.RWMutex
 
+func init() {
+	gob.Register(common.Hash{})
```
mikiquantum, line 35: What is this for?

ethereum/wait_for_event_task_test.go
```diff
@@ -91,48 +82,48 @@ func checkKwargs(t *testing.T, task *WaitForEventTask, kwargs map[string]interfa
...
-	task.filterLogsFunc = func(ctx context.Context, query ethereum.FilterQuery) (logs []types.Log, err error) {
-		return []types.Log{
-			{},
-		}, nil
-	}
-	_, err = task.RunTask()
-	assert.NoError(t, err)
-	jm.AssertExpectations(t)
-}
+// func TestWaitForEventTask_RunTask(t *testing.T) {
```
mikiquantum, line 85: Is this unit test not relevant anymore?

httpapi/v2/handler.go
```diff
@@ -36,4 +36,6 @@ func Register(ctx map[string]interface{}, r chi.Router) {
 	r.Post("/documents/{"+coreapi.DocumentIDParam+"}/push_to_oracle", h.PushAttributeToOracle)
 	r.Post("/documents/{"+coreapi.DocumentIDParam+"}/attributes", h.AddAttributes)
 	r.Delete("/documents/{"+coreapi.DocumentIDParam+"}/attributes/{"+AttributeKeyParam+"}", h.DeleteAttribute)
+	r.Post("/accounts/generate", h.GenerateAccount)
+	r.Get("/jobs/{"+jobIDParam+"}", h.Job)
```
mikiquantum, line 40: Would this be the prep endpoint that will implement the new jobs/transactions v2 service?

mikiquantum (approved):
Nice! LGTM some minor comments 👍

httpapi/v2/handler.go
vedhavyas: yeap

ethereum/wait_for_event_task_test.go
vedhavyas: this old v1 celery task will be removed soon

ethereum/geth_client.go
vedhavyas: gocelery uses go encoder to encode to persist. So apart from the primitive args, we should register the args that we send to gocelery v2

config/configstore/jobs.go
vedhavyas: these are internal names, wont be used anywhere except this job. does Id has a significance ?

PR #1466, opened 2021-02-16
update notifications with different types

### LINKED ISSUE #1057: Webhook call should not use document_id for job_id
When a webhook is called for a job on document the `document_id` is actually the `job_id`.
We should add another field job_id and always have the the `document_id` set to the actual identifier of the document

At the moment you receive you mint an NFT the NotificaitonMessage is:
```
{
  event_type: 2,
  recorded: { seconds: 1558364720, nanos: 910086000 },
  document_type: 'http://github.com/centrifuge/go-centrifuge/jobs/#Job',
  status: 'pending', // This should be success
  document_id: '0xce4f423122634f478b93ef802a0eb798', // this should be the document_id 
  account_id: '0x5f2bBFcf948a0083bbE77C0aD97A64142D6b1D48' 
}
```

vedhavyas:
Closes #1057 

This PR makes the following changes:
- Webhook message contains different payloads depending on the event type i.e.. job document etc...
- Testworld will rely on webhook for job completions instead of polling the job status
	- With job polling for a given test https://travis-ci.com/github/centrifuge/go-centrifuge/jobs/484019171#L2726
	- Without job polling for the same test https://travis-ci.com/github/centrifuge/go-centrifuge/jobs/484274096#L2588
	- We poll the job only when we receive the job complete webhook

### DIFF
--- a/build/scripts/push_to_swagger.sh
+++ b/build/scripts/push_to_swagger.sh
@@ -2,11 +2,11 @@
 
 # push the swagger api json to swagger hub
 echo "pushing swagger.json to SwaggerHub"
-VERSION=`jq -c '.info.version' httpapi/swagger.json -r`
+VERSION=$(jq -c '.info.version' http/swagger.json -r)
 
-curl -i -X POST \
+curl -f -i -X POST \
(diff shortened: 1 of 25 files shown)


mikiquantum (approved):
LGTM!

PR #1467, opened 2021-03-01
Github actions

vedhavyas:


Makefile
```diff
@@ -58,8 +41,9 @@ generate: ## autogenerate go files for config
 	go generate ./config/configuration.go
 
 install-subkey: ## installs subkey
-	curl https://getsubstrate.io -sSf | bash -s -- --fast
-	cargo install --force --git https://github.com/paritytech/substrate subkey
+	wget -P ${HOME}/.cargo/bin/ https://storage.googleapis.com/centrifuge-dev-public/subkey-rc6
```
mikiquantum, line 44: Keep in mind that this will only work for a linux distribution, prob not for macOS, in which case cargo install would still help to keep in, although we are going to start using go-subkey-port so, might not be needed at all.

version/version.go
```diff
@@ -10,7 +10,7 @@ import (
 var gitCommit = "master"
 
 // CentrifugeNodeVersion is the current version of the app
-const CentrifugeNodeVersion = "1.0.0"
+const CentrifugeNodeVersion = "2.0.0"
```
mikiquantum, line 13: 😄 nice.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,57 @@
...
+name: Tests
+jobs:
+  tests:
+    name: ${{ matrix.test }} tests on ${{ matrix.os }}
+    strategy:
+      matrix:
+        go-version: [1.15.x]
+        #TODO(ved): enable macos once we port to go-subkey on gsrpc
+        os: [ubuntu-latest]
+        test: [unit, cmd, integration, testworld]
```
mikiquantum, line 15: Do these run in parallel?

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build:
+    strategy:
+      matrix:
+        go-version: [1.15.x]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      GOPATH: ${{ github.workspace }}
+      SWAGGER_API_KEY: ${{ secrets.swagger_api_key }}
```
mikiquantum, line 15: How are these secrets configured?

mikiquantum (approved):
Very cool @vedhavyas !

Makefile
vedhavyas: This is TODO yet. i'm not able to upgrade the gsrpc because for latest go-ethereum issue. Once that is fixed, mac os should run natively. Until then, we anyway throw a helpful error so its easy to identify the cause

.github/workflows/tests.yml
vedhavyas: yeap

.github/workflows/build.yml
vedhavyas: github secrets. they are decrypted transparently only pushed to develop or master

vedhavyas: they should be present under project settings

PR #1471, opened 2021-03-22
add nft on cc beta http endpoints

vedhavyas:
This PR adds 3 new nft on centrifuge chain APIs in beta

### DIFF
--- a/http/coreapi/types.go
+++ b/http/coreapi/types.go
@@ -2,6 +2,7 @@ package coreapi
 
 import (
 	"encoding/json"
+	"fmt"
 	"time"
 
 	coredocumentpb "github.com/centrifuge/centrifuge-protobufs/gen/go/coredocument"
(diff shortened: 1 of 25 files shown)


http/coreapi/types.go
```diff
@@ -154,15 +156,23 @@ func ToDocumentsCreatePayload(request CreateDocumentRequest) (documents.CreatePa
 func convertNFTs(tokenRegistry documents.TokenRegistry, nfts []*coredocumentpb.NFT) (nnfts []NFT, err error) {
 	for _, n := range nfts {
 		regAddress := common.BytesToAddress(n.RegistryId[:common.AddressLength])
+		var owner string
 		o, errn := tokenRegistry.OwnerOf(regAddress, n.TokenId)
-		if errn != nil {
-			err = errors.AppendError(err, errn)
-			continue
+		if errn == nil {
```
mikiquantum, line 161: Is there any more unique way of identifying an nft minted on the chain vs eth, instead of trial and error?
Since NFTs are coming from a document, could be add a generic attribute to the doc that states the nft origin?

On the other hand if we are not going to support both running in parallel on mainnet, it might be ok as is and just add a comment explaining that the eth check will be removed.

mikiquantum (approved):
LGTM! The whole feature planning was very cleanly executed 👍

http/coreapi/types.go
vedhavyas: I dont think we will support this for long. In the meantime, since its guaranteed to be unique, easier to do a check on both

PR #1502, opened 2022-02-20
Add Flake

asymmetric:


flake.nix
```diff
@@ -0,0 +1,85 @@
...
+            "flake.nix"
+          ];
+        in
+        path: type:
+          srcIgnored path type
+          && builtins.all (name: builtins.baseNameOf path != name) ignoreList;
+    in
+    {
+      packages.${system} = {
+        go-centrifuge = pkgs.buildGoModule {
```
mikiquantum, line 39: Sort of same question as in the other PR, how to know the go version is used, probably part of the same presentation we have to shedule.

mikiquantum (commented):
LGTM, same comment as the other PR

flake.nix
asymmetric: The most basic way to know that is to:

- inspect which commit of nixpkgs we're using in this repo, which is found in the `flake.lock`
- Run `nix run nixpkgs/$commit#go -- version`

In this specific case:

```
❯ nix run nixpkgs/6f07605b6916abee5435ea1145a7541299a87c1b#go -- version
go version go1.16.9 linux/amd64
```

But I've also implemented a simpler solution, which works by dropping you in a shell where Go is installed. Since nixpkgs is pinned, this is the same Go used to build the app.

- `nix develop`
- `go version`

PR #1506, opened 2022-03-19
Prepare for parachain launch

mikiquantum:


.github/workflows/tests.yml
```diff
@@ -33,13 +34,17 @@ jobs:
         with:
           path: ${{ env.GOPATH }}/pkg/mod
           key: ${{ runner.os }}-go-${{ matrix.test }}
+      - name: Fix up git URLs
```
mikiquantum, line 37: Needed due to https://github.com/heroku/heroku-buildpack-nodejs/issues/959

.github/workflows/tests.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       GOPATH: ${{ github.workspace }}
+      BASE_PATH: ${{ github.workspace }}/src/github.com
```
mikiquantum, line 18: Project support now go modules, so this env will allow local setup to customize the source path.

centchain/events.go
```diff
@@ -168,25 +171,19 @@ type EventNFTTransferred struct {
...
-	MultiAccount_MultiAccountUpdated    []EventMultiAccountUpdated     //nolint:stylecheck,golint
-	MultiAccount_MultiAccountRemoved    []EventMultiAccountRemoved     //nolint:stylecheck,golint
-	MultiAccount_NewMultisig            []EventNewMultisig             //nolint:stylecheck,golint
-	MultiAccount_MultisigApproval       []EventMultisigApproval        //nolint:stylecheck,golint
-	MultiAccount_MultisigExecuted       []EventMultisigExecuted        //nolint:stylecheck,golint
-	MultiAccount_MultisigCancelled      []EventMultisigCancelled       //nolint:stylecheck,golint
-	Registry_RegistryCreated            []EventRegistryCreated         //nolint:stylecheck,golint
-	Registry_Mint                       []EventNFTMint                 //nolint:stylecheck,golint
-	Nft_Transferred                     []EventNFTTransferred          //nolint:stylecheck,golint
+	cEvents
```
mikiquantum, line 179: Simplified the events supported, most of the are imported by the centEvents above

centchain/api_integration_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
mikiquantum, line 1: New way of defining annotations since go 1.17, I might clean up the older way too

build/configs/default_config.yaml
```diff
@@ -97,7 +97,7 @@ queue:
 
 # CentChain specific configuration
 centChain:
-  nodeURL: ws://127.0.0.1:9944
+  nodeURL: ws://127.0.0.1:9946
```
NunoAlexandre, line 100: Does the fact that we changed the ports in our local env to be `para_id` + `9946` matter here?

build/scripts/migrate.sh
```diff
@@ -16,6 +16,13 @@ if [ -z "${CENT_ETHEREUM_CONTRACTS_DIR}" ]; then
     CENT_ETHEREUM_CONTRACTS_DIR=${PARENT_DIR}/build/centrifuge-ethereum-contracts
 fi
 
+echo "sourcing in nvm"
+. $NVM_DIR/nvm.sh
+
+# Migrate logic is based on an old version of zeppelin_os and transient dependencies
+current_node=$(node -v)
+nvm use v10
```
NunoAlexandre, line 24: We do here `nvm use v10` here and below `nvm use $current_node`?

build/configs/default_config.yaml
mikiquantum: The chain service is meant to be run here as docker and we do not support different ports.

build/scripts/migrate.sh
mikiquantum: Since we are forced to use node 10 in this steps of the eth deployment, this code just reverts to what it was set before it needed to use v10.

build/configs/default_config.yaml
NunoAlexandre: Great!

build/scripts/migrate.sh
NunoAlexandre: Ah ok I see 👍

NunoAlexandre (approved):
LGTM but admittedly not a great reviewer here.

PR #1514, opened 2022-04-14
Update libp2p

cdamian:


testworld/document_consensus_test.go
```diff
@@ -12,7 +13,6 @@ import (
 )
 
 func TestHost_AddExternalCollaborator(t *testing.T) {
-	t.Parallel()
```
mikiquantum, line 15: Is there any main reason to making this test serial?

p2p/server.go
```diff
@@ -200,7 +203,7 @@ func makeBasicHost(ctx context.Context, priv crypto.PrivKey, pub crypto.PubKey,
 		libp2p.AddrsFactory(addressFactory),
 	}
 
-	bhost, err := libp2p.New(ctx, opts...)
+	bhost, err := libp2p.New(opts...)
```
mikiquantum, line 206: Since `libp2p.New` doesnt take a context anymore, I am wondering if they replace some of the usages of ctx by new `opts` that we may have to set. As far as I know we used to pass a cancel context but that might not be needed there anymore? How would the cancel cascading works now?

mikiquantum (commented):
Overall looks good, just some comments I am curious about.
Thank you!

testworld/document_consensus_test.go
cdamian: When running this locally, the test failed because of this.

p2p/server.go
```diff
@@ -105,6 +104,10 @@ func (s *peer) Start(ctx context.Context, wg *sync.WaitGroup, startupErr chan<-
 	}
 
 	<-ctx.Done()
+
+	if err := s.host.Close(); err != nil {
```
cdamian, line 108: @mikiquantum This is the closest thing that we have, related to the below question.

mikiquantum (approved):
LGTM!

PR #1534, opened 2022-06-29
Replace secp256k1 by ed25519

### LINKED ISSUE #1520: Replace usage of secp256k1 by ed25519
Document signing should not be using anymore secp curve.

mikiquantum:
closes #1520


NunoAlexandre (commented):
Looks good to me but don't we need to be backwards compatible and still support verifying `secp256k1` signed messages?

cdamian (approved):
LGTM

mikiquantum:
@NunoAlexandre that is a good question. The idea is that only new asset originators will use this new version, otherwise we would need to do a full migration of identities accross chains, and we didnt think it was worth it due to the current user load we have that use the go-node. So we do not need to support legacy documents.

PR #1540, opened 2022-07-27
Update build directives as per Go 1.18 release notes

akhilesharora:
Go 1.17 introduced //go:build lines as a more readable way to write build constraints, instead of // +build lines. As of Go 1.17, gofmt adds //go:build lines to match existing +build lines and keeps them in sync, while go vet diagnoses when they are out of sync.

Since the release of Go 1.18 marks the end of support for Go 1.16, all supported versions of Go now understand //go:build lines. In Go 1.18, go fix now removes the now-obsolete // +build lines in modules declaring go 1.18 or later in their go.mod files.

For more information, see https://go.dev/design/draft-gobuild.

### DIFF
--- a/anchors/anchor_test.go
+++ b/anchors/anchor_test.go
@@ -1,4 +1,4 @@
-// +build unit
+//go:build unit
 
 package anchors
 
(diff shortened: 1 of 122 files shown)


cdamian (approved):
LGTM, thank you @akhilesharora! 

@mikiquantum do we want this in `develop` or the `v1.4.0` branch?

PR #1564, opened 2022-12-21
p2p: Do not return error if no signatures are provided

cdamian:
This change is required in order to reflect the behavior mentioned in our documentation - https://docs.centrifuge.io/build/p2p-node/#no-collaborator-signatures-required-to-anchor


mikiquantum (approved):
LGTM!

PR #1594, opened 2023-03-21
Remove extra anchoring step

cdamian:
Draft until we confirm that tests are OK.


nft/v3/jobs.go
```diff
@@ -124,21 +125,20 @@ func (c *CommitAndMintNFTJobRunner) New() gocelery.Runner {
 	mj := &CommitAndMintNFTJobRunner{
```
mikiquantum, line 125: So this runner function will now take care of minting the first time a document is committed, from a state of pending. I find the name a bit confusing, knowing that the other runner `MintNFTJobRunner` will only work once the document has been committed previously so essentially making it an update document call. Maybe we can be more explicit on the naming of each, like:
`MintNFTJobRunner ` -> `UpdateMintNFTJobRunner`
or
`CommitAndMintNFTJobRunner ` -> `PendingMintNFTJobRunner`

mikiquantum (approved):
Looks good. Just a naming suggestion change so it is more clear when should it be used.

mikiquantum (approved):

PR #1597, opened 2023-06-06
Investor access

cdamian:
Adding a new endpoint that will be used by investors to retrieve a particular document that's used as collateral in a loan.

The changes done here are meant to adjust the authentication system present in the POD to allow multiple types of checks based on the endpoint and/or proxy type that are used.


http/auth/access/wrapper.go
```diff
@@ -0,0 +1,345 @@
...
+
+	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
+	"github.com/centrifuge/pod/config"
+	"github.com/centrifuge/pod/contextutil"
+	authToken "github.com/centrifuge/pod/http/auth/token"
+)
+
+//go:generate mockery --name ValidationWrapperFactory --structname ValidationWrapperFactoryMock --filename factory_mock.go --inpackage
+
+type ValidationWrapperFactory interface {
```
mikiquantum, line 16: Does it make sense to rename this as ValidationServiceFactory or AuthValidationFactory ?
I feel that `Wrapper` doesn't convey that much information.

http/auth/access/wrapper.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+type pathMatcherFn func(path string) bool
+type tokenValidationFn func(r *http.Request) (*authToken.JW3Token, error)
+type accessValidationFn func(r *http.Request, token *authToken.JW3Token) (*types.AccountID, error)
+type postAccessValidationFn func(r *http.Request, accountID *types.AccountID) error
+
+//go:generate mockery --name ValidationWrapper --structname ValidationWrapperMock --filename wrapper_mock.go --inpackage
+
+type ValidationWrapper interface {
```
mikiquantum, line 69: Maybe ValidationService? or AuthValidationService?

mikiquantum (approved):
I like how you structured the validator wrapper with its own route triggers and correspondent validation methods 👍

PR #1599, opened 2023-06-23
deps: Update chain-custom-types to v1.0.6

cdamian:


mikiquantum (approved):

cdamian:
It seems that there are some more decoding errors, still trying to reproduce locally. Going to update as soon as I manage to do that.

